PR #5, opened 2021-07-27
Recording

frostbyte73:


livekit_models.proto
```diff
@@ -67,3 +67,38 @@ message DataMessage {
...
+
+message RecordingTemplate {
+  string type = 1;
+  string ws_url = 2;
+  string token = 3;
+  string roomName = 4;
+}
+
+message RecordingOutput {
+  string file = 1;
```
davidzhao, line 88: should this be a one_of?

livekit_models.proto
```diff
@@ -67,3 +67,38 @@ message DataMessage {
     bytes binary = 2;
   }
 }
+
+message RecordingInput {
+  string url = 1;
```
davidzhao, line 72: one_of?

livekit_room.proto
```diff
@@ -37,6 +37,12 @@ service RoomService {
 
   // Subscribes or unsubscribe a participant from tracks. Requires `roomAdmin`
   rpc UpdateSubscriptions(UpdateSubscriptionsRequest) returns (UpdateSubscriptionsResponse);
+
+  // Begins recording a room
+  rpc RecordRoom(RecordRoomRequest) returns (RecordingResponse);
```
davidzhao, line 42: are these required? or another way of recording?

davidzhao (approved):

livekit_room.proto
frostbyte73: another way, if you want to start it late or end it early

PR #7, opened 2021-08-06
move recording options, use oneof

frostbyte73:


davidzhao (approved):
looks good. much cleaner.

PR #10, opened 2021-08-18
SendData Server API

ChesterMing:


livekit_room.proto
```diff
@@ -125,3 +128,14 @@ message UpdateSubscriptionsRequest {
 message UpdateSubscriptionsResponse {
   // empty for now
 }
+
+message SendDataRequest {
+  string room = 1;
+  repeated string destination_sids = 2;
```
davidzhao, line 134: nit: since destination_sids is optional, do you mind placing this last?

davidzhao (commented):
looks good, just a small nit

livekit_room.proto
ChesterMing: I made some changes, does it look better?

ChesterMing: Is the name `destination_sids` ok? or do you have suggestions. 😁

davidzhao (approved):
looks good!

PR #13, opened 2021-08-24
packet and system node stats

matkam:


livekit_internal.proto
```diff
@@ -20,10 +20,23 @@ message NodeStats {
   int64 started_at = 1;
   // when server last reported its status
   int64 updated_at = 2;
-  uint32 num_rooms = 3;
-  uint32 num_clients = 4;
-  uint32 num_tracks_in = 5;
-  uint32 num_tracks_out = 6;
+  // when server previously reported its status
+  int64 updated_at_previous = 7;
```
davidzhao, line 24: should this be local to the node that is computing the other stats? any value in persisting?

davidzhao (approved):
lg! is there any way to capture IO data as well? i.e. ingress_per_sec, egress_per_sec

also, thinking out loud, should the load info be ammortized by # of CPUs available? If so we should probably include CPU count in here.

matkam:
I'll add IO. CPU count is a good call too.

PR #18, opened 2021-09-02
simplify recording reservation

frostbyte73:


livekit_recording.proto
```diff
@@ -34,24 +34,15 @@ message RecordingInput {
...
   string layout = 1;
-  string ws_url = 2;
   // either token or room name required
   string token = 3;
   string room_name = 4;
 }
 
 message RecordingOutput {
-  string file = 1;
-  RecordingS3Output s3 = 2;
```
davidzhao, line 45: we should leave s3 in there?

frostbyte73: it's not a RecordingS3Output anymore, just a string below

davidzhao: ohhh I see. my bad reading the diff

davidzhao (approved):

PR #21, opened 2021-09-10
Fix logr usage

hn8:
- logr.Error() doesn't check level
- logr convention: V0 info, V1 debug, V2 trace
- Also remove unused Zap dependency


hn8:
@davidliu @frostbyte73 logger.Warnw is a little awkward with logr. It is basically the same level with Info() with optionally extra error. Their philosophy is that warn should either error (should be handled) or info (can be ignored for now).
https://dave.cheney.net/2015/11/05/lets-talk-about-logging

frostbyte73 (approved):
Lgtm. I added init functions so the recorder and any future go repos can use it, but since server can't share it and it's not being used yet, might as well remove the dep

PR #23, opened 2021-09-17
Send speaker changes as deltas, instead of comprehensive list

davidzhao:
For client protocol 3, server will send speaker updates as a series of deltas, allowing client to apply the changes. This prepares for the future where a single SFU may not have access to to all active speakers in the room,


hiroshihorie (approved):
🚀

PR #25, opened 2021-09-22
notifier interface, message bus queue

frostbyte73:


utils/messaging.go
```diff
@@ -2,16 +2,21 @@ package utils
...
 	"github.com/eapache/channels"
 	"github.com/go-redis/redis/v8"
 )
 
+const lockExpiration = time.Second * 5
+
 type MessageBus interface {
 	Lock(ctx context.Context, key string, expiration time.Duration) (acquired bool, err error)
 	Subscribe(ctx context.Context, channel string) (PubSub, error)
+	Queue(ctx context.Context, channel string) (PubSub, error)
```
davidzhao, line 19:
```suggestion
	// like subscribe, but ensuring only a single instance gets to process the message.
	SubscribeQueue(ctx context.Context, channel string) (PubSub, error)
```

davidzhao (approved):
lg, would just change method name to be clear.

PR #26, opened 2021-09-23
add room metadata field, update method and update signal

lukasIO:
Additions to the protocol to add support for room metadata as proposed in https://github.com/livekit/livekit-server/issues/115#issuecomment-924623911


FeepsDev:
You may also need to add the UpdateRoomMetadataRequest message to [RTCNodeMessage](https://github.com/livekit/protocol/blob/e3e7ce02322068f1ce25fe9eddf3edb9ea798e2e/livekit_internal.proto#L65)

livekit_room.proto
```diff
@@ -39,6 +39,9 @@ service RoomService {
 
   // Send data over data channel to participants in a room, Requires `roomAdmin`
   rpc SendData(SendDataRequest) returns (SendDataResponse);
+
+  // Update room metadata, will cause updates to be broadcasted to everyone in the room, Requires `roomAdmin`
+  rpc UpdateRoomMetadata (SendDataRequest) returns (SendDataResponse);
```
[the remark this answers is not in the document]
lukasIO: thanks! pushed an update

lukasIO:
> 
> 
> You may also need to add the UpdateRoomMetadataRequest message to [RTCNodeMessage](https://github.com/livekit/protocol/blob/e3e7ce02322068f1ce25fe9eddf3edb9ea798e2e/livekit_internal.proto#L65)

thank you! I added it in 43870f0.

> I can do the server implementation if you want ;)
Sure, I'm anyways mostly guessing what the right implementation would be, just have a bit of spare time at the moment to get things moving :)

livekit_room.proto
```diff
@@ -141,3 +144,9 @@ message SendDataRequest {
 message SendDataResponse {
   //
 }
+
+message UpdateRoomRequest {
+  string room = 1;
```
lukasIO, line 149: @FeepsDev I wasn't sure if we'd need an identity field here as well? I left it out for now, but start to think that was a mistake...

livekit_internal.proto
```diff
@@ -74,6 +74,7 @@ message RTCNodeMessage {
     DeleteRoomRequest delete_room = 7;
     UpdateSubscriptionsRequest update_subscriptions = 8;
     SendDataRequest send_data = 9;
+    UpdateRoomRequest update_room = 10;
```
davidzhao, line 77:
```suggestion
    UpdateRoomMetadataRequest update_room_metadata = 10;
```

livekit_room.proto
```diff
@@ -39,6 +39,9 @@ service RoomService {
 
   // Send data over data channel to participants in a room, Requires `roomAdmin`
   rpc SendData(SendDataRequest) returns (SendDataResponse);
+
+  // Update room metadata, will cause updates to be broadcasted to everyone in the room, Requires `roomAdmin`
+  rpc UpdateRoomMetadata (UpdateRoomRequest) returns (Room);
```
davidzhao, line 44:
```suggestion
  rpc UpdateRoomMetadata (UpdateRoomMetadataRequest) returns (Room);
```

livekit_room.proto
davidzhao: I don't think identity is needed. it is a room-level request.

davidzhao (commented):
this looks good! Re: the differences of UpdateRoom vs UpdateRoomMetadata, I think the server API to update should be more focused. Since the use cases are likely changing the room metadata frequently, but not the other attributes, a focused API seems to make more sense.

on the client side, `RoomUpdate` is more appropriate, since we could notify clients of any room updates via the same message. Doing so will reduce the amount of state handling code we'd have to write on the clients.

livekit_room.proto
FeepsDev: it should returns UpdateRoomMetadataResponse

livekit_room.proto
```diff
@@ -141,3 +144,9 @@ message SendDataRequest {
 message SendDataResponse {
   //
 }
+
+message UpdateRoomRequest {
+  string room = 1;
+  // metadata to update. skipping updates if left empty
+  string metadata = 3;
```
davidzhao, line 151:
```suggestion
  string metadata = 2;
```
fields should be sequential, unless we've removed fields.

davidzhao (commented):
This is looking good! Just one nit and we are good to merge.

livekit_room.proto
```diff
@@ -39,6 +39,9 @@ service RoomService {
 
   // Send data over data channel to participants in a room, Requires `roomAdmin`
   rpc SendData(SendDataRequest) returns (SendDataResponse);
+
+  // Update room metadata, will cause updates to be broadcasted to everyone in the room, Requires `roomAdmin`
+  rpc UpdateRoomMetadata (UpdateRoomMetadataRequest) returns (Room);
```
FeepsDev, line 44: It should returns UpdateRoomMetadataResponse message

davidzhao: I think this is actually similar to the other APIs.. where Create/Update/Get returns the object itself without wrapping.

FeepsDev: ah ok, I thought rather that this API would return nothing

davidzhao:
Tagged as v0.9.4, ready for the server side now.

lukasIO:
great, thank you for the help!

PR #28, opened 2021-10-04
recording v2

frostbyte73:


recording/channels.go
```diff
@@ -0,0 +1,65 @@
...
+
+func RequestChannel(id string) string {
+	return "RECORDING_REQUEST_" + id
+}
+
+func ResponseChannel(id string) string {
+	return "RECORDING_RESPONSE_" + id
+}
+
+func RecordingRPC(ctx context.Context, mb utils.MessageBus, recordingId string, req *livekit.RecordingRequest) error {
```
davidzhao, line 29: would you ever return a non-void response?

recording/channels.go
```diff
@@ -0,0 +1,65 @@
...
+		case <-time.After(RequestTimeout):
+			return errors.New("request timeout")
+		case msg := <-sub.Channel():
+			b = sub.Payload(msg)
+			resp := &livekit.RecordingResponse{}
+			err = proto.Unmarshal(b, resp)
+			if err != nil {
+				return err
+			}
+			if resp.RequestId != req.RequestId {
```
davidzhao, line 56: what if multiple of these are going and you are seeing a timing issue? this seem to suggest only one response could be in flight.

frostbyte73: if multiple are going, there will be multiple subscribers, and each will only care about their own request/response id

recording/channels.go
frostbyte73: maybe in the future, but for now it's just an error

davidzhao (approved):
👍

PR #31, opened 2021-10-09
Add `disable_dtx` field to `AddTrack` request.

boks1971:
DTX (Discontinuous Transmission) is used to save
on audio bandwidth during silence periods.

It is enabled from server side in SDP answer.
Add a field to allow configuration from client side.


davidzhao (approved):
looks good!

PR #32, opened 2021-10-14
Track source for common types

davidzhao:
This enables clients to have a simpler API to manage tracks.


boks1971 (approved):
@davidzhao For my own education, trying to understand how it will be used on the wire and server side.

Changes look good. Just one question on specificity of sources? Do we want it be more specific like `FRONT_CAMERA`, `BACK_CAMERA` etc (that is an endless list) or just a notion of media stream source?

PR #37, opened 2021-11-02
Abiltiy to use non default logger

matkam:


logger/logger.go
```diff
@@ -19,20 +19,44 @@ func SetLogger(l logr.Logger, name string) {
...
 
 func Warnw(msg string, err error, keysAndValues ...interface{}) {
+	WithLogger(defaultLogger).Warnw(msg, err, keysAndValues...)
+}
+
+func Errorw(msg string, err error, keysAndValues ...interface{}) {
+	WithLogger(defaultLogger).Errorw(msg, err, keysAndValues...)
+}
+
+type withLogger struct {
```
davidzhao, line 37: I think you can also do `type withLogger logr.Logger`, and then

`withLogger(defaultLogger).Errorw`

davidzhao (approved):
lg!

PR #39, opened 2021-11-08
Add recording_started webhook, add ctx to notifier

frostbyte73:


davidzhao (approved):

livekit_webhook.proto
```diff
@@ -4,20 +4,29 @@ package livekit;
...
   Room room = 2;
 
   // set when event is participant_*
   ParticipantInfo participant = 3;
 
+  // set when event is recording_started
+  RecordingInfo recording_info = 5;
+
   // set when event is recording_finished
   RecordingResult recording_result = 4;
```
jhwillett, line 22: It seems odd to list `recording_info` (with tag 5) before `recording_result` (with tag 4).

PR #40, opened 2021-11-09
Add analtyics proto stub - proto defn to be updated

shishirng:


analytics.proto
```diff
@@ -0,0 +1,54 @@
...
+
+import "google/protobuf/empty.proto";
+import "google/protobuf/timestamp.proto";
+
+service AnalyticsRecorderService {
+  rpc IngestStats(stream  AnalyticsStats) returns (google.protobuf.Empty){};
+  rpc IngestEvents(stream AnalyticsEvents) returns (google.protobuf.Empty){};
+}
+
+enum STATSSTREAMTYPE {
```
davidzhao, line 14: I think for consistency sake, let's keep these to be CamelCase. same as enum values. I would just call this `StreamType`

analytics.proto
```diff
@@ -0,0 +1,54 @@
...
+  STATSUPSTREAM = 0;
```
davidzhao, line 15: We don't need the prefixes here. Should we use upstream/downstream or publish/subscribe?

analytics.proto
```diff
@@ -0,0 +1,54 @@
...
+  STATSDOWNSTREAM = 1;
+}
+
+message AnalyticsStat {
+  STATSSTREAMTYPE kind = 1;
+  google.protobuf.Timestamp timestamp = 2;
+  string node = 3;
+  optional string SID = 4;
```
davidzhao, line 23:
```suggestion
  optional string sid = 4;
```

analytics.proto
```diff
@@ -0,0 +1,54 @@
...
+  optional string projectId = 5;
```
davidzhao, line 24:
```suggestion
  optional string project_id = 5;
```

davidzhao: as a convention, we tend to use snake_case for field names.

analytics.proto
```diff
@@ -0,0 +1,54 @@
...
+  optional string participantId = 6;
+  optional string roomName = 7;
+  optional double jitter = 8;
+  optional uint64 packetLost = 9;
+  optional uint64 rrtime = 10;
+  optional uint64 bytesSent = 11;
+  optional uint64 bytesReceived = 12;
+  optional string authToken = 13;
```
davidzhao, line 32: I don't think this should be in the protobuf itself.. it should be sent in the request, but left out of the db record.

analytics.proto
```diff
@@ -0,0 +1,54 @@
...
+}
+
+message AnalyticsStats {
+  repeated AnalyticsStat stats = 1;
+}
+
+enum EventSignalTarget {
```
davidzhao, line 39: It feels that we might not need this. most events are agnostic of the direction, and for ones that are, we should re-use `StreamType`

analytics.proto
shishirng: will use Publish/Subscribe

analytics.proto
shishirng: had move this to payload, as in streaming client, the context is per message. This is not committed to DB though

davidzhao (approved):
lg!

PR #46, opened 2021-11-15
Add sdk type, region and access_key to events message

shishirng:


frostbyte73 (approved):

analytics.proto
```diff
@@ -18,7 +18,7 @@ enum StreamType {
 }
 
 message AnalyticsStat {
-  string auth_token = 1;
+  string access_key = 1;
```
davidzhao, line 21: main concern with this name is if it'll be confusing compared to our `access token`. would it be better calling this the `analytics key` ?

analytics.proto
```diff
@@ -51,6 +51,15 @@ enum AnalyticsEventType {
   RECORDING_ENDED = 9;
 }
 
+enum SdkType {
```
davidzhao, line 54: We already have ClientInfo.SDK, can we use that instead of creating a new one?

```
message ClientInfo {
  enum SDK {
     UNKNOWN = 0;
     JS = 1;
     IOS = 2;
     ANDROID = 3;
     FLUTTER = 4;
     GO = 5;
  }
```

PR #49, opened 2021-11-24
Protocol message to notify streaming status.

boks1971:
With `StreamAllocator`, SFU could pause/resume streams
based on bandwidth availability. Adding an update
message to notify clients about the state of streaming.


boks1971:
Reviewers: please let me know if you can think of better names for the fields.

davidzhao (approved):
looks good!

PR #51, opened 2021-12-04
Change structure of streaming status update

boks1971:
Use a single list with an enum for state indication.


davidzhao (approved):
looks good! this seems simpler.

Since we haven't released the protobuf definitions to any clients or servers. it seems ok to swap out the previous definitions.

boks1971:
> looks good! this seems simpler.
> 
> Since we haven't released the protobuf definitions to any clients or servers. it seems ok to swap out the previous definitions.

Yeah, that's why I just did a replace.

PR #52, opened 2021-12-05
Client provided layer size information

davidzhao:


livekit_models.proto
```diff
@@ -73,6 +73,23 @@ message TrackInfo {
...
+
+enum VideoQuality {
+  LOW = 0;
+  MEDIUM = 1;
+  HIGH = 2;
+}
+
+// provide information about available spatial layers
+message VideoLayer {
+  // for non-simulcast or SVC, this should be HIGH
```
boks1971, line 87: I am thinking SVC should report configured layers too. Not sure though if there is any event on encoder internally changing dimensions. Maybe, for now leave out comment about SVC having to use HIGH?

boks1971 (approved):

davidzhao:
@hiroshihorie do you think this could work for the dynamic-dimension feature that you had planned?
@frostbyte73 we are adding another field to TrackInfo, wanted to consider how this data is synchronized within cloud.

livekit_models.proto
davidzhao: yes, I meant to type non-(simulcast or SVC), but it came out wrong.. lol

davidzhao:
> Dimensions information will be required per layer ?

yes, this allows you to provide custom layer sizes when simulcasting.

PR #54, opened 2021-12-07
Moving generated Go protobuf files to match go_package

jhwillett:
Evaluating whether, by rearranging https://github.com/livekit/protocol, we can produce `*.proto` and `*.pb.go` pairs which can both be imported from proto files in other projects.


magefile.go
```diff
@@ -9,36 +9,23 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-
-	"github.com/magefile/mage/target"
```
jhwillett, line 13: This seems to have been removed by `goimports`.

magefile.go
```diff
@@ -9,36 +9,23 @@ import (
...
-	)
-	if err != nil {
-		return err
+	twirpProtoFiles := []string{
+		"proto/livekit_recording.proto",
+		"proto/livekit_room.proto",
 	}
-	if !updated {
-		return nil
-	}
```
jhwillett, line 34: I have removed the dirtyness check.  We do not practice it in other projects for either wire nor protobuf generation, and it feels dangerous.

I did add more explicit lists of source files.

proto/livekit_analytics.pb.go
```diff
@@ -2,9 +2,9 @@
 // versions:
 // 	protoc-gen-go v1.27.1
 // 	protoc        v3.17.3
-// source: livekit_analytics.proto
+// source: proto/livekit_analytics.proto
 
-package livekit
+package proto
```
jhwillett, line 7: We have a gotcha: an unfortunate trade-off.

With `--go_out . --go_opt=paths=source_relative`, which is the only pattern that puts `livekit_*.pb.go` in this location, the package name must match the directory name.

That is, if we wish to have Go files at `proto/livekit_*.pb.go`, they must have `package proto`.

If we wish them to have `package livekit`, we must indicate `go_package = something/livekit` in the `*.proto` files and then we will end up with `livekit/livekit_*.pb.go`.

Either way we will break some kind of backward compatibility.  Either the `livekit_*.pb.go` package names must change, or their paths must change.

jhwillett: IMO, of these alternatives the prettier one is to end up with `package livekit` and to put all the files in `livekit/livekit_*.pb.go`.

However, before changing it again I will wait for a discussion and the conclusion of my test in another project, of importing these types.

frostbyte73 (approved):

PR #59, opened 2021-12-21
Publisher controlled track subscriptions

boks1971:


livekit_rtc.proto
```diff
@@ -193,3 +198,19 @@ message SubscribedQualityUpdate {
...
+  repeated string track_sids = 3;
+}
+
+message UpdateSubscriptionPermissions {
+  bool all_participants = 1;
+  repeated TrackPermission track_permissions = 2;
+}
+
+message SubscriptionPermissionUpdate {
+  string track_sid = 1;
```
davidzhao, line 214: should we include a participant_sid too? to make it easier to identify

davidzhao (approved):

livekit_rtc.proto
boks1971: Good idea. Added in this commit - https://github.com/livekit/protocol/pull/59/commits/2819d3265d70c8d1ad194731da32b0a05bcd86d1

PR #60, opened 2021-12-21
Store connection ratings with track info

shishirng:


livekit_analytics.proto
```diff
@@ -34,6 +34,8 @@ message AnalyticsStat {
   int32 pli_count = 13;
   int32 fir_count = 14;
   string room_name = 15;
+  float rating = 16;
```
davidzhao, line 37: nit: do we use the terminology rating or score? can we prefix it with `connection_` to disambiguate?

davidzhao (approved):

livekit_analytics.proto
shishirng: i jupdated it to score, will prefix connection_

PR #62, opened 2021-12-27
LiveKit type aliases

boks1971:
Idea is to start with aliases so that this can be done in small steps.
The plan is to move to a type definition once we have most of the stuff
moved over to get stronger type checks.

### DIFF
--- a/livekit/types.go
+++ b/livekit/types.go
@@ -1,0 +1,9 @@
+package livekit
+
+type TrackSid = string
+
+type ParticipantSid = string
+type ParticipantIdentity = string
+
+type RoomSid = string
+type RoomName = string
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
lgtm! perhaps `TrackID` etc as the type name? matching the var?

boks1971:
> lgtm! perhaps `TrackID` etc as the type name? matching the var?

Cool @davidzhao. I can make the change. So, is the poll winner `trackID TrackID`?

davidzhao:
> Cool @davidzhao. I can make the change. So, is the poll winner `trackID TrackID`?

Looks that way :)

boks1971:
Changed it in this commit - https://github.com/livekit/protocol/pull/62/commits/2eae90f0d6d54d67a0658b37e55a772f39619029

PR #68, opened 2021-12-31
Make room name a type definition

boks1971:
Will submit PRs for code depending on this.

@davidzhao @frostbyte73 Was thinking of doing
- A `String()` method of `RoomName` to convert to string
- A `NewRoomName(roomName string) livekit.RoomName` to convert from string.
That would make it more idiomatic and also keep it flexible for the future (although I am guessing we are never going to change it from being a string).

But, did not do those methods in this PR and doing straight casting at call sites.

Please let me know if I should switch to these methods to make things cleaner and I can amend this PR.

### DIFF
--- a/livekit/types.go
+++ b/livekit/types.go
@@ -7,4 +7,27 @@ type ParticipantIdentity = string
 type ParticipantName = string
 
 type RoomID = string
-type RoomName = string
+
+//----------------------------------------------------------------
+
+type RoomName string
+
+func RoomNamesAsStrings(roomNames []RoomName) []string {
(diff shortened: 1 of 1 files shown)


davidzhao (approved):

boks1971:
Thank you for the review @davidzhao . Thoughts on the question I put in the PR comments ☝️ ?

davidzhao:
Straight casting seems fine to me. It's common to do so in Go between typedef'ed types. `String()` often indicates a representation of the object, rather than the object itself.

PR #71, opened 2022-01-07
Session migration

cnderrauber:


livekit/types.go
```diff
@@ -5,7 +5,7 @@ package livekit
 type TrackID string
 
 func StringsAsTrackIDs(trackIDs []string) []TrackID {
-	var asTrackID []TrackID
+	asTrackID := make([]TrackID, 0, len(trackIDs))
```
boks1971, line 8: Probably can make the same change in room names function at the bottom on this file to be consistent.

livekit_rtc.proto
```diff
@@ -217,3 +219,9 @@ message SubscriptionPermissionUpdate {
   string track_sid = 2;
   bool allowed = 3;
 }
+
+message SyncState {
+  SessionDescription anwser = 1;
```
boks1971, line 224: Can we name this `subscriber_answer` to indicate that it corresponds to subscriber peer connection?

boks1971 (approved):
A couple of small comments

livekit_internal.proto
```diff
@@ -107,6 +107,7 @@ message StartSession {
   bool recorder = 12;
   ClientInfo client = 11;
   string name = 13;
+  bool migrate = 14;
```
davidzhao, line 110: is this needed? or could we determine from protocol_version + which node is receiving the message?

cnderrauber: removed

livekit_rtc.proto
cnderrauber: In our SignalRequest, answer is only used for subscribe peer connection, so use answer for consistent

livekit/types.go
cnderrauber: Done

PR #75, opened 2022-01-13
Moved ClientInfo to models.proto, added additional device details

davidzhao:


livekit_models.proto
```diff
@@ -145,3 +145,25 @@ message ParticipantTracks {
   string participant_sid = 1;
   repeated string track_sids = 2;
 }
+
+// details about the client
+message ClientInfo {
+  enum SDK {
+     UNKNOWN = 0;
+     JS = 1;
+     IOS = 2;
```
hiroshihorie, line 154: Should we change this to SWIFT?

hiroshihorie (approved):
✅

livekit_models.proto
davidzhao: first step in iOS -> SWIFT conversion. I gotta get used to this!

PR #83, opened 2022-01-28
Capture audio_level and active_speaker info in stats for tracks

shishirng:


livekit_analytics.proto
```diff
@@ -38,6 +38,8 @@ message AnalyticsStat {
   float connection_score = 16;
   string track_id = 17;
   uint32 rtt = 18;
+  float audio_level = 19;
+  bool active_speaker = 20;
```
boks1971, line 42: How are you planning to capture this @shishirng , i. e. at what frequency? These are potentially very fast changing. So, capturing them at an interval may not provide useful information. Capturing them too often might be a strain on the system.

shishirng: @boks1971 currently this is captured every 5 seconds (connectionQualityUpdateInterval)

boks1971: Got it Shishir. My concern is that window is too long for this signal.

Maybe a better option is to have 
- audio_level: which is the sum of all audio levels in that window
- number_active_speakers: number of times the speaker crossed the active speaker threshold in that window.

Let me think a bit more about this. If you have figured out how it is going to be used, great!

PR #86, opened 2022-02-05
Adding total_frames

boks1971:
Also setting up per stream stats


livekit_analytics.proto
```diff
@@ -19,6 +19,19 @@ enum StreamType {
   DOWNSTREAM = 1;
 }
 
+message AnalyticsStream {
```
boks1971, line 22: @shishirng @davidzhao Thinking of this model for per stream stats. If simulcast, will be sending one of these per available layer. That means, this may be have missing samples as layers could be turned on/off.

The track level stuff will be an aggregate of these (except for jitter and delay which will be the max of these).

Just wanted to have granular analytics which we can use however we see fit while processing them. It gives the full picture.

For SVC codecs (VP9, etc.), there is only one stream with layers interleaved in them. So, those kind of codecs will not have all of these filled out at layer level. The things that make sense at layer level for those codecs are
```
total_packets
total_bytes
total_frames
```
Rest are not applicable at layer level and they will be in the track level stats.

What do you think?

livekit_analytics.proto
```diff
@@ -38,6 +51,8 @@ message AnalyticsStat {
   float connection_score = 16;
   string track_id = 17;
   uint32 rtt = 18;
```
boks1971, line 53: This is not used and it was at the end. So, I was tempted to take it out. But, left it in there just in case it breaks other stuff out there.

livekit_analytics.proto
```diff
@@ -38,6 +51,8 @@ message AnalyticsStat {
...
+  uint64 total_frames = 19;
```
boks1971, line 54: I was thinking of this as aggregate of all layers. If we wanted to show the fps of the highest layer only, it can be determined from the `streams` array below. As layers could come and go, sampling every 5 seconds could get tricky. For example, highest layer could have started at 4 seconds into the measurement window and done 30 fps. But, averaging over the 5 second sampling window, fps could look really low. How should we accommodate for that?

livekit_analytics.proto
boks1971: Actually, `delay` does not make sense at all. `rtt` makes more sense. But, `rtt` will be available for down tracks only (i. e. it can be calculated on media sender side using RTCP receiver report received from the other side). For up tracks, we won't have RTT. We can probably use some other mechanism (like using STUN ping time or just use down track RTT for up track also).

I will drop populating `delay` in the analytics message.

livekit_analytics.proto
boks1971: With reference to the `delay` comment above, actually this is more useful when available. `delay` is not. Will look at populating this when available.

livekit_analytics.proto
shishirng: we don't need this field. we can accumulate from streams downstream. since every layers data is available in an interval we should be covered to either get total and/or specific layer

livekit_analytics.proto
shishirng: @boks1971 this looks good

livekit_analytics.proto
boks1971: Thank you @shishirng . I know it will makes things incompatible and a breaking change. But, given that this is not released yet, wondering if we can clean up all the layer specific fields in this message and just move to an array of `AnalyticsStream` here? Like you said, down stream processing can do the necessary aggregation/filtering by taking data from layers.

shishirng: @boks1971 that should be fine (breaking changes). For audio, im assuming 1 of the layers will be used in the stream?

boks1971: That's great @shishirng . Let me clean it up and push up a change for your review.

If we are going the streams route. I will just use it for everything. For non-simulcast, there will be just be one stream with layer set to 0 (i. e. audio streams and video streams which do not use simulcast like screen share).

livekit_analytics.proto
```diff
@@ -19,9 +19,16 @@ enum StreamType {
   DOWNSTREAM = 1;
 }
 
-message AnalyticsStream {
+message AnalyticsVideoLayer {
```
boks1971, line 22: Taking the opportunity to clean up and making it possible to use it for both simulcast/non-simulcast/SVC streams.

Limiting to spatial layers only. Not doing temporal/quality layer granularity as it could get too complicated.

So, I have split things as follows
- `AnalyticsStat` -> Track level info. This has an array of `AnalyticsStream`. One per SSRC in that track. For non-simulcast/SVC tracks, there will be only one element in that list. For simulcast streams, this will have as many streams as the simulcast stream is using.
- `AnalyticsStream` -> RTP Stream (identified by SSRC) level information. This has an array of `ArrayVideoLayer`. For audio, that will be empty. For non-simulcast video also, it will be empty (or it can have one layer if that is preferable). For simulcast or SVC, the array will have as many spatial layers in the stream.
- `AnalyticsVideoLayer` -> Per spatial layer information. Stats that are applicable here are `total bytes`, `total packets` and `total frames`.

livekit_analytics.proto
boks1971: Made this commit to refactor - https://github.com/livekit/protocol/pull/86/commits/d16770b5fa695b460cb8574a2104429df3af51ed

Also added explanation in this comment - https://github.com/livekit/protocol/pull/86/commits/d16770b5fa695b460cb8574a2104429df3af51ed#r800082733

Let me know.

PR #90, opened 2022-02-18
client configuration system

lukasIO:
Tried to stick closely to @cnderrauber 's proposal. 
Biggest change to that was trying to unify how the audio and video configuration message is constructed


livekit_models.proto
```diff
@@ -167,3 +167,48 @@ message ClientInfo {
...
+    // echo cancellation
+    bool aec = 1;
+    // noise suppression
+    bool ns = 2;
+    // automatic gain control
+    bool agc = 3;
+  }
+  message AudioPreferences {
+    bool dtx = 1;
+    int32 sample_rate = 2;
```
lukasIO, line 189: I rather naively replaced nb/fwb/wb with a more generic sample_rate, not sure if that covers the exact same scenario?

livekit_models.proto
```diff
@@ -167,3 +167,48 @@ message ClientInfo {
...
+  message VideoPreferences {
+    string facing_mode = 1;
+  }
+  Codec preferred_codec = 1;
+  VideoProcessing processing = 2;
+  VideoPreferences preferences = 3;
+}
+
+message SignalConfiguration {
+	bool migration = 1;
```
lukasIO, line 213: I just blindly copied this over from the proposal, what would setting `migration` do on a client? 
Enable the ability to migrate when needed ?

livekit_models.proto
boks1971: Yeah, `sample_rate` will cover everything. The trick there would be to check how browsers behave if given random sample rates. But, I am guessing we will never give random sample rate. So, keeping it as sample rate should be more flexible.

livekit_models.proto
```diff
@@ -169,3 +169,48 @@ message ClientInfo {
   string browser_version = 8;
   string address = 9;
 }
+
+// server provided client configuration
+message RTCClientConfiguration {
+	AudioConfiguration audio = 1;
+  VideoConfiguration video = 2;
+  SignalConfiguration signal = 4;
```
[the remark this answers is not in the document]
boks1971: Maybe the 3 was meant to be ScreenshareConfiguration? I think we should have a separate screen share configuration as that allows us the flexibility of using a different codec and different settings.

livekit_models.proto
```diff
@@ -167,3 +167,48 @@ message ClientInfo {
...
+  message AudioPreferences {
+    bool dtx = 1;
+    int32 sample_rate = 2;
+    int32 channel_count = 3;
+  }
+  Codec codec = 1;
+  AudioProcessing processing = 2;
+  AudioPreferences preferences = 3;
+
+  // TODO sample/playout config(micphone, speaker, earphone ...)
```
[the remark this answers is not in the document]
boks1971: Will defer to @cnderrauber for clarification.

I think this might be more to do with mobile devices. In general, I think this would something like a exclusion list. For example, it might be possible that certain devices are not supported. Even on desktop, I have run into WebRTC not able to support certain combination of OS + device. I am thinking we can use this object to specify such exclusion and potentially emit an info/warn event from SDK so that devs have the ability to display something like "Potentially unsupported device, your mileage may vary". Just a thought.

livekit_models.proto
boks1971: I am guessing that is it @lukasIO . But, should confirmation from @cnderrauber

livekit_models.proto
```diff
@@ -169,3 +169,48 @@ message ClientInfo {
...
+}
+
+message VideoConfiguration {
+  message VideoProcessing {
+    bool hardware_decoder = 1;
+    bool hardware_encoder = 2;
```
boks1971, line 204: I am guessing this may need to be flipped. Have `software_*` here. Most probably there are going to be issues with hardware encoder in some combinations and we would want to use software encoding/decoding in such cases. As protobuf will not send anything on the wire for default value, we will not be able signal to turn off hardware encoder.

livekit_models.proto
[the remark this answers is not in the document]
boks1971: On the video render side, I have seen `hardware_acceleration` flag in browsers. I think that is the difference of using GPU/WebGL (sorry I do not know the details how these things work, need to educate myself), discrete GPU vs integrated GPU etc. There is probably some combination of CPU + GPU + OS where disabling hardware acceleration will increase chances of things working. It is probably a large matrix. Just noting down stuff from experience :-)

boks1971 (approved):

livekit_models.proto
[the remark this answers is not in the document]
lukasIO: you are totally right, initially I had screenshare configuration in there as well. Will readd.

livekit_models.proto
lukasIO: good point!
afaik the `hardware_acceleration` flag is a browser wide setting and nothing we can turn on/off from JS within a browser.

livekit_models.proto
lukasIO: generally (talking about browsers only) the way we request the preferences we set for certain tracks will *try* to get fullfilled by the browser. In cases we have a random sample rate that doesn't make sense, the brower *should* return the closest appropriate sample rate. I have tested this in the past for other constraints, but will have to confirm for `sampleRate`

livekit_models.proto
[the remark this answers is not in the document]
lukasIO: I added it back, but without resorting to an extra `ScreenConfiguration` message and instead reusing `VideoConfiguration` for simplicity. I imagine that would cover *most* cases, but unsure if we'd want to separate `VideoPreferences` from `ScreenSharePreferences` as they would probably have at least some options that don't overlap.

livekit_models.proto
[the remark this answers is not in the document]
cnderrauber: Yeah, it's about mobile devices that have more than one capture/playback devices that can't work  appropriately , we need to set device manually to make it work. Most of those cases is bluetooth earpods at android can't capture/playback at desired samplerate/channel count

livekit_models.proto
cnderrauber: Thanks, you are totally right @boks1971

livekit_models.proto
```diff
@@ -169,3 +169,49 @@ message ClientInfo {
...
+// server provided client configuration
+message RTCClientConfiguration {
+  AudioConfiguration audio = 1;
+  VideoConfiguration video = 2;
+  VideoConfiguration screen = 3;
+  SignalConfiguration signal = 4;
+}
+
+message AudioConfiguration {
+  message AudioProcessing {
```
davidzhao, line 182: how should we interpret the values in this message? does it mean

* client should use <feature> by default
* client supports/does not support <feature>

when a user requests for these features explicitly, how would we reconcile the final decision?

livekit_models.proto
```diff
@@ -169,3 +169,49 @@ message ClientInfo {
...
+  }
+  message VideoPreferences {
+    string facing_mode = 1;
+  }
+  Codec preferred_codec = 1;
+  VideoProcessing processing = 2;
+  VideoPreferences preferences = 3;
+}
+
+message SignalConfiguration {
```
davidzhao, line 215: maybe this is less about signal, and more about general capabilities of the client?

livekit_models.proto
```diff
@@ -169,3 +169,49 @@ message ClientInfo {
   string browser_version = 8;
   string address = 9;
 }
+
+// server provided client configuration
+message RTCClientConfiguration {
```
davidzhao, line 174: nit: we can drop the `RTC` prefix.

livekit_models.proto
cnderrauber: @lukasIO @boks1971 Good point! For this reason, it would be better to use option always change common behavior,  so use disable_xxx for other options like aec/ns too instead of enable it explicit.

lukasIO: from reading through the comments on the proposal, I thought that the configuration system would supply defaults, that could still be overriden by an sdk. 
with the example of `disable_dtx`, should the client implementations assume `dtx: true` if the setting is not present or should they just do nothing in that case and default to whatever is hardcoded as the default?

livekit_models.proto
lukasIO: would you prefer renaming it to `GeneralConfiguration` ?

cnderrauber (approved):

livekit_models.proto
davidzhao: maybe it doesn't need to be grouped?

PR #94, opened 2022-02-23
file upload options

frostbyte73:


livekit_egress.proto
```diff
@@ -75,7 +75,31 @@ enum EncodedFileType {
 
 message EncodedFileOutput {
   EncodedFileType file_type = 1; // required, must be compatible with codec selection
-  string http_url = 2;
+  string filepath = 2;
+  oneof output {
+    S3Upload s3 = 3;
+    AzBlobUpload az = 4;
```
davidzhao, line 81: nit: I'm not sure if az is a common acronym for azure.. is it worth spelling out?

davidzhao (approved):
lg!

PR #100, opened 2022-03-02
Server region

davidzhao:
Server will pass back the current region that the participant is connected to, which could be useful for debugging


boks1971 (approved):

hiroshihorie (approved):
Nice! We don't need `nodeId` ?

lukasIO (approved):

PR #107, opened 2022-03-20
Server-initiated unpublish & permissions

davidzhao:
Enable server to unpublish tracks. i.e. when a participant's publish permissions are revoked.

Also cleaned up participant permissions, and returning them to clients so each participant is aware of its and others' permissions.


hiroshihorie (approved):

livekit_models.proto
```diff
@@ -25,6 +25,19 @@ message Codec {
...
 
+message ParticipantPermission {
+  // allow participant to subscribe to other tracks in the room
+  bool can_subscribe = 1;
+  // allow participant to publish new tracks to room
+  bool can_publish = 2;
+  // allow participant to publish data
+  bool can_publish_data = 3;
+  // indicates that it's hidden to others
+  bool hidden = 7;
```
boks1971, line 36: gaps are due to deprecated stuff in between?

boks1971 (approved):

livekit_models.proto
davidzhao: yup, for backwards compatibility, field numbers cannot change.

PR #111, opened 2022-04-07
add target field to syncstate's datatrack info

cnderrauber:


cnderrauber:
will not merge before @boks1971 verified it works for him

boks1971 (approved):

PR #113, opened 2022-04-14
Ingress protobufs

davidzhao:


livekit_ingress.proto
```diff
@@ -0,0 +1,90 @@
...
+
+service Ingress {
+    rpc CreateIngress(CreateIngressRequest) returns (IngressInfo);
+    rpc ListIngress(ListIngressRequest) returns (ListIngressResponse);
+    rpc DeleteIngress(DeleteIngressRequest) returns (google.protobuf.Empty);
+}
+
+enum IngressInput {
+    RTMP_INPUT = 0;
+//    FILE_INPUT = 1;
```
cnderrauber, line 18: support in future?

livekit_ingress.proto
```diff
@@ -0,0 +1,90 @@
...
+}
+
+message IngressVideoOptions {
+    string name = 1;
+    TrackSource source = 2;
+    // desired mime_type to publish to room
+    string mime_type = 3;
+    // simulcast layers to publish, when empty, it'll pick default simulcast
+    // layers at 1/2 and 1/4 of the dimensions
+    repeated VideoLayer layers = 4;
```
boks1971, line 76: Guessing the idea is to encode simulcast at the bridge? Also, should we encode simulcast by default? Guess it does make sense especially to have layers if we are doing SVC.

livekit_ingress.proto
```diff
@@ -0,0 +1,90 @@
...
+
+import "google/protobuf/empty.proto";
+import "livekit_models.proto";
+
+option go_package = "github.com/livekit/protocol/livekit";
+
+service Ingress {
+    rpc CreateIngress(CreateIngressRequest) returns (IngressInfo);
+    rpc ListIngress(ListIngressRequest) returns (ListIngressResponse);
+    rpc DeleteIngress(DeleteIngressRequest) returns (google.protobuf.Empty);
```
boks1971, line 13: Should this have a response of success/failure at least?

cnderrauber (approved):

livekit_ingress.proto
```diff
@@ -0,0 +1,90 @@
...
+    bool disable_transcode = 7;
+}
+
+message IngressAudioOptions {
+    string name = 1;
+    TrackSource source = 2;
+    // desired mime_type to publish to room
+    string mime_type = 3;
+    uint32 bitrate = 4;
+    bool dtx = 5;
```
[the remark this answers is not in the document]
davidzhao: we'll transcode the input into a format supported by WebRTC.

livekit_ingress.proto
davidzhao: yeah, I think we should enable simulcast by default since it'll be needed to ensure uninterrupted streams for lower bandwidth users. 

The way to override is just to publish a single layer.

livekit_ingress.proto
davidzhao: failures would be signaled with a different status code in Twirp-land. (i.e. 500)

livekit_ingress.proto
davidzhao: yeah, will uncomment them when we start implementing them.

PR #114, opened 2022-04-14
update track and track composite

frostbyte73:


livekit_egress.proto
```diff
@@ -60,16 +60,20 @@ message TrackCompositeEgressRequest {
 message TrackEgressRequest {
   string room_name = 1;  // required
   string track_id = 2;   // required
+  bool containerize = 3; // does not work with websocket output
```
davidzhao, line 63: what does this do?

livekit_egress.proto
```diff
@@ -60,16 +60,20 @@ message TrackCompositeEgressRequest {
...
+    AzureBlobUpload azure = 6;
+    string websocket_url = 7;
   }
 }
 
 enum EncodedFileType {
-  MP4  = 0;
-  OGG = 1;
-//  WEBM = 2; coming soon
+  DEFAULT = 0;
```
davidzhao, line 73: some comments on what this is would be great!

davidzhao: and EncodedFileOutput.file_type would become optional then?

livekit_egress.proto
frostbyte73: I was confused, if you want containerization you'd use track composite

livekit_egress.proto
frostbyte73: yeah, and we'd choose it based on the input codecs and available options. Added comments for both

davidzhao (approved):

PR #116, opened 2022-04-19
Add RPCs, add default stream protocol

frostbyte73:


livekit_egress.proto
```diff
@@ -61,30 +61,36 @@ message TrackEgressRequest {
...
     S3Upload s3 = 3;
     GCPUpload gcp = 4;
     AzureBlobUpload azure = 5;
   }
 }
 
+message DirectFileOutput {
+  string filepath = 1; // (optional)
+  oneof output {       // required
+    S3Upload s3 = 3;
```
davidzhao, line 88:
```suggestion
    S3Upload s3 = 2;
```

davidzhao (approved):
lg, just one nit

PR #120, opened 2022-04-23
FPS field for VideoLayer

davidzhao:
Ingress uses this field during stream upload


livekit_models.proto
```diff
@@ -117,6 +117,8 @@ message VideoLayer {
   // target bitrate, server will measure actual
   uint32 bitrate = 4;
   uint32 ssrc = 5;
+  // publishing FPS for this layer, used with simulcasted streams
+  float fps = 6;
```
boks1971, line 121: thinking about using this with temporal layers?

boks1971: Why float?

boks1971 (approved):

livekit_models.proto
davidzhao: The intention is for ingress to reuse this structure when specifying desired layers. I think it'll be hard for clients to determine what fps temporal layers would have.

I had it as a float due to NTSC input streams having 29.97 fps. That's really the only reason :(. Any ideas on how we can better handle this case? GStreamer uses a fraction for fps. wdyt @cnderrauber @frostbyte73 ?

boks1971: Temporal layers are hard for server too :-) Have to add a frame rate measurement at temporal layer level to do it. I think fraction is fine. I just wanted to understand the reason behind a float. Not sure how exactly we can achieve fractional rate, but having it as a float is not an issue I think.

frostbyte73: the fraction is just as annoying imo, I had to do this to convert
https://github.com/livekit/server-sdk-go/blob/media/pkg/media/framerate/framerate.go

frostbyte73 (approved):

davidzhao:
closed in favor of #308

boks1971:
oh wow, totally forgot about this. Maybe, I should change #308 to float too.

PR #124, opened 2022-05-10
Node stats to keep track of retransmissions.

davidzhao:


livekit_internal.proto
```diff
@@ -64,6 +64,12 @@ message NodeStats {
   float load_avg_last5min = 19;
   float load_avg_last15min = 20;
   float cpu_load = 21;
+
+  // retransmissions
+  uint64 retransmit_bytes_out = 22;
+  uint64 retransmit_packets_out = 23;
+  float retransmit_bytes_out_per_sec = 24;
+  float retransmit_packets_out_per_sec = 25;
```
boks1971, line 72: should we also think about in bound retransmit stats?

boks1971 (approved):

livekit_internal.proto
boks1971: You mean the inbound retransmission stats are not actionable? We can tweak NACKs to control that a bit. But, it is not a significant amount. So, fine leaving it out for now.

davidzhao: Yeah, that's what I meant.. it seems that it wouldn't have a huge impact today. We can revisit and add in the future if needed.

PR #132, opened 2022-06-07
Add Segments info and support for SegmentedFileOutput in Room Composite Egress

biglittlebigben:
- Add Segments info to return information about the generated segment files in Egress status updates.
- Add support for SegmentedFileOutput in Room Composite Egress


livekit_egress.proto
```diff
@@ -234,3 +236,10 @@ message FileInfo {
   int64 size = 4;
   string location = 5;
 }
+
+message SegmentsInfo {
+  string playlist_filename = 1;
```
frostbyte73, line 241: playlist_name to match SegmentedFileOutput

frostbyte73 (approved):
Should we add a segment_count/num_segments as well?

livekit_egress.proto
biglittlebigben: Good idea.

PR #136, opened 2022-06-16
publish new codec when subscriber need

cnderrauber:


boks1971 (approved):

livekit_rtc.proto
```diff
@@ -101,6 +101,9 @@ message AddTrackRequest {
   repeated VideoLayer layers = 9;
 
   repeated SimulcastCodec simulcast_codecs = 10;
+
+  // server ID of track, publish new codec to exist track
+  string sid = 11;
```
cnderrauber, line 106: For new track, left sid as empty. For publish new codec to exist track, sid is the server side track id of the track

PR #143, opened 2022-07-06
Ingress updates

frostbyte73:


biglittlebigben:
This seems to go back to a model where the ingress is tied to a room at creation time? Could you walk me though an ingest lifecycle with this change?

frostbyte73:
> This seems to go back to a model where the ingress is tied to a room at creation time? Could you walk me though an ingest lifecycle with this change?

1. CreateIngress request sent from Server SDK -> LiveKit Server
2. LiveKit Server posts request using RPC client
3. Ingress instance accepts request and returns generated endpoint
4. Ingress instance prepares gstreamer pipeline and joins room
   1. Gstreamer rtmp2src -> decode -> encode -> appsink
   2. Appsink sends data to LiveKit Server using Go SDK
5. Once data is received, set pipeline to playing
6. Stream ends on StopIngress or EOS

frostbyte73:
For now it is a single room only, we can change it back later but my goal is just to get an mvp running so we can start uncovering issues and parallelizing the work

biglittlebigben:
> For now it is a single room only, we can change it back later but my goal is just to get an mvp running so we can start uncovering issues and parallelizing the work

Sure. We can do take this shortcut for now. However, we may want to keep in mind the requirement of having ingress disassociated from any room as we design the system.

biglittlebigben (approved):

frostbyte73:
> > For now it is a single room only, we can change it back later but my goal is just to get an mvp running so we can start uncovering issues and parallelizing the work
> 
> Sure. We can do take this shortcut for now. However, we may want to keep in mind the requirement of having ingress disassociated from any room as we design the system.

I appreciate the diligence, definitely don't want to make any decisions that become difficult to reverse

PR #145, opened 2022-07-10
Make sure we do not override the egressid when forwarding messages from the controller

biglittlebigben:
This allows to have consistent Egress IDs across API calls, and solves UpdateLayout failures


frostbyte73:
This only affects StartEgressRequest, how does this fix update layout?

frostbyte73 (approved):

PR #154, opened 2022-07-15
Packet stats from TC

matkam:


livekit_internal.proto
```diff
@@ -63,6 +63,11 @@ message NodeStats {
   float load_avg_last5min = 19;
   float load_avg_last15min = 20;
   float cpu_load = 21;
+  uint32 sys_packets_out = 28;
```
davidzhao, line 66: how does this differ from `packets_out` ?

boks1971: I think it will include everything at network stack/interface level.

davidzhao (approved):

real-danm (approved):

PR #175, opened 2022-07-26
Ingress proto updates

frostbyte73:


livekit_rpc_internal.proto
```diff
@@ -64,7 +64,10 @@ message IngressRequest {
   string request_id = 2;
   string sender_id = 3;
 
-  StopIngressRequest stop = 4;
+  oneof request {
+    UpdateIngressRequest update = 4;
+    DeleteIngressRequest stop = 5;
```
biglittlebigben, line 69: Should we rename the member from "stop" to "delete"?

biglittlebigben (approved):
Just one typo below

PR #176, opened 2022-07-27
Allow to request ingress info through RPC

biglittlebigben:
This creates a "RPC" interface that combines both client and server interfaces, and merges the client and server constructors as both sides of the RPC now act as client and server. This also adds a new channel getter for entity requests.


ingress/rpc.go
```diff
@@ -44,14 +45,21 @@ type RPCServer interface {
...
+	RPCServer
 }
 
 type RedisRPC struct {
 	nodeID livekit.NodeID
 	bus    *utils.RedisMessageBus
 }
 
-func NewRedisRPCClient(nodeID livekit.NodeID, rc *redis.Client) RPCClient {
+func NewRedisRPC(nodeID livekit.NodeID, rc *redis.Client) RPC {
```
frostbyte73, line 62: Let's remove NewRedisRPCServer since they're being combined

frostbyte73: Or, add SendRequest to RPCServer and SendResponse to RPCClient

ingress/rpc.go
```diff
@@ -44,14 +45,21 @@ type RPCServer interface {
 	SendResponse(ctx context.Context, request proto.Message, info *livekit.IngressInfo, err error) error
 	// SendUpdate sends an ingress info update
 	SendUpdate(ctx context.Context, info *livekit.IngressInfo) error
+	// GetEntityChannel returns a subscription for entity requests
+	GetEntityChannel(ctx context.Context) (utils.PubSub, error)
```
frostbyte73, line 49: This should be part of RPCClient

ingress/rpc.go
biglittlebigben: I initially removed it, but ended up adding it back specifically for the ingress handler program, since the current way of allocating node ids would cause it to share a node with the main service. Sharing this ID would not be a problem now, since the handler does not send requests, but would if it ever does.

frostbyte73 (approved):

PR #180, opened 2022-08-04
Add ping pong and pingtimeout for ws connection

lukasIO:


boks1971 (approved):
Let's confirm with @davidzhao about direction and impact for other clients.

livekit_rtc.proto
```diff
@@ -33,6 +33,8 @@ message SignalRequest {
     SyncState sync_state = 12;
     // Simulate conditions, for client validations
     SimulateScenario simulate = 13;
+    // client triggered ping to server
+    Ping ping = 14;
```
davidzhao, line 37: what about sending an int64 instead?
```suggestion
    int64 ping = 14;
```

that way when pong is received, we could compute delta on time spent.

livekit_rtc.proto
```diff
@@ -71,6 +73,8 @@ message SignalResponse {
     string refresh_token = 16;
     // server initiated track unpublish
     TrackUnpublishedResponse track_unpublished = 17;
+    // respond to ping
+    bool pong = 18;
```
davidzhao, line 77: same as above

davidzhao:
Updated JS client generation to use union types so it's no longer required to wrap within another type.

PR #185, opened 2022-08-10
Consolidate server information under ServerInfo

davidzhao:
We had sent several pieces of information back from the server before, consolidating together under ServerInfo. Also sending back software edition to help distinguish Cloud for informational purposes.


livekit_models.proto
```diff
@@ -173,6 +173,20 @@ message ParticipantTracks {
...
+// details about the server
+message ServerInfo {
+  enum Edition {
+    Standard = 0;
+    Cloud = 1;
+  }
+  Edition edition = 1;
+  string version = 2;
+  int32 protocol = 3;
+  string region = 4;
```
boks1971, line 185: can we also node_id?

boks1971 (approved):

livekit_models.proto
davidzhao: sure!

lukasIO (approved):

PR #186, opened 2022-08-12
Added connection type and network info

davidzhao:


livekit_analytics.proto
```diff
@@ -83,6 +83,10 @@ message AnalyticsClientMeta {
   string node = 2;
   string client_addr = 3;
   uint32 client_connect_time = 4;
+  // udp, tcp, turn
+  string connection_type = 5;
+  // wifi, cell, empty if not known
```
boks1971, line 89: Chrome at least reports `wired` (don't remember the name), `vpn` (this could be quite useful). This is just a string and we can store those types. But, just making a note about the comment here. Not sure if other browsers provide `vpn` detection.

boks1971 (approved):

livekit_analytics.proto
davidzhao: yeah will change comment.

Other browsers do not expose connection type, so it'll be Chrome + our mobile clients.

boks1971 (approved):

PR #196, opened 2022-08-26
Stream status and egress updated

frostbyte73:


livekit_egress.proto
```diff
@@ -230,8 +230,17 @@ message StreamInfoList {
...
 
 message StreamInfo {
+  enum Status {
+    ACTIVE = 0;
+    FINISHED = 1;
+    FAILED = 2;
+  }
+
   string url = 1;
+  int64 started_at = 2;
```
davidzhao, line 240: we had a gap here? what were the fields used for before?

davidzhao (approved):

biglittlebigben (approved):

livekit_egress.proto
frostbyte73: started_at and ended_at, which I removed a while back for duration. Would probably be helpful though since the streams can have different start and end times

PR #207, opened 2022-09-19
Create RoomInternal pb

frostbyte73:


davidzhao (approved):
a comment around RoomInternal would be helpful, otherwise lgtm

PR #209, opened 2022-09-19
Remove template base from pc request

frostbyte73:


davidzhao (approved):
should we adjust the protobuf ids?

PR #216, opened 2022-09-29
Moved AutoEgress and RoomInternal definitions

davidzhao:
Clients do not need these definitions. livekit_models also shouldn't have a dependency on livekit_egress


hiroshihorie (approved):
🙏

PR #230, opened 2022-10-18
add fps field to Subscribe Setting

cnderrauber:
The fps field might not best for the temporal layer request. Unlike resolution, I think it is hard to say I want how may frames per second, 30 , 15 ,10?
Other options is:
1. use `temporal_layer` field. but other fields are not svc specified, the subscriber should aware of svc and combine use resolution and temporal layer to request a special svc decode target.
2. import a new `VideoFluency` definition with `Low`, `Medium`, `High` values.

Or do you have other suggestions? @davidzhao @boks1971


boks1971 (commented):
@cnderrauber I think I prefer the `VideoFluency` option. If we leave it at `fps`, we will have to add doc that closest available temporal layer will be  subscribed to and that may not match the requested number.

cnderrauber:
> @cnderrauber I think I prefer the `VideoFluency` option. If we leave it at `fps`, we will have to add doc that closest available temporal layer will be subscribed to and that may not match the requested number.

Sounds good, I think `VideoFluency` is better that a detail fps number too, so will start at that, but open to change before entire feature completed.

PR #233, opened 2022-10-20
Move video output validation to protocol

biglittlebigben:


frostbyte73 (approved):

frostbyte73:
is this going to be reused?

biglittlebigben:
> is this going to be reused?

I should have given more context in the PR: I'm going to use it in cloud-io as well, at the same place where I'm testing the max resolution to fail earlier. I don't want to move it out of ingress both as a safeguard and for OSS.

PR #241, opened 2022-10-26
Support redis cluster

tomxiong:
1. add cluster_addresses property of config for redis cluster addresses
2.  use redis..UniversalClient instead of single-node client, FailoverClient and ClustertClient  follow the recommend of Go Redis: https://redis.uptrace.dev/guide/universal.html, it will simply the code to support the different Redis mode: Sentinel and Cluster

@frostbyte73 , please help to review it.


davidzhao (approved):
thank you for the PR! change LGTM.

we'd want to update docs in livekit-server and egress when integrating this change.

PR #249, opened 2022-11-11
Fallback to osstat if a matching cgroup implementation cannot be found, and fix core count computation

biglittlebigben:
This should fix WSL2 support.


davidzhao (approved):
lgtm

PR #257, opened 2022-12-09
Support for participant-level log sampling

davidzhao:
Also includes a cleanup of our logging system. Instead of using logr abstraction, we've created our own. It's still possible to use with an existing LogR logger, but by default it'll use Zap directly.

This has an added benefit of supporting Warn level logs. logr does not have a concept of warn, and it's quite annoying.

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -120,7 +120,7 @@ func (l *ZapLogger) Warnw(msg string, err error, keysAndValues ...interface{}) {
 	if err != nil {
 		keysAndValues = append(keysAndValues, "error", err)
 	}
-	l.zap.Errorw(msg, keysAndValues...)
+	l.zap.Warnw(msg, keysAndValues...)
 }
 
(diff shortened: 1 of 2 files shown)


logger/logger.go
```diff
@@ -86,30 +88,124 @@ func ParseZapLevel(level string) zapcore.Level {
...
+
+func (l *ZapLogger) Infow(msg string, keysAndValues ...interface{}) {
+	l.zap.Infow(msg, keysAndValues...)
+}
+
+func (l *ZapLogger) Warnw(msg string, err error, keysAndValues ...interface{}) {
+	if err != nil {
+		keysAndValues = append(keysAndValues, "error", err)
+	}
+	l.zap.Errorw(msg, keysAndValues...)
```
boks1971, line 123: Do we want `Warnw` here?

boks1971 (approved):

cnderrauber (approved):

logger/logger.go
davidzhao: yeesh. great catch.

PR #265, opened 2022-12-27
Add encryption field to TrackInfo

lukasIO:
In preparation for e2ee support.


livekit_models.proto
```diff
@@ -79,6 +79,12 @@ enum TrackSource {
   SCREEN_SHARE_AUDIO = 4;
 }
 
+enum EncryptionType {
+  NONE = 0;
```
davidzhao, line 83: I've learned recently that enum values are declared at the top level, vs scoped to the enum type. https://stackoverflow.com/questions/68134072/different-enums-same-value-in-protobuf

with that, it might be better to use `ET_NONE`, etc for the values.

davidzhao (approved):

livekit_models.proto
lukasIO: oh wow. yeah, makes sense!

lukasIO: what do you think about nesting it within a message? No strong preference either way, but `ET_*` doesn't feel very obvious. 
was thinking something like
```
message Encryption{
 enum Type{
     NONE=0;
     GCM=1;
     CUSTOM=2;    
 }
}
```

davidzhao: sure! that makes sense.

livekit_models.proto
```diff
@@ -79,6 +79,14 @@ enum TrackSource {
   SCREEN_SHARE_AUDIO = 4;
 }
 
+message Encryption {
+  enum Type {
+      NONE=0;
+      GCM=1;
+      CUSTOM=2;    
+  }
+ }
```
davidzhao, line 88:
```suggestion
}
```

PR #274, opened 2023-01-19
add TrackSource to AnalyticsStat

paulwe:


livekit_analytics.proto
```diff
@@ -58,6 +58,7 @@ message AnalyticsStat {
   float score = 9;
   repeated AnalyticsStream streams = 10;
   string mime = 11;
+  TrackSource source = 12;
```
boks1971, line 61: made a note about adding `TrackType` to the prom labels. Maybe, that is also needed here.

Or another option is to not add either field here (`source` and `trackType`) and just change telemetry APIs to take a `TrackSource`, `TrackType` (or maybe just `TrackInfo`) arguments and do it that way. As the `TrackSource`, `TrackType` is needed for internal use only, would be good to not add fields to the protocol.

PR #275, opened 2023-01-21
participant_rtc_init stats

real-danm:


livekit_internal.proto
```diff
@@ -92,8 +92,10 @@ message NodeStats {
   float participant_join_per_sec = 27;
   uint64 participant_rtc = 44;
```
boks1971, line 93: Rename these to `participant_rtc_connected_*` to make the difference between `init` and `connected` clear please?

boks1971 (approved):

PR #278, opened 2023-01-27
add lock tracker

paulwe:
track the time locks are attempted/granted so we can scan for hung locks without stopping the world to print stack dumps

### DIFF
--- a/utils/lock_tracker.go
+++ b/utils/lock_tracker.go
@@ -1,0 +1,224 @@
+package utils
+
+import (
+	"math"
+	"runtime"
+	"sync"
+	"sync/atomic"
+	"time"
+	"unsafe"
+)
+
+var lowResTime uint32 = uint32(time.Now().Unix())
+var weakRefs []uintptr
+var weakRefFree []int
+var weakRefLock sync.Mutex
+
+var startLockTrackerWorkerOnce sync.Once
+
+func StartLockTrackerWorker() {
+	startLockTrackerWorkerOnce.Do(func() {
+		go updateLowResTime()
+	})
+}
+
+func updateLowResTime() {
+	ticker := time.NewTicker(time.Second)
+	for t := range ticker.C {
+		atomic.StoreUint32(&lowResTime, uint32(t.Unix()))
+	}
+}
+
+func SyncLowResTime() {
+	atomic.StoreUint32(&lowResTime, uint32(time.Now().Unix()))
+}
+
+func NumMutexes() int {
+	weakRefLock.Lock()
+	defer weakRefLock.Unlock()
+	return len(weakRefs) - len(weakRefFree)
+}
+
+// ScanTrackedLocks check all lock trackers
+func ScanTrackedLocks(threshold time.Duration) bool {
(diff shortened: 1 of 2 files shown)


boks1971 (approved):
lgtm (I did not understand all the go magic in lazy init code 😄 ) @paulwe . A few comments/questions
- Is the intention to use this for debugging only or use it in live code? What do the benchmark numbers look like compared to vanilla golang mutex?
- If in live code, is the intention to do health checks only or also be able to show the deadlock paths?
- If the goal is to show deadlock paths also, guess we need some more metadata like line/file to figure out who is holding the lock. Wondering if it will be useful to add deadlock detection here as well so that we can get indication of deadlock path. Is that possible to add here?

paulwe:
- this is intended to be used in live code but not in any hot paths. the first call to `Lock` costs more than native go mutexes because of the tracker init but after that the difference is negligible.
- in live code this will be implemented as a health check and after we detect an issue hopefully we can find the cause in the stack dump. the goal is to quickly detect this type of issue and migrate users away from the unhealthy node.
- loading the program counters from `runtime.Callers` adds about 350ns of overhead... so relatively it's pretty bad but it might be worth it if we see more cases where other tools don't give us enough information to figure out what went wrong.
```
BenchmarkLockTracker
BenchmarkLockTracker/wrapped_mutex
BenchmarkLockTracker/wrapped_mutex-10         	85697451	        13.56 ns/op	       0 B/op	       0 allocs/op
BenchmarkLockTracker/native_mutex
BenchmarkLockTracker/native_mutex-10          	86315665	        13.44 ns/op	       0 B/op	       0 allocs/op
BenchmarkLockTracker/wrapped_rwmutex
BenchmarkLockTracker/wrapped_rwmutex-10       	64375668	        19.21 ns/op	       0 B/op	       0 allocs/op
BenchmarkLockTracker/native_rwmutex
BenchmarkLockTracker/native_rwmutex-10        	64665481	        18.59 ns/op	       0 B/op	       0 allocs/op
BenchmarkLockTracker/wrapped_mutex_init
BenchmarkLockTracker/wrapped_mutex_init-10    	 2870420	       430.3 ns/op	      36 B/op	       2 allocs/op
BenchmarkLockTracker/native_mutex_init
BenchmarkLockTracker/native_mutex_init-10     	76249582	        15.59 ns/op	       8 B/op	       1 allocs/op
BenchmarkLockTracker/wrapped_rwmutex_init
BenchmarkLockTracker/wrapped_rwmutex_init-10  	 2681212	       478.1 ns/op	      48 B/op	       2 allocs/op
BenchmarkLockTracker/native_rwmutex_init
BenchmarkLockTracker/native_rwmutex_init-10   	45549583	        25.12 ns/op	      24 B/op	       1 allocs/op
```

boks1971:
Got it. Thank you @paulwe

paulwe:
@boks1971 i might be being overly conservative. once we deploy this and nothing blows up i think we'll feel more confident adding more diagnostic information.

boks1971:
Not at all. Being cautious is great 😃 . We can iterate and add more functionality as necessary.

PR #280, opened 2023-01-27
add norace directive to scanTrackedLocks

paulwe:
race detector panics if we access an object that is being gced. it is safe here because until the finalizer finishes the `lockTracker` won't be freed and finalizer and scan functions all hold `weakRefLock`.

also deleting the parallel implementation of scan because it complicates this issue and incremental scanning seems preferable given the small number of locks and latency requirements.


davidzhao (approved):
lg

boks1971 (approved):

PR #282, opened 2023-01-29
Source specific permissions and protocol to update participant's metadata

davidzhao:


boks1971 (approved):

lukasIO:
should there be a similar distinction for `canSubscribe`?

lukasIO (approved):

davidzhao:
> should there be a similar granular distinction for `canSubscribe`?

hmm that might get pretty complicated. What would be the use case for that?

lukasIO:
Did not have a particular use case in mind, was more wondering about consistency between the two.

I guess a similar effect (if desired) can be achieved via track permissions as well anyways.

PR #287, opened 2023-02-01
Allow multiple outputs

frostbyte73:


livekit_egress.proto
```diff
@@ -32,31 +32,56 @@ message RoomCompositeEgressRequest {
...
+  oneof output {                // deprecated
+    EncodedFileOutput file = 6 [deprecated = true];
+    StreamOutput stream = 7 [deprecated = true];
+    SegmentedFileOutput segments = 10 [deprecated = true];
   }
   oneof options {
     EncodingOptionsPreset preset = 8; // (default H264_720P_30)
     EncodingOptions advanced = 9;     // (optional)
   }
+  EncodedFileOutput file_output = 11;
```
davidzhao, line 44: do we want to allow multiple of these too? i.e. write to two different file outputs? might be overkill?

frostbyte73: probably overkill, since the encoding parameters and output codecs need to be the same. Would just be making 2 copies of the same file

livekit_egress.proto
```diff
@@ -32,31 +32,56 @@ message RoomCompositeEgressRequest {
   bool audio_only = 3;          // (default false)
   bool video_only = 4;          // (default false)
   string custom_base_url = 5;   // template base url (default https://recorder.livekit.io)
-  oneof output {                // required
-    EncodedFileOutput file = 6;
-    StreamOutput stream = 7;
-    SegmentedFileOutput segments = 10;
+  oneof output {                // deprecated (use _output fields)
```
davidzhao, line 35: nit: let's indicate which server/egress version they will be deprecated in. so it'd be clear what's happening.

davidzhao (approved):
lg!

livekit_egress.proto
biglittlebigben: How about the case where we'd want the file uploaded to more than 1 place?

livekit_egress.proto
frostbyte73: will add once it's ready to go into livekit/egress

biglittlebigben:
I really think that by allowing only 1 output of each kind, we'll end up changing this again in the future when somebody comes up with a use case where more than 1 file or segmented output is needed, either because of a different muxer, segment format (HLS vs DASH), or upload location.

biglittlebigben (approved):

PR #291, opened 2023-02-02
reduce number of goroutines spawned during testing

paulwe:
with the new test config there seems to be a much stricter limit on the number of goroutines so this test became flaky


boks1971 (approved):
Which new test config are you referring to @paulwe ?

paulwe:
@boks1971 i synced the test options from livekit https://github.com/livekit/protocol/commit/c71216774a62f99f92572d48d20dcf6eb9a304f7#diff-fd32537cacb5ecf2cc211897a3e4cf382fd350672de01dd5bbb66bee6ef7b754R42

PR #292, opened 2023-02-03
Include room, participant and other_participants in reconnect response

lukasIO:
In order to be able to pick up 
- local participant metadata / info changes
- remote participants disconnecting / connecting 
- room metadata changes

that are happening while a client's signal connection is reconnecting


boks1971 (approved):

davidzhao:
Main question is around compatibility. Because existing clients do not support this, the server should already be sending back participant updates individually. Should we use existing signal messages to deliver this?

boks1971:
I guess that is a possibility. On a resume, server is sending a participant update. It can also add a room update which should be handled by existing clients.

lukasIO:
I think currently the participant update does not include remote participants that disconnected during the local user reconnecting, right?
If we want to use the standard participant update flow in order to synthesize disconnect events client size, the disconnected participants would need to be included with their latest state (`ParticipantInfo_State.DISCONNECTED`). 
If this seems feasible from the server side, then I agree that this sounds preferable in terms of compatibility!

lukasIO:
The difference with the participants being sent in the reconnect response is that we can do this diff on the client side (which is implemented in https://github.com/livekit/client-sdk-js/pull/572).
I assume we cannot do this diff all the time as the participant updates only include the participants that actually changed?

boks1971:
Oh yeah, brain fade on my part. Sorry @lukasIO . Makes sense.

It is not possible for server to definitively detect that it has caught all the cases as it is not possible for server to know when the client exactly lost connectivity and what was the last update it received. Even if we were to take an attempt, I think it will be significantly complex and error prone. I think `ReconnectResponse` containing the full picture is the simpler way.

davidzhao:
I'm not certain we can guarantee to deliver the full list of participants upon reconnect. For Cloud, each server is not guaranteed to have the full picture when new participant connects.

boks1971:
But, it is just like `JoinResponse`. Isn't it? Whatever the available list is at the time of connect, it is sent down. And subsequent join/leaves are sent as separate update.

We will have to make sure that the critical section ensures that all participants are either caught in `ReconnectResponse` or a subsequent update.

lukasIO:
In order to minimise the compatibility impact, we could only send down `other_participants` in the reconnect response and drop `room` and `participant`, as we can handle that via the standard update messages.

davidzhao:
> But, it is just like `JoinResponse`. Isn't it? Whatever the available list is at the time of connect, it is sent down. And subsequent join/leaves are sent as separate update.
> 
> We will have to make sure that the critical section ensures that all participants are either caught in `ReconnectResponse` or a subsequent update.

Yes, delivering participant changes in separate updates is as expected in the protocol. The part I think will cause problems is using `other_participants` to synthesize who has left the room. We do not perform that step with JoinResponse.

To put more explicitly, at no point should clients assume what's contained in `other_participants` includes everyone in the room.

Because of that limitation, there is no difference between sending down participant updates via the existing `ParticipantUpdate` message versus including it in here. So I don't think we should introduce a new signal message in this case, but instead we can think about how to send to the clients updates of participants have disconnected since they started reconnecting.

boks1971:
That's a solid consistent implementation of that message. Would be great to stick to that implementation under all conditions.

I think the real tricky bit will be to know when the reconnect actually started. Maybe, the client sends other participants it knows about in SyncState? Server can then synthesise a DISCONNECTED state for some participants?

davidzhao:
Sending a lot of participants up to the server may be inefficient. Maybe we can keep track of when the user was last active? something like:
* server to keep track of when we've received a ping message while the participant is connected (lastActiveTime)
* synthesize disconnected message for all others who's left since (lastActiveTime - N) seconds ago.

PR #296, opened 2023-02-09
Add egress_id option to list egress

frostbyte73:


livekit_egress.proto
```diff
@@ -223,7 +223,8 @@ message UpdateStreamRequest {
 }
 
 message ListEgressRequest {
-  string room_name = 1; // (optional, used to filter results)
+  string room_name = 1; // (optional, filter by room name)
+  string egress_id = 2; // (optional, filter by egress ID)
```
biglittlebigben, line 227: Is there a use case for providing a list of IDs instead of 1?

biglittlebigben (approved):

livekit_egress.proto
frostbyte73: can always add it later if people ask for it. Same with filtering by egress type or status

PR #297, opened 2023-02-12
Fix support for Redis cluster mode, updated to v9

davidzhao:


davidzhao:
this PR also raised the min supported version of Go to 1.18, as is required by go-redis/v9

boks1971 (approved):

PR #299, opened 2023-02-14
Allow using the 1st sample date/time as suffix when using the egress segmented output

biglittlebigben:


livekit_egress.proto
```diff
@@ -117,14 +117,21 @@ enum SegmentedFileProtocol {
   HLS_PROTOCOL = 1;
 }
 
+enum SegmentedFileSuffix {
+  DEFAULT_SEGMENTED_FILE_SUFFIX = 0;
```
frostbyte73, line 121: let's leave INDEX as 0 and remove default. We use default in the other enums because the default might change depending on the source

frostbyte73 (approved):

PR #305, opened 2023-02-18
move TimedVersion from livekit

paulwe:
this is backward compatible with the existing implementations but can work without allocating (>10% of allocated objects in syncstore are TimedVersions)

### DIFF
--- a/utils/timed_version.go
+++ b/utils/timed_version.go
@@ -2,9 +2,10 @@ package utils
 
 import (
 	"fmt"
-	"sync/atomic"
 	"time"
 	_ "unsafe" // required for linkname
+
+	"go.uber.org/atomic"
 
 	"github.com/livekit/protocol/livekit"
 )
@@ -14,8 +15,6 @@ func nanotime1() int64
 
 //go:linkname usleep runtime.usleep
 func usleep(usec uint32)
-
-type nocmp [0]func()
 
 const tickBits uint64 = 16
 const tickMask uint64 = 0xffff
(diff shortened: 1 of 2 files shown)


utils/timed_version.go
```diff
@@ -0,0 +1,142 @@
+package utils
+
+import (
+	"fmt"
+	"sync/atomic"
```
boks1971, line 5: We have replaced used of this with https://pkg.go.dev/go.uber.org/atomic everywhere. Is it possible to use the uber atomic package here?

utils/timed_version.go
```diff
@@ -0,0 +1,142 @@
...
+	"time"
+	_ "unsafe"
```
boks1971, line 7: Is this used?

paulwe: it's required for linkname to get the monotonic time function from runtime

utils/timed_version.go
paulwe: i added nocmp below so it has the safety and ergonomics as a memeber of the uber atomics family. using that to implement it would make it messier though

utils/timed_version.go
paulwe: added a comment to clarify

utils/timed_version.go
```diff
@@ -0,0 +1,150 @@
...
+		} else {
+			g.ts = now
+			g.ticks = 0
+		}
+		return TimedVersion{v: *atomic.NewUint64(uint64(now)<<tickBits | g.ticks)}
+	}
+}
+
+type TimedVersion struct {
+	v atomic.Uint64
```
davidzhao, line 81: neat way of packing these

davidzhao (approved):
lgtm!

utils/timed_version_test.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+func TestTimedVersion(t *testing.T) {
+	t.Run("timed versions are monotonic and comparable", func(t *testing.T) {
+		gen := NewDefaultTimedVersionGenerator()
+		tv1 := gen.New()
+		tv2 := gen.New()
+		tv3 := gen.New()
+
+		assert.True(t, tv3.After(tv1))
```
davidzhao, line 16: any reason to use assert instead of require here? testify asserts do not terminate the test, but will let execution continue. I've found it to be confusing when tests are failing.

paulwe: i assumed you had a reason for using it in the original tests so i copied it

davidzhao: hah I didn't realize I did! probably typed it out due to muscle memory.

PR #314, opened 2023-03-03
Add topic to UserPacket

lukasIO:
Introduces a `topic` on `UserPacket` in order to make dataChannel usage more convenient on client and server SDKs


boks1971 (approved):

lukasIO:
thanks for the review!
wondering whether we should mark `topic` as `optional` in order to maximise backwards compatibility out of the box?

boks1971:
> thanks for the review! wondering whether we should mark `topic` as `optional` in order to maximise backwards compatibility out of the box?

Isn't everything optional in protobuf?

lukasIO:
I think without the _optional_ keyword the field will be populated with a default value, vs. the presence of the field being tracked explicitly. 
In the context of this `topic` that would mean an empty string. 
reference: https://github.com/protocolbuffers/protobuf/blob/v3.12.0/docs/field_presence.md#using-the-generated-code

I guess we can go with both options, the only downside being that a user defined topic with an empty string would be treated as no topic - which is ok, I guess?

boks1971:
In that case, I think `optional` is good.

frostbyte73:
We can't use optionals until protoc_gen_ruby is updated to support it:

```
livekit_models.proto: is a proto3 file that contains optional fields, but code generator protoc-gen-twirp_ruby hasn't been updated to support optional fields in proto3. Please ask the owner of this code generator to support proto3 optional.--twirp_ruby_out:
```

frostbyte73:
protoc-gen-twirp_ruby has been updated, ignore above

PR #316, opened 2023-03-06
Add can_update_metadata field to participant permissions

lukasIO:


livekit_models.proto
```diff
@@ -38,8 +38,8 @@ message ParticipantPermission {
   bool hidden = 7;
   // indicates it's a recorder instance
   bool recorder = 8;
-
-  // NEXT_ID: 10
+  // indicates that participant can update own metadata
+  bool can_update_metadata = 10;
```
boks1971, line 42: Would be good to put back the `NEXT_ID` with updated value. I know it makes most sense if the last value in the list is not the highest, but I find that a good practice to be explicit. The lesser conventions to remember, the better it is for me :-)

boks1971 (approved):

davidzhao (approved):
lgtm!

PR #318, opened 2023-03-09
bytes -> string gcp credentials

frostbyte73:


biglittlebigben:
Won't this break existing code?

frostbyte73:
> Won't this break existing code?

https://protobuf.dev/programming-guides/proto/#updating
`string and bytes are compatible as long as the bytes are valid UTF-8.`

frostbyte73:
the only difference in go should be that
```
&GCPUpload{
    Credentials: []byte("...")
}
```
changes to 
```
&GCPUpload{
    Credentials: "..."
}
```

davidzhao (approved):
lgtm, we should bump protocol by a minor version, should ensure that server SDKs have a minor version bump as well.

frostbyte73:
> > the only difference in go should be that
> > ```
> > &GCPUpload{
> >     Credentials: []byte("...")
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > changes to
> > ```
> > &GCPUpload{
> >     Credentials: "..."
> > }
> > ```
> 
> Right, I was asking about SDK compatibility rather than protocol. Do we have a process to document these changes that will require code updates from SDK users?

Probably just release notes, we do not have a documented process

PR #330, opened 2023-03-30
add clock offset to timed version generator

matkam:


davidzhao (approved):
lg

utils/timed_version.go
```diff
@@ -73,6 +77,10 @@ func (g *timedVersionGenerator) Next() TimedVersion {
 	}
 }
 
+func (g *timedVersionGenerator) SetClockOffset(clockOffset time.Duration) {
+	g.clockOffset = clockOffset
```
paulwe, line 81: this and the access in `Next()` need to guarded by `g.mu` or we will get flagged by race detector

davidzhao: oops.. I missed that.. good catch Paul!

PR #331, opened 2023-03-30
Adding a time series utility.

boks1971:
Also adding a `SWEEPER` node type.

Will add more comments in line.

Adding a time series utility module for keeping stats. Planning to use it in relay stats. I think it can be more generally usable. Could not find a good existing library. As it is simple enough and also have some unique requirements, just making one. But, if you know of any existing good ones, please let me know and I can change this and make this a wrapper.


utils/timeseries.go
```diff
@@ -0,0 +1,234 @@
...
+				}
+				break
+			}
+		}
+	}
+
+	t.prune()
+}
+
+func (t *TimeSeries[T]) prune() {
```
boks1971, line 214: currently doing `prune` on both read and write.

utils/timeseries.go
```diff
@@ -0,0 +1,234 @@
...
+	thresh := time.Now().Add(-t.params.Window)
+	for idx, s := range t.samples {
+		if s.At.After(thresh) {
+			t.samples = t.samples[idx:]
+			break
+		}
+	}
+}
+
+// TODO - a bunch of stats
```
boks1971, line 224: currently supporting only `min, max, sum`. We can add more over time as the need arises.

livekit_internal.proto
```diff
@@ -16,6 +16,7 @@ enum NodeType {
   CONTROLLER = 1;
   MEDIA = 2;
   TURN = 4;
+  SWEEPER = 5;
```
boks1971, line 19: Adding a node type which we can use in cloud.

paulwe:
i think we might be able to use gonum for this

utils/timeseries.go
```diff
@@ -0,0 +1,231 @@
...
+func (t *TimeSeries[T]) addSampleAt(val T, at time.Time) {
+	// insert in time order
+	switch {
+	case len(t.samples) == 0 || at.After(t.samples[len(t.samples)-1].At): // empty or at end
+		t.samples = append(t.samples, TimeSeriesSample[T]{
+			Value: val,
+			At:    at,
+		})
+
+	case at.Before(t.samples[0].At): // at start
```
cnderrauber, line 187: If there are many insert operation at start or middle to the samples, list might be better than slice.

cnderrauber (approved):

utils/timeseries.go
boks1971: Good point @cnderrauber . Will change to a list.

boks1971:
> i think we might be able to use gonum for this

I looked at that @paulwe , but could not find a time series in there. Will check for more options. We can always switch out later too.

PR #336, opened 2023-04-05
update psrpc

paulwe:


go.mod
```diff
@@ -4,20 +4,23 @@ go 1.18
 
 require (
 	github.com/eapache/channels v1.1.0
+	github.com/facebookincubator/go-qfext v0.0.0-20210831195807-4cd18e79853b
```
boks1971, line 7: Seems like this and the cuckoo filter below are getting picked up unintentionally?

boks1971 (approved):

go.mod
paulwe: thanks 👍

PR #339, opened 2023-04-07
Validate ingress info encoding parameters

biglittlebigben:


ingress/validation.go
```diff
@@ -4,22 +4,103 @@ import (
 	"github.com/livekit/protocol/livekit"
 )
 
-func ValidateVideoOptionsConsistency(options *livekit.IngressVideoEncodingOptions) error {
+// This validates that ingress options have no consistency issues and provide enough parameters
+// to be usable. Options that pass this test may still need some fields to be poulated with default values
+// before being used in a media pipeline.
+
+func Validate(info *livekit.IngressInfo) error {
+	if info.InputType != livekit.IngressInput_RTMP_INPUT {
```
davidzhao, line 12: can we default to RTMP here? it's the only supported type.

davidzhao (approved):

PR #345, opened 2023-04-12
Improve webhook resilience with retryablehttp

davidzhao:


frostbyte73 (approved):

theomonnom:
Thanks, just thinking out loud @davidzhao.
If we retry a request, will the other requests wait?
Asking because the events may be unordered (e.g receiving ParticipantLeft before ParticipantJoined)

PR #346, opened 2023-04-13
`SubscriberAllowPause` as a connection option.

boks1971:
This allows controlling stream allocator allow pause or not decision through a connection option.

NOTE: Thinking ahead a bit, would we allow pause or not at a track level? We are going to do subscription priority. Maybe, we can also allow pause or not at track level by adding it to `UpdateTrackSettings`. Wondering if things are going to be confusing if we do that also. I think the hierarchy would be (order of increasing precedence)
- from config
- connection option added in this PR will take effect if set
- track level allow pause config if we decide to do this (have to make sure that we can determine that the value was explicitly set in the protocol message vs using default value in protocol message).


paulwe (approved):

paulwe:
`go install github.com/livekit/psrpc/protoc-gen-psrpc@latest` then rerunning `mage proto` will probably fix this

lukasIO:
> We are going to do subscription priority. Maybe, we can also allow pause or not at track level by adding it to `UpdateTrackSettings`

Just thinking out loud whether this could be handled as part of subscription priority. E.g. priority `0` as the lowest value indicates that pausing is allowed, while the next higher priority of `1` would indicate that pausing is not allowed any more but it has a low  priority otherwise (=default). But the obvious downside being that there couldn't be different priorities for "pausible" streams.

boks1971:
> > We are going to do subscription priority. Maybe, we can also allow pause or not at track level by adding it to `UpdateTrackSettings`
> 
> Just thinking out loud whether this could be handled as part of subscription priority. E.g. priority `0` as the lowest value indicates that pausing is allowed, while the next higher priority of `1` would indicate that pausing is not allowed any more but it has a low priority otherwise (=default). But the obvious downside being that there couldn't be different priorities for "pausible" streams.

I thought about overloading priority to convey this too, but then I had a vision of @davidzhao rebuking me for making to too confusing to devs  and decided to not write it 😄 .

livekit_internal.proto
```diff
@@ -150,6 +150,7 @@ message StartSession {
   //if reconnect, client will set current sid
   string participant_id = 16;
   ReconnectReason reconnect_reason = 17;
+  bool subscriber_allow_pause = 18;
```
lukasIO, line 153: just as a question: is this a publisher side or a subscriber side setting? i.e. does this indicate `tracks published by this user are allowed to be paused on subscribers` or `allow all the tracks that this user subscribed to, to be paused`

lukasIO (approved):

livekit_internal.proto
boks1971: the latter @lukasIO . This is a participant telling that my down tracks are allowed to pause or not. Please let me know if you can think of a better name.

PR #347, opened 2023-04-13
Explicitly set if subscriber_allow_pause is valid

boks1971:
Need to know if `false` (which would be default if not set) is explicitly set or not.

Can you folks think of a better way to handle this? I don't like this. But, using a combination in cloud (using config, using a feature flag and then if explicitly set in connection option). With that combination, have to know if it was explicitly set in connect option.


lukasIO:
I think you could mark the field as `optional` instead, which would mean the presence of a set value would be tracked additionally instead of a default value being assigned.

boks1971:
> I think you could mark the field as `optional` instead, which would mean the presence of a set value would be tracked additionally instead of a default value being assigned.

Thank you @lukasIO , that works. Made the change in https://github.com/livekit/protocol/pull/347/commits/774e7939adb153a5d84b66b194f187172fdc3eec

boks1971:
> > I think you could mark the field as `optional` instead, which would mean the presence of a set value would be tracked additionally instead of a default value being assigned.
> 
> Thank you @lukasIO , that works. Made the change in [774e793](https://github.com/livekit/protocol/commit/774e7939adb153a5d84b66b194f187172fdc3eec)

Can you bonk it when you get a chance?

paulwe (approved):

PR #350, opened 2023-04-15
Revamped notifier to avoid out of order delivery

davidzhao:
Moving to a structure where each URL occupies their own delivery goroutine. This avoids blocking of notifications to other URLs.

We'll also include a field indicating how many messages were dropped.


webhook/url_notifier.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+		}
+	}
+}
+
+func (n *URLNotifier) processQueue() {
+	for event := n.nextItem(); event != nil && !n.fuse.IsBroken(); event = n.nextItem() {
+		if err := n.send(event); err != nil {
+			n.params.Logger.Warnw("failed to send webhook", err, "url", n.params.URL, "event", event.Event)
+			n.dropped.Inc()
```
paulwe, line 134: if `event.NumDropped` is nonzero `n.dropped` is off after failures

paulwe (approved):
lgtm 👍

webhook/url_notifier.go
davidzhao: good point.. i'll restore.

PR #354, opened 2023-04-21
Update psrpc

frostbyte73:


rpc/egress.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.30.0
+// 	protoc-gen-go v1.28.1
```
davidzhao, line 3: can we generate these on the server to avoid incompatible versions?

paulwe (approved):

biglittlebigben (approved):

PR #358, opened 2023-04-22
Added ProtoProxy util, NumPublishers field

davidzhao:
ProtoProxy achieves two goals:
- avoids needless regeneration of readonly protobufs
- update certain changes periodically


boks1971 (approved):

utils/protoproxy.go
```diff
@@ -0,0 +1,123 @@
...
+	return p
+}
+
+func (p *ProtoProxy[T]) MarkDirty(immediate bool) {
+	p.lock.Lock()
+	p.dirty = true
+	shouldUpdate := immediate || time.Since(p.refreshedAt) > p.refreshInterval
+	p.lock.Unlock()
+	if shouldUpdate {
+		go p.performUpdate(false)
```
boks1971, line 66: Could this make things out of order? Looks like it can. If two immediately race, it is possible that the order may be flipped and the older message gets latched?

paulwe: instead of performing the update immediately it could be enqueued on the worker?

paulwe (approved):

utils/protoproxy.go
davidzhao: good point.. let me queue it in the main thread instead.

PR #363, opened 2023-04-25
Add participants sids to syncstate

lukasIO:


boks1971 (approved):
Guessing this needs server side work?

boks1971:
> yeah, it would. main questions for me are:
> 
> * does every sync state request potentially trigger a participant update to be sent down (then this addition isn't even needed)
> * does the server reconcile whether the participant_sids are different to the one's in the room and only send an update in case they differ (then this PR would be needed)
> 
> because of current efforts to minimize participant updates, I thought it would be better to let the server decide

I forget the details now. @cnderrauber did some careful updating. He will know what exactly happens.

cnderrauber:
@lukasIO When client resume connection, the server side will send participant updates for all participants, do you observe a case that a update that should be sent down is missed?
This line will sent all participants to client
https://github.com/livekit/cloud/blob/d76a0a58c953c7fde33646cc5ecac54caab3be15/pkg/rtc/room_local.go#L269

This PR([#1495](https://github.com/livekit/livekit/pull/1495)) will send disconnected participant during reconnecting to the client

davidzhao:
I think even knowing the sids the client has access to, the server would still need to send updates to the client because updates could be:
* tracks added
* track muted/unmuted
* metadata changed
It seems that these won't be easy to track based on sid alone.

Perhaps it's possible to use versions and/or delivery sequence numbers to indicate what the client knows about.. For example, if every update is stamped with a delivery sequence, we could then allow the server to buffer some number of messages.. and then decide if it should resend everything, or only send missed messages in the buffer.

boks1971:
Yeah, a sequence number of timed version would be more generic. Server can maintain a queue and replay missed messages.

lukasIO:
> the server side will send participant updates for all participants, do you observe a case that a update that should be sent down is missed?
@cnderrauber yeah, I'll ping you with details!

lukasIO:
previous comment was incorrect, participant updates do come in, just client state doesn't trigger events for it, will fix this on the client side by caching participant sids during signal reconnect!

PR #374, opened 2023-05-08
update notifier

frostbyte73:


webhook/webhook_test.go
```diff
@@ -20,7 +20,7 @@ const (
 	apiSecret            = "mysecret"
 	testAddr             = ":8765"
 	testUrl              = "http://localhost:8765"
-	webhookCheckInterval = 100 * time.Millisecond
+	webhookCheckInterval = 500 * time.Millisecond
```
davidzhao, line 23: did this change have any significance?

davidzhao (approved):

webhook/webhook_test.go
frostbyte73: changed it while testing to be sure everything was finishing, I'll change it back

PR #376, opened 2023-05-12
add b62 encoding for guids

paulwe:


utils/id.go
```diff
@@ -41,3 +44,61 @@ func LocalNodeID() (string, error) {
...
+	}
+}
+
+func MarshalGuid[T livekit.Guid](id T) livekit.GuidBlock {
+	var b livekit.GuidBlock
+	prefix := guidPrefix[T]()
+	idb := []byte(id)[len(prefix):]
+	for i := 0; i < 12; i += 4 {
+		j := (i * 6) >> 3
+		b[j] |= b62Index[idb[i]]<<2 | b62Index[idb[i+1]]>>4
```
boks1971, line 83: I am probably not reading this properly. Is there a need to do `|=` here? Seems like each byte of `b` is just set once.

boks1971 (approved):

utils/id.go
paulwe: oh, good catch. it looks like i forgot to change this when i unrolled these

PR #377, opened 2023-05-12
add structured logging for protobuf messages

paulwe:


logger/proto.go
```diff
@@ -0,0 +1,135 @@
...
+			e.AddString(k, s)
+		}
+	case protoreflect.BytesKind:
+		if b := v.Bytes(); len(b) != 0 {
+			e.AddString(k, base64.RawStdEncoding.EncodeToString(b))
+		}
+	case protoreflect.MessageKind:
+		if m := v.Message(); m.IsValid() {
+			e.AddObject(k, protoMarshaller{m})
+		}
```
boks1971, line 133: Will these kinds change (more like augmented) with protobuf version? Is there a need to marshal a generic placeholder "unknown kind"?

boks1971 (approved):

logger/proto.go
paulwe: i don't think these change in major pb versions

PR #379, opened 2023-05-15
add event emitter

paulwe:


utils/event_emitter.go
```diff
@@ -0,0 +1,109 @@
...
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	return maps.Keys(e.observers)
+}
+
+func (e *EventEmitter[K, V]) stopObserving(k K, le *list.Element) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	l := e.observers[k]
```
boks1971, line 76: Check for nil l and log an error to guard against multiple stop calls. Just in case somebody calls it spuriously.

utils/event_emitter.go
```diff
@@ -0,0 +1,109 @@
...
+		stop: stopFunc,
+		ch:   make(chan V, DefaultEventEmitterParams.QeueSize),
+	}
+	return o, o.emit
+}
+
+func (o EventObserver[V]) emit(v V) {
+	select {
+	case o.ch <- v:
+	default:
```
boks1971, line 99: Add a log to indicate channel overflow. We have had cases of unexpected channel overflow ( there might still be cases out there) that needed tuning the size.

boks1971 (approved):

PR #385, opened 2023-05-22
Client report

lukasIO:


livekit_models.proto
```diff
@@ -364,3 +364,68 @@ enum ReconnectReason {
...
+  }
+
+  message ApiError {
+    int32 code = 1;
+    int64 timestamp = 2;
+    string description = 3;
+  }
+
+  message DeviceInfo {
+    string ua = 1;
```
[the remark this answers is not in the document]
lukasIO: makes sense. 
thinking more about this, it seems to me like `DeviceInfo` would also conceptually be a one-off message as it won't change over time. Even if we include other properties here, those will be hardware bound, so adding them one time when connection establishes probably makes the most sense. 
Do you agree to remove this entirely from here and rather add `concurrency` to the already existing `ClientInfo`?

livekit_models.proto
```diff
@@ -364,3 +364,68 @@ enum ReconnectReason {
...
+    // none, cpu, bandwidth, or other
+    string qualityLimitationReason = 1;
+    // maps qualitLimitationReason to duration the limitation has been experienced
+    map<string, uint32> qualityLimitationDurations = 2;
+    uint32 width = 3;
+    uint32 height = 4;
+  }
+
+  message RTCReceiverStats {
+    int32 jitter = 1;
```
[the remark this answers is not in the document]
lukasIO: it was meant as a snapshot. but maybe makes sense to sample multiple values over time until the next report gets sent out?

livekit_models.proto
```diff
@@ -364,3 +364,68 @@ enum ReconnectReason {
...
+    int32 height = 7;
+  }
+
+  message EventReport {
+    string eventName = 1;
+    string reason = 2;
+    int64 timestamp = 3;
+  }
+
+  message ApiError {
```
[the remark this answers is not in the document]
lukasIO: if they all share the same ClientEvent message, would it suffice to have a common `message` string as payload ? 
E.g. wondering if we want more details on a permission change

livekit_models.proto
```diff
@@ -364,3 +364,68 @@ enum ReconnectReason {
...
   RR_SWITCH_CANDIDATE = 4;
 }
+
+enum SubscriptionStatus {
+  Unsubscribed = 0;
+  Desired = 1;
+  Subscribed = 2;
+}
+
+message ClientReport {
```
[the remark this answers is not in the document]
lukasIO: reduced the report to only include stats for now.

livekit_models.proto
```diff
@@ -364,3 +364,68 @@ enum ReconnectReason {
...
+
+  message RTCSenderStats {
+    // none, cpu, bandwidth, or other
+    string qualityLimitationReason = 1;
```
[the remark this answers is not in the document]
lukasIO: we have `qualitylimitationDurations` directly from the stats, without dedicated events and only reading from stats, I guess that's the best we can do for now. We'll have the _current_ limitation reason regularly whenever we send a report.

livekit_models.proto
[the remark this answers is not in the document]
lukasIO: it's total jitter as defined [here](https://www.w3.org/TR/webrtc-stats/#dom-rtcreceivedrtpstreamstats-jitter)

livekit_models.proto
```diff
@@ -364,3 +364,68 @@ enum ReconnectReason {
...
+    int32 readyState = 3;
+    string label = 4;
+    bool muted = 5;
+  }
+
+  message SubscriptionReport {
+    string sid = 1;
+    repeated RTCReceiverStats rtcStats = 2;
+    int32 readyState = 3;
+    SubscriptionStatus subscriptionStatus = 4;
```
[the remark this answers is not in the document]
lukasIO: dropped all events for now

livekit_models.proto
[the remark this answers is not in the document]
lukasIO: only doing stats for now

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+    uint32 freeze_count = 9;
+    double total_freezes_duration = 10;
+    uint32 fir_count = 11;
+    string decoder_implementation = 12;
+    bool power_efficient_decoder = 13;
+  }
+
+  message RTCAudioReceiverStats {
+    RTCReceiverStats common_stats = 1;
+    uint32 concealed_samples = 2;
```
davidzhao, line 473: I think we'll also need `total_samples` for concealed samples to be useful.

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+   RTCSenderStats common_stats = 1;
+   // media-source
+   double total_audio_energy = 2;
+   double total_samples_duration = 3;
+  }
+
+  message RTCVideoSenderStats {
+   RTCSenderStats common_stats = 1;
+    // outbound-rtp
+    map<string, uint32> quality_limitation_durations = 2;
```
davidzhao, line 433: I'm not sure if webrtc-internals tracks this differently between layers. for some reasons I remember the duration being the same across layers.

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+
+message ClientReport {
+
+  message RTCSenderStats {
+    // ice-candidate-pair
+    uint32 available_outgoing_bitrate = 1;
+    // local-candidate
+    string networkType = 3;
+    // outbound-rtp
+    uint64 packetsSent = 4;
```
davidzhao, line 415: would suggest `snake_case` for field names. 

though I'm not sure if network level stats would be helpful, since we do get these on the server side. For v1, let's start with minimal stats?

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+  Subscribed = 2;
+}
+
 enum SubscriptionError {
   SE_UNKNOWN = 0;
   SE_CODEC_UNSUPPORTED = 1;
   SE_TRACK_NOTFOUND = 2;
 }
+
+message ClientReport {
```
davidzhao, line 407: perhaps we should move this outside of livekit_models? since it's used by a very specific request.. maybe into RoomService?

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+
+  message RTCSenderStats {
+    // ice-candidate-pair
+    uint32 available_outgoing_bitrate = 1;
```
davidzhao, line 411: this seems like at the transport layer, and doesn't fit too well within each sender

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+  }
+
+  message RTCVideoSenderStats {
+   RTCSenderStats common_stats = 1;
+    // outbound-rtp
+    map<string, uint32> quality_limitation_durations = 2;
+    string quality_limitation_reason = 3;
+    uint32 quality_limitation_resolution_changes = 4;
+    string scalability_mode = 5;
+    uint32 frame_height = 6;
```
davidzhao, line 437: which layer would dimension stats be referencing? I think we also already have these pieces of data on the server side.

livekit_models.proto
davidzhao: Is `RTCVideoSenderStats` per track? or per simulcast layer?

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+    RTCAudioSenderStats rtc_stats = 2;
+    int32 ready_state = 3;
+    string label = 4;
+    bool muted = 5;
+  }
+
+  message VideoPublicationReport {
+    string sid = 1;
+    RTCVideoSenderStats rtc_stats = 2;
+    int32 ready_state = 3;
```
davidzhao, line 487: trying to understand these a bit:
* ready_state - I think this is just live or ended. do we expect to be ended?
* label - this seems a better fit as an event, since we do not expect this to change?
* muted - is this tracking MediaStreamTrack.muted property ? i.e. has data, or is it looking at our mute property? would be good to use a diff name here if it's former.

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+    bool allowed = 5;
+  }
+
+  message VideoSubscriptionReport {
+    string sid = 1;
+    RTCVideoReceiverStats rtc_stats = 2;
+    int32 ready_state = 3;
+    SubscriptionStatus subscription_status = 4;
+    bool allowed = 5;
+    uint32 width = 6;
```
davidzhao, line 506: I think this should move into `RTCVideoReceiverStats` ? since the width and height of receiver can change dynamically. 

though I'm also not sure what value it is to keep this on the client.. since the server knows exactly what width/height is sent down

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+    RTCAudioReceiverStats rtc_stats = 2;
+    int32 ready_state = 3;
+    SubscriptionStatus subscription_status = 4;
+    bool allowed = 5;
+  }
+
+  message VideoSubscriptionReport {
+    string sid = 1;
+    RTCVideoReceiverStats rtc_stats = 2;
+    int32 ready_state = 3;
```
davidzhao, line 503: same question as above

livekit_models.proto
```diff
@@ -391,8 +392,124 @@ enum ReconnectReason {
...
+    // ice-candidate-pair
+    uint32 available_outgoing_bitrate = 1;
+    // local-candidate
+    string networkType = 3;
+    // outbound-rtp
+    uint64 packetsSent = 4;
+    uint32 packetsLost = 5;
+    uint32 nack_count = 6;    
+    uint32 retransmitted_packets_sent = 7;
+    uint64 timestamp = 8;
```
davidzhao, line 419: which timestamp is this referencing? I only noticed it in the sender stats.. do we need a similar timestamp in the receiver side too? maybe this should be moved into the `*Report` message itself?

livekit_models.proto
lukasIO: it's per sender, which should be equivalent to per track. 
I think you're right, that it will not track different layers, but how would we do that in the first place?

lukasIO: Regarding the type, it's a map that includes `reason` and `duration` and not different layers.

livekit_models.proto
lukasIO: missed snake case here. Will remove `packets_sent` entirely then.

livekit_models.proto
lukasIO: the original dimension. Makes sense to omit them, if we have them already.

livekit_models.proto
lukasIO: 1. we don't expect it to be `ended`, but it might be good debugging insight to know that for sure?
2. not sure why I wanted to include this in the first place, I guess we can drop it for v1 at least
3. it's looking at our `mute` property.

livekit_models.proto
lukasIO: you're right, we would have dimension in RTC stats as well. Seems like leaving them out is the way to go!

livekit_models.proto
lukasIO: `timestamp` is referring to 
> A [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp) object indicating the time at which the sample was taken for this statistics object.

For receiver stats I included `last_packet_received_timestamp`, but probably makes sense to add the processing timestamp, too!

livekit_models.proto
lukasIO: is your suggestion to leave it out entirely or to move it to a higher level, that's not sender related?

livekit_models.proto
lukasIO: will move it to RoomService as a last step, so that keeping track of the discussions is easier until then

PR #388, opened 2023-05-26
fix numCPU (open /sys/fs/cgroup/cpu.max: no such file or directory)

m4ex:
If we try to run ingress app on Debian 11 we will get an error:
livekit-ingress[68889]: open /sys/fs/cgroup/cpu.max: no such file or directory

this merge request fix it

https://github.com/livekit/ingress/issues/61

### DIFF
--- a/utils/cpu_linux.go
+++ b/utils/cpu_linux.go
@@ -187,6 +187,10 @@ func (cg *cpuInfoGetterV2) getTotalCPUTime() (int64, error) {
 func (cg *cpuInfoGetterV2) numCPU() (int, error) {
 	b, err := os.ReadFile(numCPUPathV2)
 	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			//File may not exist in case of no quota
+			return runtime.NumCPU(), nil
+		}
 		return 0, err
 	}
 


utils/cpu_linux.go
```diff
@@ -187,6 +187,10 @@ func (cg *cpuInfoGetterV2) getTotalCPUTime() (int64, error) {
 func (cg *cpuInfoGetterV2) numCPU() (int, error) {
 	b, err := os.ReadFile(numCPUPathV2)
 	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
```
biglittlebigben, line 190: The utils package already has fallback code in `newPlatformCPUMonitor` that calls lstat on "/sys/fs/cgroup/cpu.max", and fallback to `runtime.NumCPU()` if this fails. Do you mean that in you case, lstat succeeds on "/sys/fs/cgroup/cpu.max", but open later fails? If so, the correct fix would be to modify the `fileExists` function to call open rather than lstat. Is this happening a Debian 11 with default configuration?

m4ex: No, newPlatformCPUMonitor check file /sys/fs/cgroup/cpu.stat. And this file exists. But there is no file /sys/fs/cgroup/cpu.max in my case. It checks one file then try to read another, without fallback, that's strange. May be can I propose another change:
`cpuStatsPathV2 = "/sys/fs/cgroup/cpu.stat"` -> `cpuStatsPathV2 = "/sys/fs/cgroup/cpu.max"`. But I am not sure about that. I do not know cgroups very well. 

This is happening a Debian 11 with a default configuration (as least this configuration is default in our company). A will ask my colleagues is it default cgroups configuration for Debian 11 or they customized it.

m4ex:
```
root@hostname:/# lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye

root@hostname:/# hostnamectl
...
Operating System: Debian GNU/Linux 11 (bullseye)
Kernel: Linux 5.10.0-23-amd64
Architecture: x86-64

root@hostname:/# ls /sys/fs/cgroup
cgroup.controllers	cgroup.procs		cgroup.threads	       cpuset.mems.effective  dev-mqueue.mount	io.cost.qos  memory.numa_stat  -.mount			      sys-kernel-debug.mount	user.slice
cgroup.max.depth	cgroup.stat		cpu.pressure	       cpu.stat		      init.scope	io.pressure  memory.pressure   sys-fs-fuse-connections.mount  sys-kernel-tracing.mount
cgroup.max.descendants	cgroup.subtree_control	cpuset.cpus.effective  dev-hugepages.mount    io.cost.model	io.stat      memory.stat       sys-kernel-config.mount	      system.slice

root@ hostname:/# cat /sys/fs/cgroup/cpu.stat
usage_usec 17846400000
user_usec 13688612000
system_usec 4157788000
nr_periods 0
nr_throttled 0
throttled_usec 0
```
As I understand system wide CPU implementation use some cgo module for getting current cpu usage. But in this case we can use cpu.stat like in cpuInfoGetterV2. When we don't have cpu.max file, we should handle it just like file with string max
```
if m[0] == "max" {
		// No quota
		return runtime.NumCPU(), nil
```

biglittlebigben: According to https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html, `cpu.max` doesn't exist on root cgroup, so your patch is indeed the proper fix.

biglittlebigben (approved):

utils/cpu_linux.go
biglittlebigben: Thanks for the patch!

PR #404, opened 2023-06-17
add room log sampler

paulwe:
around 5.7% of logs from cloud don't have a `room` field so for now i think we're safe to write all of these. if it turns out log sampling was culling anything especially noisy we can deal with it later but this seems low risk. 

`projectID` is present in around 86.8% of logs so we could consider applying project-specific room sampling rates... the logs where `room` is defined but `projectID` is not seem to mostly be from the parts of signaling shared with oss.


boks1971:
lgtm @paulwe ! But, this log sampling thing and the interfaces/option hurt my head. So, am going to ask some dumb questions :-) 

1. The room logger is using `ItemSampler` [here](https://github.com/livekit/livekit/blob/40f5902d36fbfe5cad52057635dd13be393a7bbd/pkg/rtc/utils.go#L145). Participant/Track loggers are derived from room logger. So, they would get the sampling too I believe (we ad hoc create logger with room/participant fields in a lot of places, they won't have sampling, with this they can get some form of sampling too). How would things look with this logger? I guess if this logger is enabled, it would override other log sampling?
2. Logger config specifically refers to sampled logs for `participant/track` (not related to this PR), but it is just item sampling right. Would be good to change the language there and mention it is a generic item sampling and participant/track can be quoted as examples of where it can be used.
3. What are you planning to use for room sampling rate with this logger.
4. The code looks like it is a generic field sampled logger. Would be good to call it `fieldSampler` (or something more generic in the logging library here which is more generic) and pass in a key. There can be a wrapper (or not) `roomSampler` in where it is used which uses the `fieldSampler` and instantiates it with `RoomKey`.
5. When doing the above, the config can be changed to be more generic (actually with the above change, config may not belong here, it could be moved to application config) to have fields like `UseFieldSampledLogging` and `FieldSampleLoggingKey`. Is that possible? I guess the main thing I am trying to suggest here is to avoid the concept of a specific thing like `roomSampler`/`RoomKey` in the logger which is currently very generic and used by a lot of repos.

boks1971:
Thinking more, I guess the generic field sampled logging config can be here (and should be here if it applies globally). It can be done using
1. `UseFieldSampledLogging`
2. `FieldSampledLoggingKey`
3. `FieldSampleLoggingRate`

paulwe:
1. it seemed like `ItemSampler` was used to create a logger with its own counters. i was going to delete these but since this would break backward compatibility with the old sampler i'll need to do something with it. good callout
2. 👍 
3. 50%? it would be nice to base this on data but i don't think dd queries are powerful enough to give us the aggregate we need to make a less arbitrary decision.
4. 👍 
5. 👍 

i'm not sure how generic this needs to be. i agree that it can accept the sample driving field but i can't really see operators getting value out of being able to configure it. it only makes sense to sample using room which was why i didn't bother with this in the first place. also, with any luck we can throw this all out in a few weeks so it may never even be part of an oss release.

paulwe:
i thought about this and ended up going a slightly different direction. we can leave oss untouched since this is sort of a stopgap until we can migrate away from dd.

the config remains unchanged so we don't have to worry about how it's presented as much - when we construct the logger in cloud we can call `WithFieldSampler` to add the room sampling behavior. the threshold is defined with an atomic so we can wire this into config reloading and make adjustments based on observed log rate. as long as legacy sampling is disabled `WithItemSampler` will just work. we don't have to change anything here.

logger/fieldsampler.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"math"
+	"reflect"
+	"sync/atomic"
+
+	"github.com/zeebo/xxh3"
+	"go.uber.org/zap/zapcore"
+)
+
+const RoomKey = "room"
```
boks1971, line 13: Guess this can be removed.

boks1971 (approved):

PR #410, opened 2023-06-30
Use floats for cpu

frostbyte73:


davidzhao (approved):
mind flipping the minor protocol version? as a reminder of the incompatible API

PR #411, opened 2023-06-30
Add stream to track request/info

cnderrauber:
Use stream to group camera/microphone, screenshare/audio for a/v sync


boks1971 (approved):

livekit_models.proto
```diff
@@ -142,6 +142,7 @@ message TrackInfo {
   // true if RED (Redundant Encoding) is disabled for audio
   bool disable_red = 15;
   Encryption.Type encryption = 16;
+  string stream = 17;
```
davidzhao, line 145: would this be randomly generated? if so should it be called stream_id ?

davidzhao (approved):

livekit_models.proto
cnderrauber: Currently, it is more like a stream name that generated by sfu automatically from the source:

```
func StreamFromTrackSource(source livekit.TrackSource) string {
	// group camera/mic, screenshare/audio together
	switch source {
	case livekit.TrackSource_SCREEN_SHARE:
		return "screen"
	case livekit.TrackSource_SCREEN_SHARE_AUDIO:
		return "screen"
	case livekit.TrackSource_CAMERA:
		return "camera"
	case livekit.TrackSource_MICROPHONE:
		return "camera"
	}
	return "unknown"
}
```

PR #412, opened 2023-06-30
Timed aggregator.

boks1971:
Online aggretor over time.

Samples added aggregate (scaled by amount of time a sample has lived for) at the time of adding sample.

Can get aggregate, aggregate duration and average value later.

First use: in scorer to keep track of aggregate bandwidth, average distance from desired layer

### DIFF
--- a/utils/timed_aggregator.go
+++ b/utils/timed_aggregator.go
@@ -9,7 +9,7 @@ import (
 // ------------------------------------------------
 
 var (
-	errAnachronousSample = errors.New("anachronous sample")
+	ErrAnachronousSample = errors.New("anachronous sample")
 )
 
 // ------------------------------------------------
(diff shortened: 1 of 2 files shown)


paulwe (approved):

utils/timed_aggregator.go
```diff
@@ -0,0 +1,173 @@
...
+import (
+	"errors"
+	"sync"
+	"time"
+)
+
+// ------------------------------------------------
+
+var (
+	errAnachronousSample = errors.New("anachronous sample")
```
paulwe, line 12: maybe export this since it's part of the interface?

paulwe:
looks good 👍

utils/timed_aggregator.go
boks1971: Will do.

I also think that I should do this error return on other functions like `GetAggregateAt` etc. as this cannot do negative time. Will make those changes also.

PR #414, opened 2023-07-05
Retry on egress timeout or resource exhausted

frostbyte73:


davidzhao (approved):
awesome to know psrpc has this thing built-in

PR #420, opened 2023-07-12
Add location to startEgressRequest

frostbyte73:


davidzhao:
I don't see location defined anywhere?

frostbyte73:
> I don't see location defined anywhere?

https://github.com/livekit/protocol/blob/9d92e351b8e1ae47eafbe6fb469ad84645c92726/rpc/egress.proto#L47

davidzhao (approved):
oops, my bad.. was looking at a single commit. instead of the whole PR.

PR #423, opened 2023-07-15
Do not update if the model does not change.

boks1971:
Trying to avoid unnecessary updates.


utils/timeseries/timeseries.go
```diff
@@ -335,7 +335,7 @@ func (t *TimeSeries[T]) ZScore(val T) float64 {
 
 	onlineStdDev := t.onlineStdDevLocked()
 	if onlineStdDev != 0.0 {
-		return (float64(val) - t.welfordM) / t.onlineStdDevLocked()
+		return (float64(val) - t.welfordM) / onlineStdDev
```
boks1971, line 338: random drive-by.

utils/protoproxy.go
```diff
@@ -43,10 +44,11 @@ type ProtoProxy[T proto.Message] struct {
 // this should be used for updates that should be sent periodically, but does not have the urgency of immediate delivery
 // updateFn should provide computations required to generate the protobuf
 // if refreshInterval is 0, then proxy will only update on MarkDirty(true)
-func NewProtoProxy[T proto.Message](refreshInterval time.Duration, updateFn func() T) *ProtoProxy[T] {
+func NewProtoProxy[T proto.Message](refreshInterval time.Duration, updateFn func() T, equalityFn func(T, T) bool) *ProtoProxy[T] {
 	p := &ProtoProxy[T]{
 		updateChan:      make(chan struct{}, 1),
 		updateFn:        updateFn,
+		equalityFn:      equalityFn,
```
davidzhao, line 51: is it meaningfully different between types? or is using [proto.Equal](https://pkg.go.dev/google.golang.org/protobuf/proto#Equal) be sufficient?

davidzhao (approved):
great fix!

utils/protoproxy.go
boks1971: I know we are calling this `protoProxy`, but as it is using a generic type, I just gave the flexibility to the instantiating module as it will know about the type. If you think, `proto.Equal` is fine, I can remove the function.

boks1971: oh `T` is just `proto.Message`. Sorry, ok, will remove it.

PR #426, opened 2023-07-25
Remove deprecated egress rpc

frostbyte73:


livekit_egress.proto
```diff
@@ -35,11 +35,6 @@ message RoomCompositeEgressRequest {
   bool audio_only = 3;          // (default false)
   bool video_only = 4;          // (default false)
   string custom_base_url = 5;   // template base url (default https://recorder.livekit.io)
-  oneof output {                // deprecated (use _output fields)
```
davidzhao, line 38: ok, we cannot deprecate this yet. this is a major change.

especially for Cloud customers, we need to ensure everyone's migrated off before pulling the trigger.

livekit_egress.proto
```diff
@@ -267,21 +252,11 @@ message EgressInfo {
     TrackEgressRequest track = 6;
     WebEgressRequest web = 14;
   }
-  oneof result { // deprecated (use _result fields)
```
davidzhao, line 270: same for all the result fields

livekit_egress.proto
frostbyte73: I can leave the fields in for now and just remove the rpc system

davidzhao (approved):
lgtm!

PR #436, opened 2023-08-02
Component support for logger

davidzhao:
With the ability to dynamically update log levels per component

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -15,6 +15,7 @@
 package logger
 
 import (
+	"strings"
 	"sync"
 	"time"
 
@@ -90,14 +91,16 @@ type Logger interface {
 
 type sharedConfig struct {
 	level           zap.AtomicLevel
-	componentLevels sync.Map // map[string]zapcore.Level
+	mu              sync.Mutex
+	componentLevels map[string]zap.AtomicLevel
 	config          *Config
 }
 
 func newSharedConfig(conf *Config) *sharedConfig {
 	sc := &sharedConfig{
-		level:  zap.NewAtomicLevelAt(ParseZapLevel(conf.Level)),
-		config: conf,
+		level:           zap.NewAtomicLevelAt(ParseZapLevel(conf.Level)),
+		config:          conf,
+		componentLevels: make(map[string]zap.AtomicLevel),
 	}
 	conf.AddUpdateObserver(sc.onConfigUpdate)
 	_ = sc.onConfigUpdate(conf)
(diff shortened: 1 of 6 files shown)


logger/logger.go
```diff
@@ -239,6 +309,10 @@ func (l *ZapLogger) WithoutSampler() Logger {
 	return &dup
 }
 
+func (l *ZapLogger) isEnabled(level zapcore.Level) bool {
+	return l.sharedConfig.isEnabled(l.component, level)
```
boks1971, line 313: I am possibly not understanding this correctly. Is it possible to update loggers with component level if the logger has a component. This one is doing a sync map load on every log line. Feels like that could be expensive. Even if not expensive, feels like this logger struct can store the level based on component and just do a simple level check?

boks1971 (commented):
Wanted to bring up the issue of hierarchical loggers. Is it possible to support that? For example, `rtc` is one level down, but there could be `rtc.room`, `rtc.participant` etc. So, we can just change log level of any depth in the hierarchy. If `rtc` is set, applies to all loggers in that namespace.

Guess that makes this more complex, but would be nice to have that granularity. If it is not too much effort, probably worth doing it now. If not, maybe v2 then?

logger/logger.go
paulwe: we could probably load the log levels into something like `map[component]zap.AtomicLogLevel` then load the `zap.AtomicLogLevel` once when the logger is created and use `(zap.AtomicLogLevel).SetLevel` when the config updates. then we wouldn't need the config callback thing or custom level filtering. 

any log levels from the map unspecified by the updated config would revert to the default level. eliminating the callback list also removes the risk of a memory leak when creating these loggers outside of component init.

paulwe (approved):

paulwe:
log level cascading for hierarchical loggers could be applied once when the config is translated to the map of atomic levels or during `WithComponent` for unspecified subcomponents... this could be implemented without adding a meaningful amount of complexity.

boks1971:
> log level cascading for hierarchical loggers could be applied once when the config is translated to the map of atomic levels or during `WithComponent` for unspecified subcomponents... this could be implemented without adding a meaningful amount of complexity.

Yeah agreed, that makes sense. if we parse config and apply to loggers on config change, hierarchical  should be straightforward.

davidzhao:
Thanks for the feedback guys! made this hierarchical and using AtomicLevel as @paulwe suggested.

paulwe:
sweet, lgtm! 🚀

PR #439, opened 2023-08-03
Add playoutdelay to CreateRoomRequest

cnderrauber:


boks1971 (approved):

livekit_room.proto
```diff
@@ -74,6 +74,8 @@ message CreateRoomRequest {
   string metadata = 5;
   // egress
   RoomEgress egress = 6;
+  // playout delay
+  PlayoutDelay playout_delay = 7;
```
davidzhao, line 78: does it still make sense to send this structure back? with `min`, `max` ?

it would be simpler from an API standpoint to specify a `min_playout_delay`. I don't know that we should expose max here.

cnderrauber: Do we have a case to limit the max playout delay? Can remove it if we don't have want that case now

davidzhao: I don't believe we do.. and I think this would depend on the jitter conditions right? (Chrome will not honor max if network conditions are poor)

cnderrauber: I'll remove it now if we don't have case require it.

I have not test the max value but from the source code, the VCM timing will try to make sure the delay stays in the range [min,max].

davidzhao: yes let's remove the max.. it's simpler to use when creating a room to say:

```
CreateRoom({
  minPlayoutDelay: 100
})
```

instead of

```
CreateRoom({
  playoutDelay: {
    enabled: true,
    min: 100,
    max: 1000,
  }
})
```

davidzhao: mainly because users would have to think about if they have to set max.. and what value they should be set to.

cnderrauber: Sounds good, I'll remove it from the config structure/file as well

davidzhao (approved):

PR #443, opened 2023-08-09
Participant composite

frostbyte73:


livekit_egress.proto
```diff
@@ -83,6 +84,19 @@ message WebEgressRequest {
   repeated SegmentedFileOutput segment_outputs = 11;
 }
 
+// record audio and video from a single participant
+message ParticipantEgressRequest {
+  string room_name = 1; // required
+  string identity = 2;  // required
```
davidzhao, line 90: would be great to have `TrackSource source`, defaulting to Camera, but can be ScreenShare

livekit_egress.proto
```diff
@@ -277,9 +291,10 @@ message EgressInfo {
   string error = 9;
   oneof request {
     RoomCompositeEgressRequest room_composite = 4;
+    WebEgressRequest web = 14;
+    ParticipantEgressRequest participant = 19;
```
davidzhao, line 295: nit:

would be useful to keep a // NEXT_ID comment.

livekit_egress.proto
```diff
@@ -330,6 +345,15 @@ message SegmentsInfo {
   int64 ended_at = 7;
 }
 
+message AutoParticipantEgress {
+  oneof options {
+    EncodingOptionsPreset preset = 1; // (default H264_720P_30)
+    EncodingOptions advanced = 2;     // (optional)
+  }
+  repeated EncodedFileOutput file_outputs = 3;
+  repeated SegmentedFileOutput segment_outputs = 4;
```
davidzhao, line 354: would be good to have a `repeated TrackSource sources` here to filter the sources they want.. by default it's all.

davidzhao (commented):
lg! just a comment re ability to capture screenshare tracks.

livekit_egress.proto
biglittlebigben: I agree it's a good feature to have, but any reason not to add it shortly after the 1st release to avoid creeping the initial scope?

davidzhao: yeah that'd be fine too. though if it's one of those things that's simple to design in, then that'd be preferred. I think its utility would be severely limited if it could only handle a participant's camera tracks

biglittlebigben (approved):

davidzhao (approved):

PR #447, opened 2023-08-11
Update ingress layer validation logic to make sure both dimensions in crease for each layer

biglittlebigben:
This is a preamble to changing the ingress logic to boxing the input video into the layer dimension, instead of resizing to the layer dimensions regardless of aspect ratio. 

This may cause some existing ingress to stop working.

### DIFF
--- a/ingress/validation.go
+++ b/ingress/validation.go
@@ -175,18 +175,21 @@ func ValidateVideoEncodingOptionsConsistency(options *livekit.IngressVideoEncodi
 		layersByQuality[layer.Quality] = layer
 	}
 
-	var oldLayerArea uint32
+	var oldW, oldH uint32
 	for q := livekit.VideoQuality_LOW; q <= livekit.VideoQuality_HIGH; q++ {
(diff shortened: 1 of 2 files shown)


davidzhao:
> This may cause some existing ingress to stop working

Can you say more about this? What would stop working?

biglittlebigben:
> > This may cause some existing ingress to stop working
> 
> Can you say more about this? What would stop working?

currently, 

low -> 320x240
high -> 180x480 

would be a supported layer configuration since 180*480 > 320*240. However, with the code I'm adding to ingress to maintain the source aspect ratio, you'd end up (for a 4/3 ratio source) with 

low -> 320x240
high -> 180x136

which doesn't make sense.

As we validate layer settings before starting an ingress (in addition to creation), if somebody already created an ingress such as above, it'd fail when trying to start it after this change gets released.

davidzhao (approved):
that makes sense. I think it's fine to actually do the right thing here.

the only cases where it'll stop working is when users likely have made a mistake with dimensions.

PR #456, opened 2023-08-28
RTPDrift as a separate model.

boks1971:


livekit_models.proto
```diff
@@ -378,9 +391,12 @@ message RTPStats {
   uint32 layer_lock_plis = 35;
   google.protobuf.Timestamp last_layer_lock_pli = 36;
 
-  double sample_rate = 42;
-  double drift_ms = 43;
-  // NEXT_ID: 44
+  // DEPRECATED double sample_rate = 42;
```
davidzhao, line 394: we can delete this right?

davidzhao (approved):
lgtm

livekit_models.proto
boks1971: Yeah. Will do.

PR #468, opened 2023-09-11
Log Pion trace level logs at Debug level.

boks1971:
Trying to debug connectivity issues in a deploy.
Need Pion trace level logs to check ICE details (candidates, STUN binding request/responses, etc.). So, log pion trace at Debugw instead of ignoring it.

### DIFF
--- a/logger/pionlogger/logadapter.go
+++ b/logger/pionlogger/logadapter.go
@@ -28,11 +28,11 @@ type logAdapter struct {
 }
 
 func (l *logAdapter) Trace(msg string) {
-	// ignore trace
+	l.logger.Debugw(msg)
 }
 
 func (l *logAdapter) Tracef(format string, args ...interface{}) {
-	// ignore trace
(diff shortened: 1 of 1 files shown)


paulwe (approved):

logger/pionlogger/logadapter.go
```diff
@@ -28,11 +28,11 @@ type logAdapter struct {
 }
 
 func (l *logAdapter) Trace(msg string) {
-	// ignore trace
+	l.logger.Debugw(msg)
 }
 
 func (l *logAdapter) Tracef(format string, args ...interface{}) {
-	// ignore trace
```
davidzhao, line 35: Should we do the l.shouldIgnore check with these?

davidzhao (approved):

logger/pionlogger/logadapter.go
boks1971: Intentional skip as we want verbose details.

PR #475, opened 2023-09-18
Use a different object to hold lock.

boks1971:
There were places were the default config object is copied using a = b. Those places are warning about copying locks.

Using a wrapper struct to put locks there.
Had to add a global though. Not great. Open to suggestions to not use a global.


logger/config.go
```diff
@@ -38,37 +38,45 @@ type Config struct {
...
 type ConfigObserver func(*Config) error
 
+type updatableConfig struct {
+	*Config
+
+	lock               sync.Mutex
+	onUpdatedCallbacks []ConfigObserver
+}
+
+var uConfig updatableConfig
```
boks1971, line 52: Don't like the global and having to set the `Config` field in logger.go. Could not think of a better way. Open to suggestions.

paulwe (approved):
seems fine unless being a global breaks something

logger/config.go
davidzhao: would prefer to not have the global reference. it can introduce unexpected side-effects for other Config users.

what about changing the Logging field to a pointer type?

boks1971:
Closing this one in favour of copying config via yaml marshal/unmarshal to avoid lock copy warnings.

PR #483, opened 2023-09-28
Add SyncStreams flag to room

cnderrauber:


livekit_models.proto
```diff
@@ -34,8 +34,9 @@ message Room {
   uint32 num_publishers = 11;
   bool active_recording = 10;
   PlayoutDelay playout_delay = 12;
+  bool sync_streams = 13;
```
boks1971, line 37: Did not ask this the last time, but wondering if this and `playout_delay` should be part of the room model. WDYT?

boks1971 (approved):
One question @cnderrauber . Do you think we should have `sync_streams` and `playout_delay` in room model?

livekit_models.proto
cnderrauber: It is room model here, do you mean there is another room model or move them out?

boks1971: Oh sorry :-) I was asking if this should even be here. This is mostly used to send `RoomInfo` to participants. So, is there use to having these in the model? There are two other configurations there (empty_timeout and max_participants). So, maybe it is good to have these in there too. My general thought was the usefulness of having config kind of things in the room model which is mainly used to convey room info to participants. 

Oh, maybe these are required in relay case when remote sets up a room?

cnderrauber: We are using the room model for both config and `RoomInfo`, so these two fields (and other non-info) are added to it for remote nodes to set up the room because we don't have a dedicated `RoomInfo` message to notify to participants only. It makes sense to separate them as the participants don't care about the config fields, we can add the Info message in future.

boks1971: Sounds good. Thank you @cnderrauber !!

davidzhao: maybe we should add this to the internal part of Room model? that is persisted, but not sent back to users

davidzhao: It's [here](https://github.com/livekit/protocol/blob/main/livekit_internal.proto#L182)

PR #484, opened 2023-09-28
Move playout_delay & sync_streams to internal model

cnderrauber:


livekit_internal.proto
```diff
@@ -182,6 +182,8 @@ message KeepAlive {
 message RoomInternal {
   AutoTrackEgress track_egress = 1;
   AutoParticipantEgress participant_egress = 2;
+  PlayoutDelay playout_delay = 3;
+  bool sync_streams = 4;
```
boks1971, line 186: might be too late now, can we also move `empty_timeout` and `max_participants` here and deprecate it in room model? Not sure if any client is using it now. Would be good to have a clean separation of internal.

boks1971 (approved):

livekit_internal.proto
cnderrauber: I believe they are not used too, but double confirm with clients would be better before removing them.

PR #489, opened 2023-10-03
check psrpc compiler version

paulwe:
provide more friendly feedback when running psrpc codegen with an outdated compiler


boks1971 (approved):
Thank you!!

PR #496, opened 2023-10-23
add room psrpc interface

paulwe:


rpc/typed_api.go
```diff
@@ -29,3 +34,31 @@ func NewTypedSignalClient(nodeID livekit.NodeID, bus psrpc.MessageBus, opts ...p
 func NewTypedSignalServer(nodeID livekit.NodeID, svc SignalServerImpl, bus psrpc.MessageBus, opts ...psrpc.ServerOption) (TypedSignalServer, error) {
 	return NewSignalServer[livekit.NodeID](string(nodeID), svc, bus, opts...)
 }
+
+type ParticipantTopic string
+type RoomTopic string
+
+func FormatParticipantTopic(roomName livekit.RoomName, identity livekit.ParticipantIdentity) ParticipantTopic {
+	return ParticipantTopic(fmt.Sprintf("%s_%s", roomName, identity))
```
boks1971, line 42: Is there any splitting happening based on de-limiter? Guessing not. Just want to make sure that we are using `base62` (or some other encoding) with user provided names/identities if we need to split later.

boks1971 (approved):

rpc/typed_api.go
paulwe: no, i try to avoid that sort of thing as much as possible...

PR #507, opened 2023-10-28
Add SIP Support

Sean-Der:


livekit_sip.proto
```diff
@@ -0,0 +1,163 @@
...
+// * `auto_bridge_from` is set and `pin` is set
+//   User is added to a room if value matches calling phone number and the
+//   correct pin is entered
+//
+message CreateSIPBridgeRequest {
+  // What Room is this SIP Traffic Bridged into
+  string room_name = 1;
+
+  // If call is received on specified Trunk bridge into the room
+  string auto_bridge_trunk = 2;
```
davidzhao, line 96: would be better if we always specify the trunks that this API applies to:

```suggestion
  repeated string trunk_ids = 2;
```

livekit_sip.proto
```diff
@@ -0,0 +1,163 @@
...
+
+  // If call is received from a specific number bridge into the room
+  string auto_bridge_from = 3;
+
+  // Pin required to use this bridge
+  string pin = 4;
+
+  // By default the From value (Phone number) is used as the participant identity
+  // If true a random value will be used instead
+  bool use_random_participant_identity = 5;
```
davidzhao, line 106: might be more direct? wdyt

```suggestion
  bool hide_phone_number = 5;
```

davidzhao (approved):
🚀 🚀 

can we put the dispatch docs right next to their oneof? i.e. `SIPDispatchRuleDirect dispatch_rule_direct = 1;`

it'll help codegen to attach these to all of our server API docs automatically

biglittlebigben (approved):

PR #509, opened 2023-11-01
split room service

paulwe:
split roomservice backend apis into room and participant services to simplify pubsub topic registration


rpc/typed_api.go
```diff
@@ -17,11 +17,66 @@ package rpc
...
+	Observer         middleware.MetricsObserver
+	VersionGenerator utils.TimedVersionGenerator
+}
+
+func NewClientParams(
+	config PSRPCConfig,
+	bus psrpc.MessageBus,
+	logger logger.Logger,
+	observer middleware.MetricsObserver,
+) ClientParams {
```
boks1971, line 58: No version generator? The structure above has that.

boks1971 (approved):

rpc/typed_api.go
paulwe: no, i missed this, thanks 👍

PR #511, opened 2023-11-01
Agents

frostbyte73:


livekit_agent.proto
```diff
@@ -0,0 +1,111 @@
...
+        // server confirms the registration, from this moment on, the worker is considered active
+        RegisterWorkerResponse register = 1;
+        // server asks worker to confirm availability for a job
+        AvailabilityRequest availability = 2;
+        JobAssignment assignment = 3;
+    }
+}
+
+enum JobType {
+    JT_ROOM = 0;
```
keepingitneil, line 62: Not sure how we like to do it, but I think best practice is to have the 0 value be JT_UNSPECIFIED

https://protobuf.dev/programming-guides/style/

keepingitneil (approved):

livekit_agent.proto
frostbyte73: that is how we do it with egress/ingress, but in this case a value will always be required

PR #514, opened 2023-11-08
Add SIP RPC

Sean-Der:


livekit_sip.proto
```diff
@@ -66,20 +66,31 @@ service SIP {
...
+  // Username and password used to authenticate SIP invites
+  // May be empty to have no Authentication
+  string username = 5;
+  string password = 6;
 }
 
 message SIPTrunkInfo {
   string sip_trunk_id = 1;
+
+  repeated string addresses = 2;
```
dennwc, line 91: Maybe name fields here the same as in `CreateSIPTrunkRequest`? It's missing one field as well.

livekit_sip.proto
```diff
@@ -66,20 +66,31 @@ service SIP {
...
+  string outbound_number = 3;
 
   // Accepted `To` values. This Trunk will only accept a call made to
   // these numbers. This allows you to have distinct Trunks for different phone
   // numbers at the same provider.
-  // An empty list means all dialed numbers are accepted.
-  repeated string allowed_destinations_regex = 3;
+  repeated string inbound_numbers_regex = 4;
+
+  // Username and password used to authenticate SIP invites
```
dennwc, line 82: Probably worth mentioning that these are inbound invites, right?

rpc/io.proto
```diff
@@ -55,3 +59,44 @@ message UpdateIngressStateRequest {
...
+
+  // What Number is calling
+  string from = 2;
+
+  // What Number was called
+  string to = 3;
+
+  // What is the IP address of the called number
+  string src_address = 4;
+}
```
dennwc, line 91: We need a PIN here as well.

rpc/io.proto
```diff
@@ -55,3 +59,44 @@ message UpdateIngressStateRequest {
   string ingress_id = 1;
   livekit.IngressState state = 2;
 }
+
+message GetSIPTrunkAuthenticationRequest {
+  // What Number is calling
+  string from = 2;
```
dennwc, line 65: Maybe use the same inbound/outbound naming here?

livekit_sip.proto
Sean-Der: Fixed! It now mentions they can be used for inbound and outbound

rpc/io.proto
Sean-Der: Done!

livekit_sip.proto
```diff
@@ -66,20 +66,47 @@ service SIP {
...
   // these numbers. This allows you to have distinct Trunks for different phone
   // numbers at the same provider.
-  // An empty list means all dialed numbers are accepted.
-  repeated string allowed_destinations_regex = 3;
+  repeated string inbound_numbers_regex = 4;
+
+  // Username and password used to authenticate inbound and outbound SIP invites
+  // May be empty to have no Authentication
+  string username = 5;
+  string password = 6;
```
biglittlebigben, line 85: Not exactly a protocol issue, but are there requirements about encrypting user credentials we need to store?

livekit_sip.proto
```diff
@@ -66,20 +66,47 @@ service SIP {
...
+
+  // Accepted `To` values. This Trunk will only accept a call made to
+  // these numbers. This allows you to have distinct Trunks for different phone
+  // numbers at the same provider.
+  repeated string inbound_numbers_regex = 5;
+
+  // Username and password used to authenticate inbound and outbound SIP invites
+  // May be empty to have no Authentication
+  string username = 6;
+  string password = 7;
```
davidzhao, line 109: we should mask or return blank username/password when listing/getting trunks.

rpc/sip.proto
```diff
@@ -0,0 +1,31 @@
...
+syntax = "proto3";
+
+package rpc;
+
+option go_package = "github.com/livekit/livekit/pkg/service/rpc";
+
+import "options.proto";
+
+service SIPInternal {
+  rpc UpdateSIPParticipant(InternalUpdateSIPParticipantRequest) returns (InternalUpdateSIPParticipantResponse) {
```
davidzhao, line 24: looks like this is just a placeholder? I think we'll want topic based routing to the node hosting the participant instead of affinity based selection

davidzhao (approved):
lg!

PR #520, opened 2023-11-15
Update SIP protocol

dennwc:
Two changes in this PR:
- Remove the `SIPDispatchRulePin`. We can just use `SIPDispatchRuleDirect` with a pin.
- Add NoPin flag to `EvaluateSIPDispatchRulesRequest`. This will allow the caller to cancel pin prompt and (possibly) join public rooms on the same trunk.


livekit_sip.proto
```diff
@@ -151,10 +143,6 @@ message SIPDispatchRule {
     // enter the room
     SIPDispatchRuleDirect dispatch_rule_direct = 1;
 
-    // SIPDispatchRulePin is a `SIP Dispatch Rule` that allows a user to choose between multiple rooms.
```
Sean-Der, line 154: @dennwc I think you need to update docs in a few other places.

Mind also updating the id so it is `1 -> 2` since we aren't public yet, should be ok?

dennwc: Done!

Sean-Der (approved):

PR #529, opened 2023-11-29
process monitoring

frostbyte73:


utils/cpu.go
```diff
@@ -92,9 +113,103 @@ func (c *CPUStats) monitorCPULoad() {
...
+			return
+		case <-ticker.C:
+			nextStats := make(map[int]procfs.ProcStat)
+			procs, err := procfs.AllProcs()
+			if err != nil {
+				logger.Errorw("failed to read processes", err)
+				continue
+			}
+
+			total, err := fs.Stat()
```
biglittlebigben, line 154: This seems to get data from /proc/stat. Do you know if that returns data for the pod only or for the whole system? I believe it does matter here as we are normalizing the result to `platform.numCPU()` which will return the amount of cores associated to the pod.

frostbyte73: Ah, you're right. Looks like this will cause issues
https://academic.oup.com/gigascience/article/doi/10.1093/gigascience/giad069/7250999

utils/cpu_linux.go
```diff
@@ -260,3 +262,11 @@ func readIntFromFile(filename string) (int, error) {
 
 	return strconv.Atoi(s)
 }
+
+func getPodCPUCount(fs procfs.FS) (int, error) {
```
biglittlebigben, line 266: Should this be `getHostCPUCount` as it seems to be getting the total core count for the host?

utils/cpu.go
```diff
@@ -92,9 +113,109 @@ func (c *CPUStats) monitorCPULoad() {
...
+				if proc.PID != self.PID {
+					ppids[proc.PID], err = getPPID(proc.PID)
+					if err != nil {
+						logger.Errorw("failed to get PPID", err)
+						continue
+					}
+				}
+			}
+
+			nextTotalTime := total.CPUTotal.User + total.CPUTotal.Nice + total.CPUTotal.System + total.CPUTotal.Idle
```
biglittlebigben, line 182: I can't really judge whether or not that is the proper measure of total time. Could you document/link to the logic behind it?

frostbyte73: Might be the reason for the discrepancy - according to https://stackoverflow.com/questions/23367857/accurate-calculation-of-cpu-usage-given-in-percentage-in-linux, it's common to only use those 4, but there are more that should go into it

biglittlebigben (approved):

PR #534, opened 2023-11-30
Move SIP rules evaluation from LiveKit server

dennwc:


dennwc:
Wait, actually I need one more function here...

dennwc:
Should be good now.

Sean-Der (approved):

PR #542, opened 2023-12-11
Add Type for a Participant

dennwc:
The main idea is to have a few pre-defined types, but also allow specifying "type name", which can be used by app developers to make their own types. The same mechanism can be used for Agent types and SIP inbound/outbound.


livekit_models.proto
```diff
@@ -99,6 +99,16 @@ message ParticipantInfo {
...
   }
+  enum Type {
+    // normal participants, e.g. web clients
+    NORMAL = 0;
+    // SIP participants
+    SIP = 1;
+    // LiveKit agents
+    AGENT = 2;
+    // app-specific participants
+    CUSTOM = 3;
```
lukasIO, line 110: would we want to add a type for `egress`, too, or is that considered `agent` now as well?

dennwc: Do we currently list egress as a participant? What about ingress? I'm not yet familiar with that part, unfortunately.

lukasIO: egress is not really treated as regular participants, but rather as `hidden` one's, so they generally don't show up on client side libs. also unsure about ingress

livekit_models.proto
```diff
@@ -113,6 +123,9 @@ message ParticipantInfo {
   // indicates the participant has an active publisher connection
   // and can publish to the server
   bool is_publisher = 13;
+  Type type = 14;
```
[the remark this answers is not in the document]
lukasIO: oh, that's a good point. we have `kind` on the track level already, maybe we can re-use that here?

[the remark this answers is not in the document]
dennwc:
```protobuf
message TrackInfo {
  string sid = 1;
  TrackType type = 2;
```

Looks like it's actually "type" as well on a Track.

Having said that, switching to "kind" for Participant makes sense.

livekit_models.proto
dennwc: > I would suggest dropping the `CUSTOM` type. We can add it later if/when needed

Agree, don't have specific semantic for it in mind yet.

> the same is true for `type_name`

I'd still want to keep it. I think it will be useful to have inbound/outbound types for SIP, for example. Similarly, it would be useful for Agents.

The name for this field is bad, I must admit. Maybe call it `user_agent`?

livekit_models.proto
```diff
@@ -99,6 +99,16 @@ message ParticipantInfo {
     // WS disconnected
     DISCONNECTED = 3;
   }
+  enum Type {
+    // normal participants, e.g. web clients
+    NORMAL = 0;
```
biglittlebigben, line 104: "Normal" here may communicate that the other kinds are abnormal somehow. Would "standard" or "default" work better here?

dennwc: Agree, good idea!

livekit_models.proto
davidzhao: how would that be set? this permission bit is simply reflecting the state to others.

we already have ClientInfo for analytics purposes. I think this field is more for the developer's application to be able to differentiate between diff kinds of participants

dennwc: > I think this field is more for the developer's application to be able to differentiate between diff kinds of participants

This is exactly why I propose to keep it. Type/kind works for filtering large categories, but not for a particular Agent, for example.

Lets say we want to show some icon/image for a well-known audio-only Agent in Meet (KITT?). Currently it's done by emitting a placeholder video steam, I believe. With this new field, we can instead check `if kind==AGENT && type_name=="io.livekit.kitt"` in Meet and show a specific image for it. Maybe @lukasIO can comment more on usefulness of this approach (or lack of it).

If we consider the upcoming V2 CLI, this would also make sense too. We may decide to always expose all participant kind including ingress/egress/etc. In that case, clients may need additional info to decide which of them should be displayed in the room. Coarse grained filtering would be done by `kind` and fine-grained by `type_name`.

lukasIO: If I understand correctly, what you propose with `type_name` is maybe more application layer logic. I think the use cases you have in mind could be achieved e.g. with participant metadata updates?

dennwc: Oh, right, we still have metadata. That would work. Is there a specific format for it? Is it a string or JSON? I'll add a comment for it in protobuf.

lukasIO: no specific format, completely up to the user how to manage it. Generally metadata is a string!

dennwc:
Updated, should be good to go.

lukasIO (approved):
lgtm!

davidzhao (approved):

PR #545, opened 2023-12-13
Implement exponential backoff in egress RPC client retry logic

biglittlebigben:
This uses exponential increases for both the timeout and backoff parameters. The formula is for the increase is:

`base * 2 ^ (attempt - 1) * rand[1,2)`

This means that for the different attempts, we get:

```
1: base * [1-2)
2: base * [2-4)
3: base * [4-8)
4: base * [8-16)
...
```


biglittlebigben:
Thinking about this again, there is no reason to have a random component to the timeout parameter. Will change this.

frostbyte73 (approved):

PR #547, opened 2023-12-16
Adds FilteredNotifier

itzloop:
I have changed `QueuedNotifier`, adding `Stop(bool)`. This way we could use the same `DefaultNotifier` and then pass different implementations of `QueuedNotifier`. Now by calling `NewDefaultNotifierWithFilter`, one can only specify desired events and only receive those events and nothing more. The events are shared for all URLs but it's effortless to implement per URL event filter.


itzloop:
@davidzhao can you please review this pr?

itzloop:
Any updates on this?

itzloop:
@davidzhao Can you review this please.

PR #551, opened 2023-12-21
add deferred logger

paulwe:
allows creating a deferred zap logger that buffers log lines until the resolve function is called. when the logger is resolved the buffered logs are flushed with their original timestamps and the resolved values added as if they had been included using `WithValues`. resolved loggers function the same as loggers created using `WithValues`.

### DIFF
--- a/logger/deferred.go
+++ b/logger/deferred.go
@@ -1,0 +1,106 @@
+package logger
+
+import (
+	"sync"
+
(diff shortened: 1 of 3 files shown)


davidzhao (approved):
awesome!

boks1971 (approved):
Nice!!!

PR #554, opened 2023-12-21
analytics: send local room state

shishirng:


livekit_analytics.proto
```diff
@@ -152,3 +153,27 @@ message AnalyticsEvent {
...
+  string id = 1;
+  string name = 2;
+  google.protobuf.Timestamp created_at = 3;
+  uint32 num_participants = 4;
+  repeated AnalyticsRoomParticipant participants = 5;
+}
+
+message AnalyticsNodeRooms {
+  string node_id = 1;
+  uint64 sequence_number = 2;
```
davidzhao, line 176: what is this for?

livekit_analytics.proto
```diff
@@ -152,3 +153,27 @@ message AnalyticsEvent {
...
+  ParticipantInfo.State state = 4;
+  ParticipantInfo.Kind kind = 5;
+  google.protobuf.Timestamp joined_at = 6;
+}
+
+message AnalyticsRoom {
+  string id = 1;
+  string name = 2;
+  google.protobuf.Timestamp created_at = 3;
+  uint32 num_participants = 4;
```
davidzhao, line 170: what do we use these for?

livekit_analytics.proto
```diff
@@ -152,3 +153,27 @@ message AnalyticsEvent {
 message AnalyticsEvents {
   repeated AnalyticsEvent events = 1;
 }
+
+message AnalyticsRoomParticipant {
+  string id = 1;
+  string identity = 2;
+  string name = 3;
+  ParticipantInfo.State state = 4;
+  ParticipantInfo.Kind kind = 5;
```
davidzhao, line 162: nbd, but do we need name and kind? is the idea to reconstruct things from here?

livekit_analytics.proto
shishirng: auto incremented counter to stamp messages. Will be useful in debugging/tracking missed messages

livekit_analytics.proto
shishirng: will remove, its redundant, we can infer from message

livekit_analytics.proto
shishirng: thats right, if there are missed join/active event and participant was still active, we could reconstruct participant with the data

davidzhao (approved):
lgtm!

PR #569, opened 2024-01-11
tls config extension for redis

wolf510:
Supported:
  - client certificates,
  - serverName
  - insecure

### DIFF
--- a/redis/redis.go
+++ b/redis/redis.go
@@ -29,19 +29,21 @@ import (
 var ErrNotConfigured = errors.New("Redis is not configured")
 
 type RedisConfig struct {
-	Address           string      `yaml:"address,omitempty"`
-	Username          string      `yaml:"username,omitempty"`
-	Password          string      `yaml:"password,omitempty"`
-	DB                int         `yaml:"db,omitempty"`
-	TLS               xtls.Config `yaml:"tls,omitempty"`
-	MasterName        string      `yaml:"sentinel_master_name,omitempty"`
-	SentinelUsername  string      `yaml:"sentinel_username,omitempty"`
-	SentinelPassword  string      `yaml:"sentinel_password,omitempty"`
-	SentinelAddresses []string    `yaml:"sentinel_addresses,omitempty"`
-	ClusterAddresses  []string    `yaml:"cluster_addresses,omitempty"`
-	DialTimeout       int         `yaml:"dial_timeout,omitempty"`
-	ReadTimeout       int         `yaml:"read_timeout,omitempty"`
-	WriteTimeout      int         `yaml:"write_timeout,omitempty"`
+	Address  string `yaml:"address,omitempty"`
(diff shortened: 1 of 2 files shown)


wolf510:
@davidzhao can you please review this pr?

wolf510:
@paulwe can you review? please..

redis/redis.go
```diff
@@ -28,19 +29,19 @@ import (
...
-	SentinelAddresses []string `yaml:"sentinel_addresses,omitempty"`
-	ClusterAddresses  []string `yaml:"cluster_addresses,omitempty"`
-	DialTimeout       int      `yaml:"dial_timeout,omitempty"`
-	ReadTimeout       int      `yaml:"read_timeout,omitempty"`
-	WriteTimeout      int      `yaml:"write_timeout,omitempty"`
+	Address           string      `yaml:"address,omitempty"`
+	Username          string      `yaml:"username,omitempty"`
+	Password          string      `yaml:"password,omitempty"`
+	DB                int         `yaml:"db,omitempty"`
+	TLS               xtls.Config `yaml:"tls,omitempty"`
```
davidzhao, line 36: this will break existing users that do use TLS. I would recommend the following:

- keep `use_tls`, but mark it as deprecated
- use a pointer type `*xtls.Config` for `tls`
- when reading config, if `tls` is not set, then defer to `use_tls` value

then in the next major version, we could then remove it.

davidzhao (commented):
Thanks for the PR! apologies for the delays in the review.

I think as long as we can make this backwards compatible and have sane defaults, we'd be happy to merge.

redis/redis.go
wolf510: done

redis/redis.go
```diff
@@ -80,9 +83,12 @@ func GetRedisClient(conf *RedisConfig) (redis.UniversalClient, error) {
 	var rc redis.UniversalClient
 	var tlsConfig *tls.Config
 
-	if conf.UseTLS {
-		tlsConfig = &tls.Config{
-			MinVersion: tls.VersionTLS12,
+	if (conf.TLS != nil && conf.TLS.Enabled) || conf.UseTLS {
+		var err error
+		tlsConfig, err = conf.TLS.ClientTLSConfig()
```
davidzhao, line 88: conf.TLS could be nil here and would panic.

wolf510: https://go.dev/play/p/YhBSjgK_yDY
@davidzhao

davidzhao: your playground is demonstrating the check around if.. that's not where the panic happens. It takes place within `ClientTLSConfig` where you access member variables of a nil TL struct.

wolf510: @davidzhao Yes, you're right!
Fixed

davidzhao (approved):
lg

PR #571, opened 2024-01-11
Pass participant kind in the grant

dennwc:
Follow up for #542. Pass the participant kind in the auth grant.

Ingress/Egress will automatically start receiving new tokens, while SIP will need to be updated to use the new helper for building the token.


davidzhao (approved):
lgtm!

biglittlebigben (approved):

PR #574, opened 2024-01-17
Add topics to ListActiveEgress

frostbyte73:


davidzhao (commented):
how does this work in terms of backwards compat?

frostbyte73:
> how does this work in terms of backwards compat?

Using an empty topic is the same as using no topics

davidzhao (approved):

PR #585, opened 2024-01-25
Augment LeaveRequest with alternate regions.

boks1971:


livekit_models.proto
```diff
@@ -339,6 +339,8 @@ enum DisconnectReason {
   ROOM_DELETED = 5;
   STATE_MISMATCH = 6;
   JOIN_FAILURE = 7;
+  MIGRATION = 8;
+  SIGNAL_CLOSE = 9;
```
boks1971, line 343: Adding a couple of reasons to give more clarity. Previously, `MIGRATION` was getting labelled as `DUPLICATE_IDENTITY` which is a bit confusing.

lukasIO (approved):
lgtm!

PR #595, opened 2024-02-07
agents protocol proposal

theomonnom:
Multiple things here:
### Orchestration improvements
1. Ping requests: When the worker register itself to the controller, we setup a ping interval, if the ping is too late, the controller will invalidate this worker (See 5. for retry) 
2. The Python framework will automatically create an HTTP endpoint for health check (for k8).  This will easily be configurable by providing an async function to our users. (It will defaults to a basic OK response) 
     NOTE: When combining the health checks to the new pipeline API, we could automatically check if a specific pipeline is stuck. (This is important to detect issues with other providers like ElevenLabs/Deepgram/OAI...)
3. Better job dispatching:
   1. It's up to the users to know how/when scaling up the workers but Cloud needs to be smarter on which worker the job should be dispatched. (e.g We want to avoid situation where we have a lot of nodes that handles only one job.)
   2. Cloud needs to know of much jobs one Worker could theoretically handle
        1. Users must define a function that returns the current load of a worker, this value is sent on each PingRequest and when accepting a job (Load is [0;1])
             2. ~~Users must define the cost of a job, we don't know if a job will use CPU/GPU or bandwidth, so we need this "abstract" and configurable cost that determines how much it'll will increase the load. (Cost is [0;1])~~
        3. ~~Depending on the job, the cost may be different, so the cost must be reported when accepting a job~~
        4. Then Cloud will determine which worker is the best to assign a specific job:
                 1. Send a request to every worker,
                 2. Determine the best worker with the current load ~~and the returned cost~~
                 3. How much answers should we wait for tho? this needs to be as fast as possible

4. When a worker becomes overloaded, it can request to migrate a specific job with a MigrateRequest, Cloud will then try to reassign the job to another worker (This should happens in the worst scenario) 

5. Automatic retry of jobs. Cloud and OSS must automatically detect when a job is failing and send a AvailabilityRequest with `resuming=true` to another worker.
   This could happens when the agent participant disconnects without sending JobState.JS_SUCCESS, or when the ping interval is too late.
   Also when we invalidate a worker due to health checks failing, all jobs inside this worker must be resumed.
   To know if a specific job can be resumed, this is specified inside the AvailabilityResponse (should_retry field).
   This also applies when a media node crashes or becomes unavailable.

### Third party workers

>  Agent that can connect to any project: e.g DirectAI or our own transcription agent

1. After thinking, we don't need a general subdomain like agents.livekit.cloud. What we're doing today is OK.
2. The main difference is how the jobs are dispatched, I propose to introduce the concept of namespaces.
   It has the goal to resolve multiple issues (multiple jobs for one room & third party workers where a company can connect to rooms of other projects).The namespaces in Cloud are handled differently from OSS. A namespace is optional, when NONE, it is the same as if the namespace was tied to a single project (the worker only receive jobs from the project):

#### OSS: 
1. Namespaces are created/removed on the fly when a worker registers using a specific namespace
2. Or another option for OSS is to just configure namespaces inside the yaml configuration file.

#### Cloud:
1. A namespace must be unique across all our project, they must be created inside our Cloud dashboard (name may be generated?).
     1. namespace can have different properties:
            1. `cloud global` means that this namespace can be used by other projects. 
     2. On the Cloud dashboard, users can enable/disable a specific namespaces 
            1. Only the project who created a namespace can create a Worker listening to this specific namespace. 
            2. We don't show the list of existing namespaces, the users must put it inside an input box, and when trying to activate it we show the name of the project that created this namespace (like LiveKit or DirectAI)

### Miscellaneous improvements
1. Moving JobSimulation to the SFU (today I fake this inside the Python framework, but I want to reduce the code there)
2. Job can send metadata to Cloud so we can show some information on our dashboard.


livekit_agent.proto
```diff
@@ -74,25 +78,49 @@ enum JobStatus {
...
+    uint32 ping_interval = 5;
+    optional string namespace = 6;
 }
 
 message RegisterWorkerResponse {
     string worker_id = 1;
     string server_version = 2;
 }
 
+message MigrateJobRequest {
```
keepingitneil, line 106: Migration might be a premature optimization. Instead - when requesting a job we could have a field (id) that signals if the job is a new job or a restart of a failed job.

The agent would would be storing state during it's lifetime (i.e. store chatgpt messages in redis keyed off of the job id).

In the retry case, the agent would pull it's state from some external storage as a part of its startup sequence.

keepingitneil: Yup I agree, it's up to the agent developer to store state. Do they have enough information to identify the state for the resumed job?

theomonnom: The JOB_ID will be the same

livekit_agent.proto
```diff
@@ -44,6 +44,9 @@ message WorkerMessage {
         // worker can update its status to the server, including taking itself out of the pool
         UpdateWorkerStatus status = 3;
         JobStatusUpdate job_update = 4;
+	PingRequest ping = 5;
+	SimulateJobRequest simulate_job = 6;
+	MigrateJobRequest migrate_job = 7;
```
keepingitneil, line 49: Do we need a MigrateJobResponse? Wondering how to minimize the time the agent is not running

Proposal of migration flow:
1. Overloaded Worker Requests a Migration
2. SFU receives migration, creates migration job request (different than normal job request because it should have the same agent identity?)
3. New worker picks up resumed job and connects
4. Overloaded worker disconnects because of duplicate identity?

theomonnom: So when migrating, it comes from the current worker, so it'll just disconnect. If the worker is overloaded something is not working anyway

theomonnom: The goal here is to remove a job asap to make the other running jobs run smoothly

keepingitneil: Hrmm maybe the migration should be thought of as more eager (to prevent downtime) and the overloaded case is considered more of a failure case captured with health checks?

Migration might also update the job cost so the SFU can schedule with more up-to-date information. Job cost might be something we want to be updateable (i.e. the egress problem of not knowing how much decoding it will have to do upfront).

Wdyt?

theomonnom: I think every app should optimize it to not happen. Also the cost is defined by the local worker. So the cost is someway already changed

keepingitneil: I'm just trying to think of the common case for migration. Consider an agent that takes 30 second clips that composes audio/video tracks together. As new users join, the agent will use more and more load. It's not easy to predict how much load the agent will consume up-front. 

So:
- we may want to update the cost periodically so the SFU has an accurate view (we can provide a default cost updater based on CPU / #Cores)
- as a single agent takes up more and more resources, the worker may want to migrate smaller jobs off before the worker becomes overloaded

theomonnom: The cost is only an abstract value. The SFU already know the load with our ping requests 🤷

theomonnom: I think u can have jobs that take more resource over time, but you can plan it with the costs

livekit_agent.proto
```diff
@@ -74,25 +78,49 @@ enum JobStatus {
...
+}
+message PingResponse {}
+
 message RegisterWorkerRequest {
     JobType type = 1;
     string worker_id = 2;
     string version = 3;
     string name = 4;
+    uint32 ping_interval = 5;
+    optional string namespace = 6;
```
davidzhao, line 98: should the worker indicate if they support migrations or not? thinking that if a job cannot be migrated (because they are not ready to handle the state), then it would be unexpected to receive another request for it.

livekit_agent.proto
```diff
@@ -74,25 +78,49 @@ enum JobStatus {
     JS_FAILED = 2;
 }
 
+message SimulateJobRequest {
+    JobType type = 1;
+    string room_id = 2;
+    string participant_id = 3;
+}
+
+message PingRequest {
```
davidzhao, line 87: It's rather unexpected for Ping to signal more than liveliness. If we are updating the server about information.. I think we should rename it to something like `UpdateStatus`.

How do we tell the server that the worker is draining? (i.e. still alive and processing jobs, but don't send it additional ones)

What about updating the server about the current jobs it's processing?

livekit_agent.proto
```diff
@@ -74,25 +78,49 @@ enum JobStatus {
...
+    double current_load = 1;
+}
+message PingResponse {}
```
davidzhao, line 90: what do we use the response for?

davidzhao (commented):
Looks good!

I think workers should also report job completion to the server too, so it becomes easier to determine how each job is going.

livekit_agent.proto
theomonnom: There is already the should_retry value when accepting a specific job. Mb this should be renamed?

livekit_agent.proto
theomonnom: Added tts

livekit_agent.proto
theomonnom: Added WorkerStatusUpdate.
Also do we want to more WorkerState? 
Just thinking out loud maybe the worker could just decline the job inside the request_handler

theomonnom:
Added the load of a job on the JobStatusUpdate message.
I think both WorkerStatusUpdate and JobStatusUpdate will be sent periodically

livekit_agent.proto
```diff
@@ -55,6 +60,7 @@ message ServerMessage {
         // server asks worker to confirm availability for a job
         AvailabilityRequest availability = 2;
         JobAssignment assignment = 3;
+	WorkerPong ping = 4;
```
davidzhao, line 63:
```suggestion
	WorkerPong pong = 4;
```

livekit_agent.proto
davidzhao: ah ha.. ok. maybe `supports_resume`? (signals if the current job can be retried..) having it on the Job is even more flexible than doing it on worker.

thinking `resume` because you have `is_resume` on the job request itself.

livekit_agent.proto
```diff
@@ -74,38 +80,72 @@ enum JobStatus {
...
+    optional string namespace = 6;
 }
 
 message RegisterWorkerResponse {
     string worker_id = 1;
     string server_version = 2;
 }
 
+message MigrateJobRequest {
+    Job job = 1;
```
davidzhao, line 112: should this just be a job_id? if server has knowledge of all of the jobs, then it seems that just a job id is sufficient.

livekit_agent.proto
```diff
@@ -74,38 +80,72 @@ enum JobStatus {
...
+    bool resuming = 2;
 }
 
 message AvailabilityResponse {
     string job_id = 1;
     bool available = 2;
+    bool should_retry = 3;
+
+    // How much the worker thinks the job will cost in terms of resources
+    double job_cost = 4;
```
davidzhao, line 129: would this be required?

livekit_agent.proto
```diff
@@ -74,38 +80,72 @@ enum JobStatus {
...
     string user_data = 4;
+    
+    // The load that the job currently has on the worker
+    double load = 5;
+}
+
+message WorkerStatusUpdate {
+    WorkerStatus status = 1;
+    double load = 1;
+    int32 running_jobs = 2;
```
davidzhao, line 145: I think sending ids instead of a count could allow us to reconcile the server understanding of jobs that are running on each client. otherwise if server keeps track of this separately.. they might get out of sync.

livekit_agent.proto
theomonnom: renamed to supports_resume

livekit_agent.proto
theomonnom: yeah sgtm

livekit_agent.proto
theomonnom: Changed to a list of jobs but do you think this should be sent periodically?
WorkerStatusUpdate is a periodic update atm

livekit_agent.proto
theomonnom: After receiving a list of worker that can handle this job, the job_cost could help to pick the worker we want to chose. This may be useful for users that want that prefer to accept a job on a specific worker.
So this is indeed optional. This is similar of what psrpc is doing with "affinity" I think.
Wdyt? The naming here may not be very good

davidzhao: if the worker has already accepted the job, then it would mean that it has capacity for it (it should be up to the worker to decide)

psrpc affinity is using current load metrics to decide.

davidzhao (approved):

frostbyte73 (approved):

PR #601, opened 2024-02-09
Add js protocol package

lukasIO:
The main points:

- publish `@livekit/protocol` as its own package to be consumed by other JS libs
- generate the files on every push to `main` and commit them
- on tag pushes, the CI will also 
   1. version the package 
   2. publish the new version to NPM
- the package exposes its own version (which will be semver) through `import {version} from '@livekit/protocol'`


theomonnom (approved):

theomonnom:
if the server-sdk-js is going to use this package, every proto should be generated (ingress + egress + sip)

js/src/index.js
```diff
@@ -0,0 +1,3 @@
+export * from "./gen/livekit_models_pb.js";
```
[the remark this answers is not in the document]
lukasIO: yeah, you're right! added those

lukasIO:
that makes sense, do we need the `livekit_internal.proto`, too?

theomonnom:
I don't think we'll need it, is is indeed internal

packages/js/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@livekit/protocol",
+  "version": "13.0.0",
```
davidzhao, line 3: should the version match whatever the tag is? to keep things consistent?

packages/js/src/gen/livekit_webhook_pb.d.ts
```diff
@@ -0,0 +1,105 @@
+// Copyright 2023 LiveKit, Inc.
```
davidzhao, line 1: Nit: I think it's worth adding a `.gitattributes` file with something like this:

```
packages/js/src/gen/* linguist-generated=true
```

This would hint to GH to not show diffs for them

davidzhao (approved):
lg!

packages/js/package.json
lukasIO: yeah, it will once we trigger the workflow for the first time! the `Version Package` step takes care of that.
Was just thinking it would make sense that the major version of the tag reflects our "protocol version" that we bump for changes.

PR #609, opened 2024-02-21
allow intercepting logging

paulwe:


logger/zaputil/sampler.go
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2016-2022 Uber Technologies, Inc.
```
boks1971, line 1: Is this directly from source? The copyright?

boks1971: Oh, I see the note below 👍

boks1971 (approved):
Feels like this logging stuff should become a module of its own. It is getting a lot of features, feels like a standalone module could be useful in general.

PR #612, opened 2024-02-23
Add ImageOutput to egress EncodedOutput utils

biglittlebigben:


egress/types.go
```diff
@@ -35,12 +36,20 @@ type EncodedOutput interface {
 	GetFileOutputs() []*livekit.EncodedFileOutput
 	GetStreamOutputs() []*livekit.StreamOutput
 	GetSegmentOutputs() []*livekit.SegmentedFileOutput
+	ImageOutput
 }
 
 type EncodedOutputDeprecated interface {
```
frostbyte73, line 42: This is for deprecated requests which don't implement `Get_Outputs()`

egress/types.go
```diff
@@ -35,12 +36,20 @@ type EncodedOutput interface {
...
 	GetFile() *livekit.EncodedFileOutput
 	GetStream() *livekit.StreamOutput
 	GetSegments() *livekit.SegmentedFileOutput
+	GetFileOutputs() []*livekit.EncodedFileOutput
+	GetStreamOutputs() []*livekit.StreamOutput
+	GetSegmentOutputs() []*livekit.SegmentedFileOutput
+}
+
+type ImageOutput interface {
+	GetImageOutputs() []*livekit.ImageOutput
```
frostbyte73, line 52: Is there a reason for this to be separate?

frostbyte73: nvm, we'll now have 2 kinds of deprecated requests - the oldest having only GetX, one having GetXOutputs without Images, and the newest having GetXOutputs including images

biglittlebigben:
The answer to both remarks is that I was hoping to replace the types in https://github.com/livekit/egress/blob/2b48c8a5b576017cdfe71944dcb9713a5ca17eb8/pkg/config/output.go#L43 with the ones from protocol above to avoid type duplication.

That being said, after checking, the EncodedOutputDeprecated type in egress never uses the Get_Outputs getters, or the ImageOutput interface directly, so we can clean this up indeed.

egress/types.go
davidzhao: I think it's ok to leave it out of the legacy API? i.e. if you want new features then use the new API.

frostbyte73 (approved):

PR #621, opened 2024-02-26
Update usage of core.Fuse

dennwc:
Constructor for `core.Fuse` was removed in favor of using zero value directly (similar to `sync.Mutex` and friends).


dennwc:
CI fails because psrpc needs an update as well: https://github.com/livekit/psrpc/pull/83

biglittlebigben (approved):
Please update the dependencies that cause the unit test to fail as well.

PR #623, opened 2024-02-29
Expose trunks/dispatch validation for API/CLI. Deprecate regexps for …

dennwc:
Expose trunk/dispatch validation for API/CLI. Also, deprecate regexps for inbound - we can't validate them properly. Will keep evaluating existing ones, but reject new Trunks with regexps.


biglittlebigben (approved):
LGTM, but nut super familiar with the trunk rules...

PR #629, opened 2024-03-06
Separate protobufs and set up changesets publishing workflow

lukasIO:
Set up changesets to track both the go module and the javascript package. 
I  declared them as `fixed` in the changesets config so that if either one gets version bumped, the other will as well.

Moved all .proto files into `./protobufs`.

The release action will automatically publish the javascript package to npm and generate a git tag (which is practically also a release of the go module) whenever the `Version Packages` PR gets merged into main. 
This PR will automatically be created when the release action detects a new changeset in the repo.

For changes that should cause a version bump, either 
- run `pnpm changeset` locally within the repo and commit the changeset or
- use the changesets github bot that will automatically comment on PRs to create a changeset on the fly


paulwe (approved):

.github/workflows/buildtest.yaml
```diff
@@ -17,41 +17,43 @@ name: Test
...
+            ~/go/bin
+            ~/.cache
+          key: livekit-protocol
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.21
+          cache-dependency-path: |
+            packages/golang/go.sum
```
dennwc, line 42: This is no longer necessary, right?

package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "github.com/livekit/protocol",
+  "private": true,
+  "version": "1.10.1",
```
dennwc, line 4: Would we need to update the version here when tagging a new release?

Same question for `packages/javascript/package.json`.

What's the difference between the two, btw? It's for compatibility, I assume?

dennwc (approved):
Only one nit for `cache-dependency-path`, LGTM!

.github/workflows/buildtest.yaml
lukasIO: ah, thanks! thought I caught all instances of that. fixed in [323b7c6](https://github.com/livekit/protocol/pull/629/commits/323b7c6b0969173afd0c60e972aaade14a6b9b4f)

package.json
lukasIO: No, this will automatically be done by the changesets PR. 
[Here's](https://github.com/livekit/client-sdk-js/pull/1066/files) an example of what that looks like on the client-sdk-js. 
Once merged, the release action will create the git tag. 

One (the one on root) is for tracking the changes of the JS package, the other one (in root) will track the golang module.

lukasIO:
hm, not sure why the test action started to fail with [323b7c6](https://github.com/livekit/protocol/pull/629/commits/323b7c6b0969173afd0c60e972aaade14a6b9b4f). Failures look unrelated

lukasIO:
re-run of the tests was successful

PR #646, opened 2024-03-12
Add a v-version tag for golang module

lukasIO:
When running the publish+release action, the changesets publish command will automatically create a github tag for each package within the monorepo. The format for this is `packageName@x.x.x`. Which arguably makes a lot of sense for monorepos, but because we also want to use the git tags for golang module package releases we need a git tag with the format `vX.X.X`. 

This PR creates that additional `v` version tag manually whenever the changeset action has created another release tag.

### DIFF
--- a/.changeset/loud-goats-dream.md
+++ b/.changeset/loud-goats-dream.md
@@ -1,0 +1,6 @@
+---
+"github.com/livekit/protocol": minor
+"@livekit/protocol": minor
+---
+
+Align package version with manually tagged golang module
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -47,6 +47,6 @@ jobs:
         uses: rickstaa/action-create-tag@v1
         id: tag_create
         with:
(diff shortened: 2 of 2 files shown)


dennwc:
Please note that the latest tagged version in the repo is now `v1.11.0`, while all automated tags were `v1.10.x`.  I think before merging it we should at adjust version numbers to match. Otherwise we'll have to do ugly tricks if it publishes current `main` as Go package version `v1.10.x`.

lukasIO:
thanks for calling that out, that makes sense! I'll add a `minor` changeset to this PR so that the next release will be `v1.11.0`

dennwc:
But it must be either `v1.12.0` or `v1.11.1`. If we retag `v1.11.0` this will also be a mess. Go monitors all packages for changes. If hash of a previous release changes, it would error out.

lukasIO:
I don't think it will retag with v1.11.0. It will just skip it for now as the tag is already present.

lukasIO:
for reference: https://github.com/rickstaa/action-create-tag?tab=readme-ov-file#tag_exists_error

PR #647, opened 2024-03-13
Transcription protocol

davidzhao:
Creating a protocol for agents/clients to exchange transcription data. Neil and I came up with this today.

Transcription from each section of speech will be assigned a unique ID, and attributed it to a track/participant. Transcriptions will contain segments of text. Each segment is tagged with start/end times as well as if the segment has been deemed final by the underlying STT engine.

When a segment of text becomes final, it will not be changed. Otherwise, they could be updated either with new tokens or existing token updates.

The same protocol can be used for translations with the inclusion of `language` code. A backend agent could send multiple transcriptions (either selectively or broadcast), and the frontend could choose which one they would want to display.

Any comments or feedback, lmk!


protobufs/livekit_models.proto
```diff
@@ -257,6 +259,21 @@ message SipDTMF {
...
+  repeated TranscriptionSegment segments = 4;
+  string language = 5;
+}
+
+message TranscriptionSegment {
+  string text = 1;
+  int64 start_time = 2;
+  int64 end_time = 3;
+  bool final = 4;
+}
```
boks1971, line 275: Is it possible to send the unique ID of the segment in the protocol? I am guessing that will make it easier for clients to look up existing segments and update UI. Else, they would have to hash on combination of `start_time` + `end_time` which would work, but not as easy it feels like.

boks1971: Never mind, I see the ID in the above message.

boks1971: Do segments have unique IDs too? Or clients have to look through all segments to update one?

davidzhao: we can def think about generating one.. currently I was thinking the indices of the segments would be stable.. so it would be transcription_id+segment_idx

boks1971: oh all segments are transmitted with every message. Wouldn't that be a bit wasteful?

davidzhao: only for this transcription section (or paragraph). it's up to us to define how long that is.. it could be a single sentence, or until the user has a meaningful pause.

since we are sending over data channel. a user who joins halfway through the section could be missing some context. in terms of the overhead, I would think that a section is no more than 30s of speech.. so that's about 75 words if someone's speaking fast, and ~450 chars.

while we'll update when new tokens appear, we could rate limit it to every second.

lukasIO: +1 for an id per transcription segment. 
Maybe I'm not understanding it correctly, but due to each segment having the `final` flag, I imagine it to be cleaner to handle if we can match the id for segment updates. Also imagine that the start and end time might change for non-final segments? Or would only the last index of the segments array be allowed to be `final: false`?

boks1971: I believe (although I am not 100%) any segment can change Lukas.

DZ is suggesting clients index on `transcription_id` which is in the envelope (you can think of that envelope as one paragraph, but we can change it to one sentence if need be) and that envelope has an array of segments. Although I am not sure if anything but the last one can change, but even if it does, all segments getting sent down with the envelope on any change means the ordering of that array is in tact. So, clients can index on `trasciption_id`(from envelope) + `segment_idx`(index into the segments array in the envelope).

lukasIO: thanks, Raja! 
that makes sense, but then why have the `final` flag on the segment instead of on the envelope?

boks1971: Each segment can change or be in final state. Once the transcription service is confident (or if it cannot do any more) that a segment is finished, it will mark as final. Some prior segment may change based on context that is coming later (am guessing a bit here based on things I have read here and there). So, a later segment could be clear and it is done transcription and will be marked `final: true`. But, an earlier segment may have had low confidence interval and based on context of a later segment it might be modified.

`final` indicates that the segment is done and will not change any more. Maybe, `final` is a bit confusing and maybe there is a better name?

protobufs/livekit_models.proto
```diff
@@ -257,6 +259,21 @@ message SipDTMF {
   string digit = 4;
 }
 
+message Transcription {
+  string id = 1;
+  string participant_identity = 2;
+  string track_id = 3;
+  repeated TranscriptionSegment segments = 4;
```
theomonnom, line 266: segments are more like alternatives

davidzhao:
Benjamin has merged a separate PR that supersedes this, closing.

PR #650, opened 2024-03-13
Add bounds check when trimming SIP number

dennwc:


dennwc:
@lukasIO Should your change (#646) go first, so that we don't have to sync the version number again?

sip/sip.go
```diff
@@ -381,7 +381,11 @@ func EvaluateDispatchRule(rule *livekit.SIPDispatchRuleInfo, req *rpc.EvaluateSI
 	if rule.HidePhoneNumber {
 		// TODO: Decide on the phone masking format.
 		//       Maybe keep regional code, but mask all but 4 last digits?
-		from = from[len(from)-4:]
+		n := 4
+		if len(from) <= 4 {
+			n = 1
+		}
+		from = from[len(from)-n:]
```
lukasIO, line 388: I'm missing context, just wanted to double check that it is intended behaviour to only show 1 digit for short sip numbers?

dennwc: Short numbers in SIP usually come from local devices or in test environment (`10X`). So it's mostly just to illustrate that yes, trunk is hiding the caller number.

lukasIO (approved):

PR #651, opened 2024-03-13
Add signal requests for local track updates

lukasIO:


protobufs/livekit_rtc.proto
```diff
@@ -50,6 +50,10 @@ message SignalRequest {
     // update a participant's own metadata and/or name
     UpdateParticipantMetadata update_metadata = 15;
     Ping ping_req = 16;
+    // Update local audio track settings
+    UpdateLocalAudioTrack local_audio_track = 17;
```
davidzhao, line 54: nit: update_audio_track?

davidzhao (commented):
lg!

boks1971 (approved):
How is this intended to be used? What will the server do when this is received?

protobufs/livekit_rtc.proto
```diff
@@ -223,6 +227,20 @@ message UpdateTrackSettings {
...
+message UpdateLocalAudioTrack {
+  string track_sid = 1;
+  repeated AudioTrackFeature features = 2;
+}
+
+message UpdateLocalVideoTrack {
+  string track_sid = 1;
+  uint32 width = 2;
+  uint32 height = 3;
+  repeated VideoTrackFeature features = 4;
```
boks1971, line 241: If this is simulcast, may be have `width/height` per layer?

lukasIO: tbh, I'm not sure if `simulcast` is even desired at this point. Is it something that we want to be able to turn on/off during a publications' lifecycle? 
cc @davidzhao

boks1971: I think changing publication (any feature) mid-stream is a discussion worth having. There are things that cannot be changed without re-acquiring a stream (things like AGC/echo cancellation for audio).

For simulcast specific, on the wall on that. Ideally, we don't even have to explicitly specify `simulcast` flag/feature. Based on number of layers, it can be implicit. 

But, there is a need to distinguish between VP9/AV1 Simulcast vs SVC. If/when we support VP9/AV1 simulcast, we will need to indicate if it is simulcast or SVC.  So, maybe, we can continue to have that flag. If VP9 simulcast, that flag can indicate it is simulcast, else treated as SVC.

lukasIO: to keep it simple I removed (after discussing with dz) the `VideoFeature` part from the PR. 
For the beginning mostly resolution updates will be relevant for video tracks. We can add other things like layers later on if we see the need for it.

PR #657, opened 2024-03-14
call config observers synchronously

paulwe:


davidzhao (approved):

boks1971 (approved):
What was breaking due to async/out-of-order?

PR #668, opened 2024-03-26
Support for converting Logger into slog.Handler

dennwc:
This change allows to take an existing `Logger` and convert it into `slog.Handler`.

The reason behind this is that eventually Go ecosystem will likely converge on `slog` as a default logging interface. For example `pion/logging` could provide a wrapper for it, instead of us defining our own (`logger/pionlogger`). Same for SIP library that we use. 

The new `ToSlogHandler` function will check if the `Logger` implementation is a known one (Zap or Logr) and will convert it with package-specific functions. Both loggers that we use support slog natively now.

This change doesn't affect how our logger is configured or used. We will still configure it the old way and then it could be converted to slog when needed.


paulwe (approved):
lgtm

PR #670, opened 2024-03-28
add hedge retry helper

paulwe:


utils/hedge.go
```diff
@@ -0,0 +1,57 @@
...
+
+	for {
+		select {
+		case <-delay.C:
+			go race()
+			if attempt++; attempt < params.MaxAttempts {
+				delay.Reset(params.RetryDelay)
+			}
+		case res := <-ch:
+			if res.err == nil || params.IsRecoverable == nil || !params.IsRecoverable(err) {
```
boks1971, line 48: should this be checking `res.err` for `IsRecoverable`?

utils/hedge.go
```diff
@@ -0,0 +1,57 @@
...
+	RetryDelay    time.Duration
+	MaxAttempts   int
+	IsRecoverable func(err error) bool
+	Func          func(context.Context) (T, error)
+}
+
+// race retries if the function takes to long to return
+// |---------------- attempt 1 ----------------|
+// |    delay    |--------- attempt 2 ---------|
+func HedgeCall[T any](ctx context.Context, params HedgeParams[T]) (v T, err error) {
```
boks1971, line 21: There is a named return for `err` and use of `err` inside also. I think it is fine. Just wanted to flag it because we got bitten by it recently :-)

boks1971 (approved):

utils/hedge.go
paulwe: `err` accumulates errors from `res.err` and if the call fails it returns all of the errors that happened during the call. but if we reach the return condition we only want to return the value and error from res

utils/hedge.go
paulwe: yes

paulwe: thatnks, good catch 🙏

utils/hedge.go
paulwe: oh the `err` in the closure. no that won't affect the err from the return

PR #675, opened 2024-04-05
add Clock interface and LeakyBucket

lherman-cs:
Modified version of https://github.com/uber-go/ratelimit/tree/main. 

Main changes:
  * Allow SetRateLimit on the fly
  * Remove other variants
  * Simplified configuration

### DIFF
--- a/utils/clock.go
+++ b/utils/clock.go
@@ -1,0 +1,20 @@
+package utils
+
+import "time"
+
+type Clock interface {
+	Now() time.Time
+	Sleep(time.Duration)
+}
+
+type SystemClock struct{}
+
+var _ Clock = &SystemClock{}
+
+func (clock *SystemClock) Now() time.Time {
+	return time.Now()
+}
+
+func (clock *SystemClock) Sleep(d time.Duration) {
+	time.Sleep(d)
+}
--- a/utils/rate.go
+++ b/utils/rate.go
@@ -1,0 +1,92 @@
+// Copyright (c) 2022 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// EDIT: slight modification to allow setting rate limit on the fly
+// SCOPE: LeakyBucket
+package utils
+
+import (
+	"sync/atomic"
+	"time"
+)
+
+type LeakyBucket struct {
+	//lint:ignore U1000 Padding is unused but it is crucial to maintain performance
+	// of this rate limiter in case of collocation with other frequently accessed memory.
+	prepadding [64]byte // cache line size = 64; created to avoid false sharing.
+	state      int64    // unix nanoseconds of the next permissions issue.
+	//lint:ignore U1000 like prepadding.
+	postpadding [56]byte // cache line size - state size = 64 - 8; created to avoid false sharing.
+
+	perRequest time.Duration
+	maxSlack   time.Duration
+	clock      Clock
+}
+
+func NewLeakyBucket(rateLimit int, slack time.Duration, clock Clock) *LeakyBucket {
+	var lb LeakyBucket
+	lb.SetRateLimit(rateLimit)
+	lb.maxSlack = slack * lb.perRequest
+	lb.clock = clock
+	atomic.StoreInt64(&lb.state, 0)
+	return &lb
+}
+
+func (lb *LeakyBucket) SetRateLimit(rateLimit int) {
+	lb.perRequest = time.Second / time.Duration(rateLimit)
+}
+
+// Take blocks to ensure that the time spent between multiple
+// Take calls is on average time.Second/rate.
+func (t *LeakyBucket) Take() time.Time {
+	var (
+		newTimeOfNextPermissionIssue int64
+		now                          int64
+	)
+	for {
+		now = t.clock.Now().UnixNano()
+		timeOfNextPermissionIssue := atomic.LoadInt64(&t.state)
+
+		switch {
+		case timeOfNextPermissionIssue == 0 || (t.maxSlack == 0 && now-timeOfNextPermissionIssue > int64(t.perRequest)):
+			// if this is our first call or t.maxSlack == 0 we need to shrink issue time to now
+			newTimeOfNextPermissionIssue = now
+		case t.maxSlack > 0 && now-timeOfNextPermissionIssue > int64(t.maxSlack)+int64(t.perRequest):
+			// a lot of nanoseconds passed since the last Take call
+			// we will limit max accumulated time to maxSlack
+			newTimeOfNextPermissionIssue = now - int64(t.maxSlack)
+		default:
+			// calculate the time at which our permission was issued
+			newTimeOfNextPermissionIssue = timeOfNextPermissionIssue + int64(t.perRequest)
+		}
+
+		if atomic.CompareAndSwapInt64(&t.state, timeOfNextPermissionIssue, newTimeOfNextPermissionIssue) {
+			break
+		}
+	}
+
+	sleepDuration := time.Duration(newTimeOfNextPermissionIssue - now)
+	if sleepDuration > 0 {
+		t.clock.Sleep(sleepDuration)
+		return time.Unix(0, newTimeOfNextPermissionIssue)
+	}
+	// return now if we don't sleep as atomicLimiter does
+	return time.Unix(0, now)
+}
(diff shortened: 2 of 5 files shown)


utils/rate.go
```diff
@@ -0,0 +1,92 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// EDIT: slight modification to allow setting rate limit on the fly
+// SCOPE: LeakyBucket
+package utils
+
+import (
+	"sync/atomic"
```
boks1971, line 26: we use `uber/atomic` everywhere instead of `sync/atomic`. We went through a round of changing everything a couple of years back. Guess Uber does not do that :-)

utils/rate.go
```diff
@@ -0,0 +1,92 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// EDIT: slight modification to allow setting rate limit on the fly
```
boks1971, line 21: Can you please add a link to the source where this is taken from?

Is the only change to add `SetRateLimit` API compared to the original?

utils/rate.go
```diff
@@ -0,0 +1,92 @@
...
+	return &lb
+}
+
+func (lb *LeakyBucket) SetRateLimit(rateLimit int) {
+	lb.perRequest = time.Second / time.Duration(rateLimit)
+}
+
+// Take blocks to ensure that the time spent between multiple
+// Take calls is on average time.Second/rate.
+func (t *LeakyBucket) Take() time.Time {
```
boks1971, line 58: Make consistent use of name. Use `lb` for struct variable name consistently. Guessing `staticcheck` may report it?

utils/rate.go
boks1971: uber/atomic is nicer to use

utils/rate.go
```diff
@@ -0,0 +1,92 @@
...
+	)
+	for {
+		now = t.clock.Now().UnixNano()
+		timeOfNextPermissionIssue := atomic.LoadInt64(&t.state)
+
+		switch {
+		case timeOfNextPermissionIssue == 0 || (t.maxSlack == 0 && now-timeOfNextPermissionIssue > int64(t.perRequest)):
+			// if this is our first call or t.maxSlack == 0 we need to shrink issue time to now
+			newTimeOfNextPermissionIssue = now
+		case t.maxSlack > 0 && now-timeOfNextPermissionIssue > int64(t.maxSlack)+int64(t.perRequest):
```
boks1971, line 71: `perRequest` is not locked and also not atomic. Are there any down sides to this running and rate limit being set on the fly?

utils/rate.go
```diff
@@ -0,0 +1,92 @@
...
+
+	perRequest time.Duration
+	maxSlack   time.Duration
+	clock      Clock
+}
+
+func NewLeakyBucket(rateLimit int, slack time.Duration, clock Clock) *LeakyBucket {
+	var lb LeakyBucket
+	lb.SetRateLimit(rateLimit)
+	lb.maxSlack = slack * lb.perRequest
```
boks1971, line 46: In the previous implementation, remember this being -ve. Has that changed?

utils/rate_test.go
```diff
@@ -0,0 +1,497 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+// EDIT: slight modification to allow setting rate limit on the fly
```
boks1971, line 21: can you add source of this?

also, were you able to augment tests with rate limit change on the fly? I am not reading this as this looks like some existing lib.

utils/rate.go
lherman-cs: I did add a test here with a simulated clock, https://github.com/livekit/cloud-protocol/pull/747/files#diff-bf1c81b11717ede4526978ba2d3024bb771a5732c0768b6bafebb86af9f6f516R218-R241

I'll look into adding a unit test here as well

utils/rate.go
```diff
@@ -0,0 +1,92 @@
...
+	"time"
+)
+
+type LeakyBucket struct {
+	//lint:ignore U1000 Padding is unused but it is crucial to maintain performance
+	// of this rate limiter in case of collocation with other frequently accessed memory.
+	prepadding [64]byte // cache line size = 64; created to avoid false sharing.
+	state      int64    // unix nanoseconds of the next permissions issue.
+	//lint:ignore U1000 like prepadding.
+	postpadding [56]byte // cache line size - state size = 64 - 8; created to avoid false sharing.
```
paulwe, line 36: do we really need padding before? should we use `x/sys/cpu.CacheLinePad` for portability?

utils/rate.go
```diff
@@ -0,0 +1,92 @@
...
+	var lb LeakyBucket
+	lb.SetRateLimit(rateLimit)
+	lb.maxSlack = slack * lb.perRequest
+	lb.clock = clock
+	atomic.StoreInt64(&lb.state, 0)
+	return &lb
+}
+
+func (lb *LeakyBucket) SetRateLimit(rateLimit int) {
+	lb.perRequest = time.Second / time.Duration(rateLimit)
```
paulwe, line 53: are `SetRateLimit` and `Take` meant to be externally synchronized? the go race detector will raise errors about this. we don't seem to have a test for this

paulwe (approved):

utils/rate.go
lherman-cs: I didn't have my IDE integrated with `staticcheck`, this is a great idea.

lherman-cs: Renamed to `lb`.

utils/rate.go
lherman-cs: Replaced with uber/atomic

utils/rate.go
lherman-cs: Source: https://github.com/uber-go/ratelimit/blob/main/limiter_atomic_int64.go. I'll add this in the comment too.

lherman-cs: Yes, it is only adding SetRateLimit. Per comments in this PR, the change will evolve to modify the internal logic a bit.

utils/clock.go
```diff
@@ -0,0 +1,30 @@
...
+
+var _ Clock = &SystemClock{}
+
+func (*SystemClock) Now() time.Time {
+	return time.Now()
+}
+
+func (*SystemClock) Sleep(d time.Duration) {
+	time.Sleep(d)
+}
```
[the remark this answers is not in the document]
lherman-cs: good point. It's an empty struct and there's no state management.

utils/rate.go
lherman-cs: Yes, there were meant to be externally synchronized. But, rethinking about this, it would make more sense to handle the synchronization here, especially `Take` is already synchronized. I'll work on this.

utils/rate.go
lherman-cs: I think misunderstood the question. It was originally meant to be synchronized externally. @paulwe also raised a similar concern, https://github.com/livekit/protocol/pull/675#discussion_r1556084383. At the end, I think I should handle the synchronization here after all.

utils/rate.go
lherman-cs: Spent too much time to make this work with atomic. At the end, the code became complicated since `state` has occupied 64 bits. It's possible to cramp the rate limit state to the 64 bits, but it has to be u64. Then, handling time without negative numbers became tricky... 

At the end, I reverted to the good ol' mutex. For now, we're not going to run into a performance issue since we only use this in QoS from a single goroutine. In the future, we might to revisit if there's a need for large amount of goroutines accessing this.

utils/rate_test.go
lherman-cs: added source and added a test for setting rate limit

utils/rate.go
lherman-cs: no longer needed.

utils/rate.go
lherman-cs: @paulwe added a test

utils/rate.go
lherman-cs: Replaced the synchronization handling with mutex instead. I put the details in @paulwe comment thread

utils/rate.go
lherman-cs: reverted to the original

utils/rate.go
```diff
@@ -0,0 +1,97 @@
...
+	// We shouldn't allow sleepFor to get too negative, since it would mean that
+	// a service that slowed down a lot for a short period of time would get
+	// a much higher RPS following that.
+	if lb.sleepFor < lb.maxSlack {
+		lb.sleepFor = lb.maxSlack
+	}
+
+	// If sleepFor is positive, then we should sleep now.
+	if lb.sleepFor > 0 {
+		lb.clock.Sleep(lb.sleepFor)
```
boks1971, line 89: did you determine that sleeping with lock held is fine?

boks1971: If it is fine, a comment here explaining why it is okay to sleep with lock held. 

From experience, three months down the line if somebody is debugging some hairy issue, this (i. e. sleep with lock held) will pop up as somewhat of a mystery and if this is contributing to some performance issue or some bad behaviour or some bug. 

You have mentioned `BLOCKING`, but an explicit ack that there is a sleep with lock held and it is by design would give more confidence in somebody reading this three months later :-)

boks1971 (approved):
lgtm! just one comment about sleeping with lock held.

utils/rate.go
lherman-cs: Synced offline. For now, this is good enough for performance and safety. We will revisit this if needed.

PR #679, opened 2024-04-12
Transcription/Timed metadata protocol

biglittlebigben:


protobufs/livekit_models.proto
```diff
@@ -248,15 +250,35 @@ message UserPacket {
...
+  // NEXT_ID: 11
 }
 
 message SipDTMF {
   uint32 code = 3;
   string digit = 4;
 }
 
+message Transcription {
+  string participant_identity = 2;
```
biglittlebigben, line 268: This entry was taken from the initial proposal. Is it needed considering that the top level `DataPacket` message provide a `participant_identity` field as well?

davidzhao: yeah, I think it's needed within as well. in DataPacket, `participant_identity` represents the user that sent it. here it represents the user who's transcription we are providing. For example, if a backend agent is performing transcriptions, the outer envelope would be the Agent's identity, and here it would be the track publisher's.

davidzhao (approved):
🚀

PR #680, opened 2024-04-15
WHIP protocol change

biglittlebigben:
This PR adds an optional "enable_transcoding" entry to the ingress APIs and deprecates the bypass_transcoding entry. This is in preparation of disabling "enable_transcoding" by default for WHIP.


protobufs/livekit_ingress.proto
```diff
@@ -45,10 +45,12 @@ message CreateIngressRequest {
   string participant_name = 5;
   // metadata associated with the publishing participant
   string participant_metadata = 10;
-  // whether to pass through the incoming media without transcoding, only compatible with some input types
-  bool bypass_transcoding = 8; 
+  // bool bypass_transcoding = 8 [deprecated = true];
```
dennwc, line 48: I assume that it's okay here that we do not consider old `bypass_transcoding` when `bypass_transcoding` is not set.

dennwc (approved):

dennwc:
Also please don't forget to add a changeset!

protobufs/livekit_ingress.proto
biglittlebigben: Thinking again about the how we would roll this out, I only deprecated the bypass_transcoding entries instead of removing them. This allows keep using older tools/SDKs after we roll out the server changes.

protobufs/livekit_ingress.proto
```diff
@@ -113,6 +113,9 @@ message IngressVideoEncodingOptions {
   repeated VideoLayer layers = 3;
 }
 
+message IngressVideoEncodingPassthrough {
```
frostbyte73, line 116: what's this for?

biglittlebigben: Leftover from a previous version. I removed it. Thanks.

protobufs/livekit_ingress.proto
```diff
@@ -45,10 +45,12 @@ message CreateIngressRequest {
   string participant_name = 5;
   // metadata associated with the publishing participant
   string participant_metadata = 10;
-  // whether to pass through the incoming media without transcoding, only compatible with some input types
-  bool bypass_transcoding = 8; 
+  bool bypass_transcoding = 8 [deprecated = true];
+  optional bool enable_transcoding = 11;
```
davidzhao, line 49: can we add docstring here explaining the nuance of this param?

PR #683, opened 2024-04-16
support manual dispatch to agent workers

theomonnom:
See https://github.com/livekit/agents/issues/214


protobufs/livekit_agent.proto
```diff
@@ -78,6 +78,7 @@ message ServerMessage {
 enum JobType {
     JT_ROOM = 0;
     JT_PUBLISHER = 1;
+    JT_CUSTOM = 2;
```
keepingitneil, line 81: maybe JT_MANUAL? Custom implies it still happens automatically but with customer conditions.

protobufs/livekit_room.proto
```diff
@@ -58,7 +59,10 @@ service RoomService {
   rpc SendData(SendDataRequest) returns (SendDataResponse);
 
   // Update room metadata, will cause updates to be broadcasted to everyone in the room, Requires `roomAdmin`
-  rpc UpdateRoomMetadata (UpdateRoomMetadataRequest) returns (Room);
+  rpc UpdateRoomMetadata(UpdateRoomMetadataRequest) returns (Room);
+
+  // Start a new custom agent job
+  rpc StartAgentJob(StartAgentJobRequest) returns (StartAgentJobResponse);
```
davidzhao, line 65:
```suggestion
  rpc DispatchAgent(DispatchAgentRequest) returns (DispatchAgentResponse);
```

davidzhao (commented):
this makes a lot of sense. there are many situations where manual dispatching is preferred.

PR #688, opened 2024-04-18
Add SIP analytics

dennwc:


biglittlebigben (approved):

protobufs/livekit_analytics.proto
```diff
@@ -151,8 +155,9 @@ message AnalyticsEvent {
   string error = 20;
   RTPStats rtp_stats = 21;
   int32 video_layer = 22;
+  SIPCallInfo sip_call = 24;
```
davidzhao, line 158: we also need to send analytics a copy of the trunks. this will allow us to keep historical records

shishirng (approved):

PR #692, opened 2024-04-21
reduce allocations from guids

paulwe:
```
BenchmarkNewGuid
BenchmarkNewGuid/old
BenchmarkNewGuid/old-10         	  290210	      3458 ns/op	    2682 B/op	     100 allocs/op
BenchmarkNewGuid/new
BenchmarkNewGuid/new-10         	11576660	        92.23 ns/op	      16 B/op	       1 allocs/op
```

### DIFF
--- a/utils/id.go
+++ b/utils/id.go
@@ -15,14 +15,19 @@
 package utils
 
 import (
+	"crypto/rand"
 	"crypto/sha1"
+	"encoding/binary"
 	"fmt"
 	"os"
+	"sync"
+	"time"
 
 	"github.com/jxskiss/base62"
 	"github.com/lithammer/shortuuid/v4"
 
 	"github.com/livekit/protocol/livekit"
+	"github.com/livekit/protocol/logger"
 )
 
 const GuidSize = 12
@@ -45,8 +50,10 @@ const (
 	AgentJobPrefix        = "AJ_"
 )
 
+var defaultGuidGenerator = newGuidGenerator(4096, GuidSize+10)
+
 func NewGuid(prefix string) string {
-	return prefix + shortuuid.New()[:GuidSize]
+	return defaultGuidGenerator.NewGuid(prefix)
 }
 
 // HashedID creates a hashed ID from a unique string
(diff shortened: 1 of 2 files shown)


boks1971 (approved):
Awesome!!!

utils/id.go
```diff
@@ -66,17 +73,91 @@ func LocalNodeID() (string, error) {
...
+	mu   sync.Mutex
+	buf  []byte
+	pos  int
+}
+
+func newGuidGenerator(bufSize, scratchSize int) *guidGenerator {
+	return &guidGenerator{
+		pool: sync.Pool{
+			New: func() any {
+				b := make([]byte, scratchSize)
```
boks1971, line 98: is there a need for `scratchSize` here? Trying to understand how this size works as a desired size slice is appended when used.

utils/id.go
```diff
@@ -66,17 +73,91 @@ func LocalNodeID() (string, error) {
 	return fmt.Sprintf("%s%s", NodePrefix, HashedID(hostname)[:8]), nil
 }
 
-var b62Index = newB62Index()
-var b62Chars = []byte(shortuuid.DefaultAlphabet)
+var b57Index = newB57Index()
```
davidzhao, line 76: why do we move from 62 to 57?

utils/id.go
```diff
@@ -66,17 +73,91 @@ func LocalNodeID() (string, error) {
...
+				}
+			}
+			r >>= 6
+		}
+	}
+}
+
+func (g *guidGenerator) NewGuid(prefix string) string {
+	b := g.pool.Get().(*[]byte)
+	defer g.pool.Put(b)
```
davidzhao, line 153: not following here.. if we've put it back after returning it cast as a string, would that cause it to be overridden by the next `NewGuid` call?

utils/id.go
paulwe: yeah, this variable has been mislabeled for a while. the author of the shortuuid lib made up his own encoding and we're sort of stuck with it now.

utils/id.go
paulwe: strings are immutable and bytes are not so to avoid confusion when you cast from one to the other go always allocates. to get a string that references the memory from a byte you have to do ```
s := unsafe.String(unsafe.SliceData(b), len(b))
```

utils/id.go
paulwe: i defined it as `GuidSize+10`. i think the biggest prefix might be `TR_RDC` but i didn't think we would miss the extra 4 bytes. if we need more `NewGuid` will grow the slice it's just less efficient.

utils/id.go
davidzhao: got it. that makes sense! I had incorrectly assumed it was a straight cast.

PR #701, opened 2024-04-30
Expose Logger constructor for Zap

dennwc:
As per the community request, expose the `Logger` constructor for a pre-configured Zap logger.


logger/logger.go
```diff
@@ -200,8 +200,11 @@ type zapLogger[T zaputil.Encoder[T]] struct {
 	sampler   *zaputil.Sampler
 }
 
-func NewZapLogger(conf *Config, opts ...ZapLoggerOption) (ZapLogger, error) {
-	zap := zap.New(nil).WithOptions(zap.AddCaller(), zap.AddStacktrace(zap.ErrorLevel)).Sugar()
+func NewZap(log *zap.Logger, conf *Config, opts ...ZapLoggerOption) (ZapLogger, error) {
```
davidzhao, line 203: The two API names are too similar and I think it'll be confusing.

how about having the core method to be `FromZapLogger`, and then `NewZapLogger` would create an instance of zap and passing it in

biglittlebigben (approved):

davidzhao (approved):
lgtm!

PR #702, opened 2024-04-30
Add Twirp interceptor that prints curl commands

dennwc:
Add Twirp interceptor that prints `curl` commands for each request made. It will be then used in our CLI.


biglittlebigben (approved):

davidzhao (approved):
awesome! do you see other things using it too? or is it a cli-specific interceptor?

dennwc:
@davidzhao I was thinking about this too. Maybe it would be useful for `lkctl`, but apart from that ti could be moved to `livekit-cli` instead of the `protocol`. The only reason I put it here is because it does depend on the protocol used (Twirp).

PR #706, opened 2024-05-07
Rename participant_identity in Transcription

dennwc:
There are two fields named `participant_identity` in context of `Transcription`. It's not clear which one corresponds to the transcriber agent, and which one is for the participant which got its speech transcribed.

This change renames the `Transcription` field to `transcribed_participant_identity` so to clarify that it's used for the transcribed participant.


theomonnom:
Do we need the usecase where participant_identity is the transcriber, and the participant_identity inside the transcription is the participant that got the speech transcribed?

dennwc:
That's a very good question :thinking:

The shared identity in `DataPacket` is used by all SDKs to route this data packet to a specific `RemoteParticipant` handler. In that sense, it would be easier for end users if this would be the identity of the transciption (same as audio pre-transcription).

But the problem with this approach is that currently LK server will force-set `DataPacket.participant_identity` to match identity of the transcription _agent_. We could special-case transcriptions, but this would kind of defeat the purpose of a shared field that always means the same thing.

Thus, what you propose makes more sense. So `DataPacket.participant_identity` is the _agent_ who made the transcription, while `Transcription.participant_identity` is the identity of the participant's audio that was transcribed.
 
If that make case, let me rename `Transcription.participant_identity` to `transcribed_participant_identity` and remove the compatibility code from https://github.com/livekit/client-sdk-js/pull/1130.

biglittlebigben:
> Do we need the usecase where participant_identity is the transcriber, and the participant_identity inside the transcription is the participant that got its speech transcribed?

This is indeed the intent of the extra field.

biglittlebigben:
I have no strong opinion wrt renaming the field or not.

dennwc:
@biglittlebigben @theomonnom I removed the notice and renamed the field. WDYT?

biglittlebigben (approved):
LGTM

PR #710, opened 2024-05-09
avoid allocation from timeout queue iterator

paulwe:


utils/event_emitter.go
```diff
@@ -129,7 +133,7 @@ func (l *EventObserverList[V]) Observe() *EventObserver[V] {
 	le := l.observers.PushBack(o)
 	l.mu.Unlock()
 
-	o.stop = append(o.stop, func() { l.stopObserving(le) })
+	o.stop = func() { l.stopObserving(le) }
```
boks1971, line 136: does this need chaining like abive?

boks1971 (approved):

utils/event_emitter.go
paulwe: no, eventemitter delegates observer construction to observerlist so so it's wrapping this stop closure. there's no way stop could be defined here because we just made the observer a few lines above

PR #713, opened 2024-05-10
Adding a simulation option for leave request.

boks1971:
@lukasIO I was thinking of making the simulation option just a `LeaveRequest`. And when simulating, client side can just set `Action` field of `LeaveRequest`.
But, decided against it as I did not see the use case for it and also wanted to keep it simple/specific.

Let me know if I should change that to a generic `LeaveRequest`. If doing generic `LeaveRequest`, server behaviour would be
1. `Action = DISCONNECT`: server will close the participant and send a `LeaveRequest` with `Action = DISCONNECT`.
2. `Action = RESUME`, server will send `LeaveRequest` with `Action = RESUME` and it will not include self region in `RegionSettings`.
3. `Action = RECONNECT`, server will send `LeaveRequest` with `Action = RECONNECT` and it will not include self region in `RegionSettings`.

As the current simulation is only for full reconnect, it will behave like `#3` above.

Also, this option will be supported only in cloud.

Please let me know if I should change this or just leave it as a boolean.


lukasIO (approved):
this makes sense to me as is, thank you!

PR #717, opened 2024-05-16
Additions to agent protocol

biglittlebigben:
This PR:
- Adds StartAgentJobRequest/StopAgentJobRequest calls to start an handle manually mange an agent lifecycle
- Deprecates the JobType field in WorkerInfo as workers will now be expected to handle both kinds jobs (if the namespace supports both kinds in the first place)
- Adds a `metadata` string -> string map to Job to allow agent implementation to take anonymous (to us) parameters
- Adds a JobStatus and error fields to Job for status reporting (Similar to egress and ingress)
- Makes livekit_agent.proto indentation consistent with other proto files
- Add room configuration and management APIs to the Room service. Rooms configuration allow bundling most room configuration parameters under a single id, that can eventually be put as part of the room creation token.


paulwe (approved):
lgtm 🚀

protobufs/livekit_agent.proto
```diff
@@ -21,154 +21,174 @@ option ruby_package = "LiveKit::Proto";
...
 
+service AgentService {
+  rpc StartAgentJob(StartAgentJobRequest) returns (Job);
+  rpc StopAgentJob(StopAgentJobRequest) returns (Job); 
+}
+
+message StartAgentJobRequest {
+  JobType type = 1;
+  Room room = 2;
+  optional ParticipantInfo participant = 3;
```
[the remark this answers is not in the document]
theomonnom: Same for just the room name mb?

protobufs/livekit_agent.proto
```diff
@@ -21,154 +21,174 @@ option ruby_package = "LiveKit::Proto";
...
 
-    string participant_name = 4;
-    string participant_identity = 5;
-    string participant_metadata = 6;
+  string participant_name = 4;
+  string participant_identity = 5;
+  string participant_metadata = 6;
 }
 
 message UpdateJobStatus {
```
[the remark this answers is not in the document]
theomonnom: I think it may be OK, but we may want to implement Job metadata in the future (not a priority for now)

protobufs/livekit_agent.proto
```diff
@@ -21,154 +21,174 @@ option ruby_package = "LiveKit::Proto";
...
+  rpc StartAgentJob(StartAgentJobRequest) returns (Job);
+  rpc StopAgentJob(StopAgentJobRequest) returns (Job); 
+}
+
+message StartAgentJobRequest {
+  JobType type = 1;
+  Room room = 2;
+  optional ParticipantInfo participant = 3;
+  string namespace = 4;
+  map<string, string> metadata = 5;
```
theomonnom, line 34: Can this just be a string?

protobufs/livekit_agent.proto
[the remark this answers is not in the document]
paulwe: job metadata is immutable. the agent can use participant metadata or its own external storage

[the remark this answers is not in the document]
biglittlebigben: Stepping back: what information is conveyed in the `participant*` fields? Aren't the participants identified by the Job? If not, what participant is this referring to?

protobufs/livekit_agent.proto
biglittlebigben: We currently use a map for egress, a string for other entities (participant, etc). Either has pros and cons. The map lets us potentially stash fields in there as well using a reserved namespace. The string is more flexible...

protobufs/livekit_agent.proto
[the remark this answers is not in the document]
biglittlebigben: Yes

protobufs/livekit_agent.proto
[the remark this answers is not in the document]
paulwe: i was referring only to `UpdateJobStatus` - gh included irrelevant lines. to answer your question though - the agent service sends an availability request to the agent client. if the client is available to handle the job the response includes participant info used to create a token. the token is then passed back to the client in the job assignment

protobufs/livekit_agent.proto
paulwe: in cases where we might be tempted to extend the user supplied metadata with values from the service we should probably add structured fields. for example we might include an optional field for sip data with a struct containing phone #, trunk id, etc...

protobufs/livekit_agent.proto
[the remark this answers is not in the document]
biglittlebigben: Ok. This PR deprecates the field for now. from the above, it looks like I should rather remove it entirely?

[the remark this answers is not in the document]
paulwe: we should probably clean it up in lk first - the service implements it - but yeah, i think so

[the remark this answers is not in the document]
biglittlebigben: Sure. Next step is to implement this in lk anyway. Probably makes sense to do it before I merge this.

theomonnom (approved):

protobufs/livekit_models.proto
```diff
@@ -266,7 +266,8 @@ message SipDTMF {
 }
 
 message Transcription {
-  string participant_identity = 2;
+  // Participant that got its speech transcribed
+  string transcribed_participant_identity = 2;
```
davidzhao, line 270: seems redundant to me? everything in a Transcription should be scoped to the transcription.

protobufs/livekit_room.proto
```diff
@@ -59,11 +60,22 @@ service RoomService {
 
   // Update room metadata, will cause updates to be broadcasted to everyone in the room, Requires `roomAdmin`
   rpc UpdateRoomMetadata (UpdateRoomMetadataRequest) returns (Room);
+
+  // Create a room configuration. 
+  rpc CreateRoomConfiguration(CreateRoomConfigurationRequest) returns (RoomConfiguration);
```
davidzhao, line 65: worth discussing: should these be APIs on room service? or Project settings in Cloud (and config in OSS)?

davidzhao: if they are on room, we'd have to figure out permissions.. what token permissions would they require

protobufs/livekit_agent.proto
```diff
@@ -21,154 +21,174 @@ option ruby_package = "LiveKit::Proto";
...
 
-    // the metadata can be updated multiple times
-    optional string metadata = 4;
+  // the metadata can be updated multiple times
+  // optional string metadata = 4 [deprecated=true];
     
-    // The load that the job currently has on the worker
-    float load = 5;
+  // The load that the job currently has on the worker
+  float load = 5;
```
paulwe, line 170: as convenient as it would be to have this does the agent client have any way to give us job load info? isn't most of the work done in a single process? @theomonnom ?

protobufs/livekit_agent.proto
```diff
@@ -21,154 +21,174 @@ option ruby_package = "LiveKit::Proto";
...
+  Room room = 3;
+  optional ParticipantInfo participant = 4;
+  string namespace = 5;
+  string metadata = 6;
+  JobStatus status = 7;
+  string error = 8;
+  float load = 9;
+  int64 started_at = 10;
+  int64 ended_at = 11;
+  int64 updated_at = 12;
```
paulwe, line 53: i'm not sure we want any of these on `Job` - this is the job description passed to the worker to initialize a job.

protobufs/livekit_agent.proto
```diff
@@ -21,154 +21,174 @@ option ruby_package = "LiveKit::Proto";
...
-message WorkerInfo {
-    string id = 1;
-    string namespace = 2;
-    string version = 3;
-    string name = 4;
-    JobType type = 5;
-    ParticipantPermission allowed_permissions = 6;
+service AgentService {
+  rpc StartAgentJob(StartAgentJobRequest) returns (Job);
+  rpc StopAgentJob(StopAgentJobRequest) returns (Job);
```
paulwe, line 26: when we stop a job how are we going to signal this to the agent? cut their participant session? i want to be sure i'm handling this the same in cloud as oss

protobufs/livekit_models.proto
dennwc: That's a part of https://github.com/livekit/protocol/pull/706 probably. It was confusing to see two `participant_identities` in data packets in SDK.

protobufs/livekit_room.proto
biglittlebigben: A "configuration service" would work too. I have no strong opinion either way. Does anybody else have any input? Either way, we'll need to sort out permissions.

protobufs/livekit_agent.proto
biglittlebigben: Egress runs each worker in a separate process and reports load for each ingress job separately, but that took a not insignificant amount of effort to implement and may be a too high burden to put on all agent implementations. I agree we should delete it unless @theomonnom has some further input on how this would be used?

protobufs/livekit_models.proto
davidzhao: I can see that.. would comments in the protobuf help to clear that up? Just looking at a `Transcription` object (without context on what is using the message), it seems to break some naming patterns to have `transcribed_` again

protobufs/livekit_agent.proto
biglittlebigben: This mixes Job Model and Execution status in the same message indeed. This is the pattern that is also used by Egress. Most consumers of the Job object need its current or latest state as well, which makes it convenient to have a message that makes it easy to send them together.

The IngressInfo message attempts to separate the Model and State somehow by packaging all the state related entries into a an IngressState message that is in turns included inside the Ingress message. The IngressState entry is null when the Ingress is used purely as model (which only happens at initialization).

An laternative here would be to flip the Job and JobState relationship compared to ingress:

```
message JobState {
   JobStatus status = 1;
   string error = 2;
   float load = 3;
   int64 started_at = 4;
   int64 ended_at = 5;
   int64 updated_at = 6;
  Job job = 7; 
}
```

protobufs/livekit_agent.proto
biglittlebigben: In a first implementation where there is no adversarial relationship between the room and the agent (as they belong to the same entity), I think adding a JobTermination ServerMessage may be enough. For better security and resilience, we should probably eventually enforce this by terminating the room/participant RTC connection to the agent indeed.

protobufs/livekit_agent.proto
```diff
@@ -21,154 +21,174 @@ option ruby_package = "LiveKit::Proto";
...
+
+  // Time the job was started
+  int64 started_at = 6;
+
+  // Time the job was ended
+  int64 ended_at = 7;
+
+  // Time the job state was last updated
+  int64 updated_at = 8; 
+
```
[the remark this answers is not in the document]
biglittlebigben: Ingress and Egress send these as having the worker generate them makes them more accurate and consistent (as they are all generated by the same worker server). However, we do not control the agent/worker server here so I agree that livekit-server should probably maintain the JobState entirely here. I'll change the protocol accordingly.

We use the updated_at field to order messages that may be delivered out of order by PSRPC for egress and ingress. Is there a different scheme we may be able to use here?

protobufs/livekit_agent.proto
paulwe: this makes the most sense to me because the job config is an immutable property of the running process but for the sake of consistency we could compromise and copy ingress so they're at least separate objects

protobufs/livekit_agent.proto
paulwe: that works too - we could set a timeout and kill the peer connection if it doesn't close on its own but giving the client an opportunity to close gracefully (ie after a keyframe) is a good idea.

protobufs/livekit_models.proto
dennwc: Maybe `source_` then?

protobufs/livekit_agent.proto
davidzhao: We have a `RemoveParticipant` API that would end up sending a `LEAVE` signal event to the user.. is that sufficient for this?

protobufs/livekit_models.proto
davidzhao: that sgtm!

protobufs/livekit_room.proto
davidzhao: I would propose not to have a service for this initially. OSS would configure these in the config yaml. Cloud would store it per project.

biglittlebigben: How would cloud allow defining these configurations?

protobufs/livekit_agent.proto
biglittlebigben: I think it's a bit of a roundabout way of doing it. I typically rather have explicit APIs. Particularly, if there is a `Start` API, I think a `Stop` is typically expected.

That being said, we should define the behavior if RemoveParticipant is called on the agent participant.

davidzhao: Totally agreed with an explicit API.. I'm more talking about "how to actually stop the agent".. you could use the same mechanisms as `RemoveParticipant`

protobufs/livekit_agent.proto
```diff
@@ -21,154 +21,178 @@ option ruby_package = "LiveKit::Proto";
...
-    string id = 1;
-    string name = 2;
-    string version = 3;
+message StartAgentJobRequest {
+  JobType type = 1;
+  string room_name = 2;
+  optional string participant_identity = 3;
+  string namespace = 4;
+  string metadata = 5;
+}
```
paulwe, line 36: in the current implementation publisher jobs apply to all participants once they begin publishing. this doesn't fit into this api. multiple job ids could result from a single invocation and more jobs could start after the initial call.

i think we've accidentally overloaded the `Job` concept - in the original design this represented an an invocation of the agent worker. what we're creating with this api call is a rule for dispatching jobs ie start a job for the room or start a job for publishers or for one publisher with the identity x.

davidzhao: maybe just `StartAgent` ?

biglittlebigben: I think we have 2 options:
- Change the Job definition to represent 1 invocation for either the whole room, or 1 participant
- Keep the current Job definition and remove the `participant_identity` argument in the `Start` message (and rename the rpc indeed). 

Is there a use case for starting an Agent Job for a single participant? At first glance, it seems so in context where there is a main speaker. Short of doing that, we may eventually need some kind of filter argument to tell a 3rd party agent what participants to process.

biglittlebigben: The current (before this PR) Job definition already has an optional `ParticipantInfo` field. How are agents currently expected to use it?

paulwe: we'll want the option to specify one or more publishers a job should start for. we don't want the user to have to specify the identities of every publisher in the room if they want the job to start automatically for publishers though. i think we need a new concept to describe what this api operates on - i don't have a great name for it but `WorkOrder` fits the job/worker metaphor.

the publisher's `ParticipantInfo` is stored in this field by the agents framework when dispatching a publisher type job.

paulwe:
can we add a field for the number of jobs active during the period the load sample is taken from to the `UpdateWorkerStatus` message.

what is the process for associating running jobs with a new worker connection when the websocket reconnects? should we expect a flood of `MigrateJobRequest` messages? maybe that message type should support migrating multiple jobs at once?

biglittlebigben:
> can we add a field for the number of jobs active during the period the load sample is taken from to the `UpdateWorkerStatus` message.
> 

Added. We have to define what the active Jon count is though: the count at capture time, the max since last capture, ...?

> what is the process for associating running jobs with a new worker connection when the websocket reconnects? should we expect a flood of `MigrateJobRequest` messages? maybe that message type should support migrating multiple jobs at once?

We have not implemented this anywhere yet AFAICT. I did change the message to use a list of ids though.

PR #719, opened 2024-05-21
Add support for additional SIP transports

dennwc:
Add support for additional SIP transports:
- TCP (in progress)
- TLS (for the future)


protobufs/livekit_sip.proto
```diff
@@ -90,6 +90,13 @@ message CreateSIPTrunkRequest {
   string metadata = 11;
 }
 
+enum SIPTransport {
+  SIP_AUTO = 0;
```
biglittlebigben, line 94: Shouls we have "TRANSPORT" as part of the constant names?

dennwc:
@biglittlebigben Done! Decided to keep a single `transport` on the Trunk level and added a helper for internal RPC (will be shared by OSS and Cloud).

biglittlebigben (approved):

davidzhao (approved):
lg!

PR #721, opened 2024-05-23
fix: rate intermittent tests

lherman-cs:
changes:
  * add retries to advance clock to unblock rate limiters
  * rewrite some tests to be more robust and simpler
  * add assertCount with noise to consider oversubscribed CI machines

### DIFF
--- a/utils/rate_test.go
+++ b/utils/rate_test.go
@@ -31,7 +31,6 @@ import (
 	"go.uber.org/atomic"
 
 	"github.com/benbjohnson/clock"
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


utils/rate_test.go
```diff
@@ -242,6 +258,19 @@ func (r *runnerImpl) assertCountAt(d time.Duration, count int) {
...
 }
 
+// assertCountAtWithNoise like assertCountAt but also considers possible noise in CI
+func (r *runnerImpl) assertCountAtWithNoise(d time.Duration, count int, noise int) {
+	r.wg.Add(1)
+	r.afterFunc(d, func() {
+		actual := int(r.count.Load())
+		boundMin, boundMax := count-noise, count+noise
+		withinRange := boundMin < actual && actual < boundMax
+		assert.True(r.t, withinRange, "expected count to be within [%d, %d], but got %d",
```
boks1971, line 268: Is this using `assert` because of the original package? We use `require` which has functions with check for things within range etc. natively.

boks1971 (approved):

utils/rate_test.go
lherman-cs: yes, the original tests were already written with assert. That's good to know, I'll go ahead and replace other `assert` with `require` as well

lherman-cs: replaced them

paulwe (approved):

PR #738, opened 2024-06-13
Split SIP Trunk configuration to inbound and outbound parts

dennwc:
Initial decisions made on the naming of the fields in `SIPTrunkInfo` leads to a lot of confusion for our users.

Most importantly, the `outbound_number` field is actually used for both inbound and outbound as the number associated with this LK SIP Trunk. Ideally it should be named `number`.

Unfortunately, we cannot rename it without breaking compatibility. This is mainly because our CLI uses `protojson` for decoding these fields. Changing the name will break old JSON request files. 

Some users also asked if it's possible to map multiple numbers to one (inbound) Trunk. Again, we cannot change the type of the field, but we could introduce a new one. At the same time, multiple numbers may be unnecessary for outbound Trunks, but they will still appear there because the structure is shared for inbound/outbound.

At the end, it's clear that there's no reason to share inbound and outbound configuration in one entity. Thus, this change splits Trunk into inbound and outbound configs.

Specifically:
- Introduce `SIPInboundTrunkInfo` and `SIPOutboundTrunkInfo`, add deprecation note for `SIPTrunkInfo`.
- Introduce new creation and list APIs for new types. Mark old ones as deprecated. Keep Delete, because it's agnostic.
- Add conversion functions between the new and old types.

We will keep DB structure mostly the same and convert objects in flight. One column will be added: `kind`. New Trunks will have it set to either inbound or outbound, so that new objects appear in one list only, while old ones appear in both. Eventually we could migrate to separate DB tables.


protobufs/livekit_sip.proto
```diff
@@ -132,16 +147,90 @@ message SIPTrunkInfo {
...
+
+  // Hostname or IP that SIP INVITE is sent too.
+  // Note that this is not a SIP URI and should not contain the 'sip:' protocol prefix.
+  string address = 4;
+
+  // SIP Transport used for outbound call.
+  SIPTransport transport = 5;
+
+  // A number used to make the calls.
+  string number = 6;
```
davidzhao, line 203: On the trunking provider side, it's still possible to associate multiple numbers to the SIP trunk. but I guess the point here is when making a call, a particular number will always be used?

protobufs/livekit_sip.proto
```diff
@@ -47,8 +47,13 @@ option ruby_package = "LiveKit::Proto";
 
 
 service SIP {
-  rpc CreateSIPTrunk(CreateSIPTrunkRequest) returns (SIPTrunkInfo);
-  rpc ListSIPTrunk(ListSIPTrunkRequest) returns (ListSIPTrunkResponse);
+  rpc CreateSIPTrunk(CreateSIPTrunkRequest) returns (SIPTrunkInfo) { option deprecated = true; };
+  rpc ListSIPTrunk(ListSIPTrunkRequest) returns (ListSIPTrunkResponse) { option deprecated = true; };
```
davidzhao, line 51: what should the behavior be for a deprecated trunk? should it continue to return all trunks, and indicating the type? or should it only list trunks are created with the old API?

davidzhao (approved):
lgtm! this seems cleaner and easier to understand.

protobufs/livekit_sip.proto
dennwc: Yes, it's possible to use multiple numbers for outbound on the provider's side.

Exactly, this is to specify which specific number to use. Having said that, I was also considering an array here. The implementation could pick a random number from an array before making a call.

Lets allow multiple outbound numbers then.

protobufs/livekit_sip.proto
dennwc: My plan was that the list will still return old trunks + all new ones (converted with the helpers). So the user will see correct metadata, but the config for them may be incomplete (e.g. only the first `number` visible).

protobufs/livekit_sip.proto
davidzhao: so maybe for the protocol, it'd be better to leave it as a repeated field? but in usage, we would let people know that only the first is used?

in the future, we can then add ways of choosing a trunk at run-time

davidzhao (approved):

protobufs/livekit_sip.proto
dennwc: Already updated to repeated field :+1: It will actually use the random one from the list if the Trunk is created with new API. And if not - it will be the same as before (one number only).

dennwc:
I'll keep it open for a bit longer - preparing PRs to other repos.

dennwc:
Works well, merging

PR #742, opened 2024-06-15
Get time inside lock.

boks1971:
There was a very small window (I think) where if entries are getting reset around the same time, the time ordering could get reversed in the queue. Not a big issue, but was chasing something else and noticed this window and thought of making the change.


paulwe:
this is an intentional tradeoff to avoid putting the syscall in the critical section. the accuracy is limited by the polling rate which is always slow - usually several seconds.

boks1971:
Ah got it. Closing this then.

PR #743, opened 2024-06-19
Add ErrorResponse

lukasIO:
skipping one id in `UpdateParticipantMetadata` in order not to interfere with #733. 
Will wait for #733 to be merged before merging this.


paulwe (approved):

protobufs/livekit_rtc.proto
```diff
@@ -416,3 +418,10 @@ message SubscriptionResponse {
   string track_sid = 1;
   SubscriptionError err = 2;
 }
+
+message ErrorResponse {
+  uint32 request_id = 1;
+  // http error code
+  uint32 code = 2;
```
boks1971, line 425: Does this need to be http error code? Can everything be mapped to this?

lukasIO: doesn't need to be! Just wanted to give users a familiar set of codes to work with. 
What would you suggest as an alternative?

boks1971: If it can be mapped, fine. Too often I have found that it ends up being an exercise is trying to decide between a couple of different options. If we can do enum, that will be great.

lukasIO: I see. Do you think we need anything beyond: a default/unspecified error, a `not_found` and a `not_allowed` error?

boks1971: Probably not. But things always have a way of expanding. We can see close reason/ disconnect reason. It has grown to a few. This could also grow. It may never, but just leaving the possibility open.

paulwe (approved):

protobufs/livekit_rtc.proto
lukasIO: updated the proto file to use a `Reason` enum instead of the http error code

davidzhao (approved):
lgtm

PR #756, opened 2024-07-10
Update to go/jose/v4

davidzhao:


boks1971 (approved):

paulwe (approved):

auth/verifier_test.go
```diff
@@ -15,19 +15,20 @@
...
 	"github.com/stretchr/testify/require"
 
 	"github.com/livekit/protocol/auth"
 )
 
 func TestVerifier(t *testing.T) {
 	apiKey := "APID3B67uxk4Nj2GKiRPibAZ9"
-	secret := "YHC-CUhbQhGeVCaYgn1BNA++"
-	accessToken := "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2MDg5MzAzMDgsImlzcyI6IkFQSUQzQjY3dXhrNE5qMkdLaVJQaWJBWjkiLCJuYmYiOjE2MDg5MjY3MDgsInJvb21fam9pbiI6dHJ1ZSwicm9vbV9zaWQiOiJteWlkIiwic3ViIjoiQVBJRDNCNjd1eGs0TmoyR0tpUlBpYkFaOSJ9.cmHEBq0MLyRqphmVLM2cLXg5ao5Sro7am8yXhcYKcwE"
+	secret := "YHC-CUhbQhGeVCaYgn1BNA++HC-CUhbQhHC"
```
davidzhao, line 30: jose v4 enforces hmac key length to be 32 bytes. our secrets have always been 43 bytes. (256 bit secret base62 encoded)

paulwe: the block size for sha256 is 32 bytes so our key was getting truncated silently in jose v3. if we start with a 32 byte random secret then base64 encode it and take the first 32 bytes of the 43 byte ascii string we're reducing the strength of the key.

davidzhao: are you sure about truncation behavior? IIRC keys longer than length is first hashed to 32 bytes.

davidzhao: https://datatracker.ietf.org/doc/html/rfc2104#section-3

paulwe: yeah, nvm, you're right. the key is hashed not truncated

PR #775, opened 2024-07-30
Partial documentation of RoomConfiguration parameters

biglittlebigben:


rektdeckard:
This looks great! Unfortunately, there is an [open issue](https://github.com/stephenh/ts-proto/issues/975) in `ts-proto` from 7 months ago that breaks compilation with the option needed to generate TS bindings for FieldOptions. But.. I was able to hack around it and produce the annotated bindings:

image

Have not done deep testing to make sure this doesn't break things elsewhere, but it's possible the fix in `ts-proto` is as simple as that.

davidzhao (approved):

biglittlebigben:
Are we ready to merge this? Once we do, we'll have to maintain these tags, so I'd like to be sure that we are planning to use them.

PR #776, opened 2024-08-01
Customizable retry parameters and improved logging

davidzhao:


webhook/url_notifier.go
```diff
@@ -141,3 +163,46 @@ func (n *URLNotifier) send(event *livekit.WebhookEvent) error {
...
+		if event.EgressInfo.Error != "" {
+			fields = append(fields, "error", event.EgressInfo.Error)
+		}
+	}
+	if event.IngressInfo != nil {
+		fields = append(fields,
+			"ingressID", event.IngressInfo.IngressId,
+		)
+		if event.IngressInfo.State != nil {
+			fields = append(fields, "status", event.IngressInfo.State.Status)
```
boks1971, line 201: guess ingress and egress cannot co-exist, so field collision of `status` and `error` cannot happen, but wondering if it makes sense to just call them `egressStatus`/`egressError`. Just a thought, but `status`/`error` works fine.

boks1971 (approved):

webhook/url_notifier.go
davidzhao: right, they are not able to coexist.. the goal here is to standardize our logging fields to make it easier for searches/etc.

boks1971: yep, makes it easier to search, good point.

PR #777, opened 2024-08-01
initial metrics protocol

theomonnom:
Motivation:
 - add the ability to send metrics to the client SDKs
     - e.g: LLM time to first token, SFU e2e latency, ...
 - the SFU will forward the metrics to analytics-recorder to be able to show the metrics on the Cloud dashboard (future work)

At the moment, only two metrics are supported: time series and annotations. Both can be synchronized with a track if `track_sid` specified.

The protocol is general-purpose. Developers using the agents framework can use the same API to send their own metrics to clients SDKs.
Only specific labels will be recorded on our servers for the Dashboard. (e.g `lk-speech-probability`, `lk-llm-ttft`)

The MetricsBatch message is used to send the metrics in batches. I'm still undecided on what the appropriate interval should be.

Time series values are currently `float` — Do you think it is a right assumption to make?

At the moment, only audio use cases are the primary focus. I think `AnnotationMetric` could have a bounding box field in the future for videos for e.g

The protocol is currently minimalistic. Happy to get suggestions :)


protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,55 @@
...
+    IntervalEncoding interval_encoding = 6;
+  }
+
+  message DeltaEncoding {
+    // protobuf encodes int32 as varint, delta times are usually small
+    repeated int32 delta_times = 1; // Delta time in milliseconds from the previous timestamp
+    repeated float values = 2;
+  }
+
+  message IntervalEncoding {
```
keepingitneil, line 35: Do we need this? Seems like a special case of delta encoding? Is it just meant to be a size optimization?

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,55 @@
...
+
+message TimeSeriesMetric {
+  // Metric name, e.g., "speech_probability"
+  string label = 1;
+  // optional participant identity
+  string participant_identity = 2;
+  // optional track sid
+  string track_sid = 3;
+  // if track_sid is set, timestamp represents the RTP timestamp
+  int64 start_timestamp = 4;
```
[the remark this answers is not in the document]
bcherry: is the implication that, as-is, this is wall clock if track_sid isn't set? if so - i agree with Neil it would be good to always have the wall clock value

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,55 @@
...
+  // Metric name e.g "assistant_interrupted", "speaking"
+  string label = 1;
+   // optional participant identity
+  string participant_identity = 2;
+  // optional track sid
+  string track_sid = 3;
+  // if track_sid is set, timestamp represents the RTP timestamp
+  int64 timestamp = 4;
+   // optional duration of an event
+  optional float duration = 5;
```
bcherry, line 51: wouldn't you want this to be in the same basic data type as `timestamp`?

protobufs/livekit_models.proto
```diff
@@ -226,8 +228,9 @@ message DataPacket {
     ActiveSpeakerUpdate speaker = 3 [deprecated=true];
     SipDTMF sip_dtmf = 6;
     Transcription transcription = 7;
+    MetricsBatch metrics = 8;
```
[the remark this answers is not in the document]
lukasIO: Curious to know what the ideas are behind having this sent via DC instead of WS ?

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,55 @@
...
+    repeated float values = 2;
+  }
+
+  message IntervalEncoding {
+    uint32 interval = 1; // Interval in milliseconds between data points
+    repeated float values = 2;
+  }
+}
+
+message AnnotationMetric {
```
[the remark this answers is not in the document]
nbsp: if we want to go with prometheus nomenclature, this should be called a `Sample`

protobufs/livekit_metrics.proto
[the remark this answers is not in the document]
nbsp: agreed -- have a timestamp, and optionally a (`track_sid` + `track_timestamp`)

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,55 @@
...
+
+
+message MetricsBatch {
+  repeated TimeSeriesMetric time_series = 1;
+  repeated AnnotationMetric annotations = 2;
+}
+
+message TimeSeriesMetric {
+  // Metric name, e.g., "speech_probability"
+  string label = 1;
```
lukasIO, line 16: > Only specific labels will be recorded on our servers for the Dashboard. (e.g lk-speech-probability, lk-llm-ttft)

Do we want to define these somewhere ?
I understand the sentiment that new labels shouldn't require protocol updates, but for stuff that we record, it might be nice for maintainability across different implementations.

protobufs/livekit_metrics.proto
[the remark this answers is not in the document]
paulwe: samples in prometheus are data points. it doesn't have good support for event data.

protobufs/livekit_metrics.proto
paulwe: we should capture start/end time. timestam/duration is ambiguous

protobufs/livekit_metrics.proto
paulwe: if we're capturing data with high enough granularity that this sort of micro optimization makes sense processing and storing it will be prohibitively expensive

protobufs/livekit_models.proto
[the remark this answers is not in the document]
theomonnom: I don't have a strong preference, only important thing is that the metrics should be able to be transmitted across relays.

protobufs/livekit_metrics.proto
theomonnom: I just removed the size implementation, as we will most likely always sample our real-time data at a maximum of 2 samples per second. (e.g speaking probability)

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,55 @@
+syntax = "proto3";
+
+package livekit;
+option go_package = "github.com/livekit/protocol/livekit";
+option csharp_namespace = "LiveKit.Proto";
+option ruby_package = "LiveKit::Proto";
+
+
+message MetricsBatch {
```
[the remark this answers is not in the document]
theomonnom: Are you suggesting we use Prometheus nomenclature, or are you considering setting up a Prometheus server for our metrics? I initially discussed with DZ the idea of setting up a Prometheus server for customer access. However, Prometheus isn't ideal for handling a lot of sessions, especially if you want to view metrics for a specific session, as it is better suited for aggregated data.

protobufs/livekit_metrics.proto
theomonnom: It's hard to know what are the labels we will record today (there is a lot..)
I was thinking about doing something similar to how we handle data messages topics (for e.g `lk-chat-topic`).

We could create an enum holding all our internal/records labels. On downside is that protobuf doesn't support string enums, so in the future when we're going to export data to our customers. We will have to provide two different ways of using our labels? (oneof str/our_enum)

protobufs/livekit_metrics.proto
theomonnom: Just added end_timestamp

protobufs/livekit_metrics.proto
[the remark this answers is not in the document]
theomonnom: No strong opinion here

protobufs/livekit_metrics.proto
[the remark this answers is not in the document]
theomonnom: Agree that it may be confusing for a same field to have 2 different clocks.
Looking at the current code, if we have to add the wall clock, we would need to duplicate a bunch of fields (start_timestamp, end_timestamp, timestamp of values)

Do you think the wall clock is necessary when we already know the start_time of a room, or the published time of a track for e.g? Also wondering about the use cases on when you really need the wall clock of your annotated data?

theomonnom:
Hey guys, to add to the PR description. This protocol is also going to be used by our Client SDKs (both receiving & sending).
The goal would be for the client to forward some webrtc stats (jitter buffer size, stats wrt echo cancellation, etc ...)

protobufs/livekit_models.proto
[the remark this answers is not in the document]
lukasIO: I was thinking primarily about the sending use case, but if receiving metrics is also planned for clients (thanks for the additional context) then DC makes sense to me

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,79 @@
...
+  SFU_MESH_LATENCY = 11;           // Mesh latency in SFU
+  SFU_SUBSCRIBE_LATENCY = 12;      // Subscribe latency in SFU
+  CLIENT_JITTER_BUFFER_DELAY = 13; // Jitter buffer delay on the client
+  CLIENT_PLAYOUT_DELAY = 14;       // Total delay on client-side from receiving a packet to playback
+  CLIENT_RESOLUTION_LIMIT = 15;    // Resolution limit for video publishing
+  CLIENT_FRAME_FROZEN = 16;        // Number of frame frozen events
+};
+
+message MetricsBatch {
+  map<int32, string> label_dict = 1; // reserve the first 4096 labels for predefined labels (`MetricLabel`)
```
paulwe, line 41: this can be a `repeated string` - we don't need explicit keys because we know the map keys are contiguous and begin with the first unreserved value

protobufs/livekit_metrics.proto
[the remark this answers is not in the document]
paulwe: are these annotations in the ml sense? like supplemental information to contextualize raw data?

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,79 @@
...
+  repeated AnnotationMetric annotations = 3;
+}
+
+message TimeSeriesMetric {
+  // Metric name, e.g., "speech_probability"
+  int32 label = 1;
+  // optional participant identity
+  string participant_identity = 2;
+  // optional track sid
+  string track_sid = 3;
```
paulwe, line 52: if these are repeated in multiple metrics we may want to put them in the dictionary

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,79 @@
...
+  AGENTS_STT_OUTPUT_TOKENS = 7;    // Number of tokens produced by STT
+  AGENTS_TTS_INPUT_TOKENS = 8;     // Input tokens for TTS
+  AGENTS_TTS_OUTPUT_DURATION = 9;  // Output duration for TTS
+  SFU_PUBLISH_LATENCY = 10;        // Publish latency in SFU
+  SFU_MESH_LATENCY = 11;           // Mesh latency in SFU
+  SFU_SUBSCRIBE_LATENCY = 12;      // Subscribe latency in SFU
+  CLIENT_JITTER_BUFFER_DELAY = 13; // Jitter buffer delay on the client
+  CLIENT_PLAYOUT_DELAY = 14;       // Total delay on client-side from receiving a packet to playback
+  CLIENT_RESOLUTION_LIMIT = 15;    // Resolution limit for video publishing
+  CLIENT_FRAME_FROZEN = 16;        // Number of frame frozen events
```
paulwe, line 37: should we have an enum entry for the end of the reserved range?

paulwe:
could we generate a representative sample of metrics we might send to help ground this?

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,79 @@
...
+
+message MetricsBatch {
+  map<int32, string> label_dict = 1; // reserve the first 4096 labels for predefined labels (`MetricLabel`)
+  repeated TimeSeriesMetric time_series = 2;
+  repeated AnnotationMetric annotations = 3;
+}
+
+message TimeSeriesMetric {
+  // Metric name, e.g., "speech_probability"
+  int32 label = 1;
```
boks1971, line 48: With the dictionary, this would be used only for ones that are not reserved, right? Maybe update comment to reflect that?

Also, should there be a field for `MetricLabel` enum in this proto message?

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,79 @@
...
+}
+
+message MetricSample {
+  int64 timestamp = 1;
+  float value = 2;
+}
+
+message AnnotationMetric {
+  // Metric name e.g "assistant_interrupted", "speaking"
+  int32 label = 1;
```
boks1971, line 68: Can we do the same dictionary technique for annotations too?

boks1971:
> Time series values are currently float — Do you think it is a right assumption to make?

I think that should be fine. Can't think of a non-numeric metric. So, `float` should encompass all.

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,79 @@
...
+  // Metric name, e.g., "speech_probability"
+  int32 label = 1;
+  // optional participant identity
+  string participant_identity = 2;
+  // optional track sid
+  string track_sid = 3;
+  // if track_sid is set, timestamp represents the RTP timestamp
+  int64 start_timestamp = 4; // samples
+  // if track_sid is set, timestamp represents the RTP timestamp
+  int64 end_timestamp = 5;
```
boks1971, line 56: Is this dual purpose? Is this RTP clock when track SID is present and wall clock if not?

protobufs/livekit_metrics.proto
boks1971: Updating my comment, this is an int. So that is the dictionary index I guess. Would be good to update the comment to explain that.

Should we add a string field for user generated keys?

theomonnom:
I just updated the protocol to reflect the feedbacks. 
For clarity, here are some examples on how this protocol could be used to send some data.


**Agent sending the time it took to get the first token of an LLM**
```python
event = proto.EventMetric(
    label=4096, # see MetricLabel.AGENTS_LLM_TTFT (4096 + 0)
    participant_identity=0,
    start_timestamp=time.now()
)

batch = proto.MetricsBatch(
    str_data=["my_agent_identity"],
    events=[event],
)

```


**The client sending the jitter buffer size**
timeseries = proto.TimeSeriesMetrics(
    label=4099, # see MetricLabel.CLIENT_JITTER_BUFFER_SIZE_MS (the label doesn't exist yet, this is just for the e.g)
    participant_identity=0,
    start_timestamp=0,
    end_timestamp=10.0,
    samples=[
         proto.MetricSample(0, 100),
         proto.MetricSample(0, 100),
         proto.MetricSample(0, 100),
         proto.MetricSample(0, 100),
         proto.MetricSample(0, 100),
    ]
)

batch = proto.MetricsBatch(
    str_data=["my_local_participant_identity"],
    time_series=[timeseries],
)

```

File: protobufs/livekit_metrics.proto [hunk 14]
```diff
@@ -0,0 +1,62 @@
...
+  AGENTS_TTS_TTFB = 2; // time to first byte
+}
+
+message MetricsBatch {
+  // To avoid repeating string values, we store them in a separate list and reference them by index
+  // This is useful for storing participant identities, track names, etc.
+  // There is also a predefined list of labels that can be used to reference common metrics starting at 4096.
+  // E.g: for the llm TTFT, the int value would be 4096 + 0 (MetricLabel.AGENTS_LLM_TTFT)
+  // Those predefined labels are not stored in the string list
+  repeated string str_data = 1;
```
paulwe, line 33: the low values should be reserved for predefined values since we aren't likely to define more than maybe 100 of them. in large aggregate batches there could be thousands of track/participant ids.

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,62 @@
...
+
+  A metric can be linked to a specific track by setting `track_sid`.
+*/
+
+
+// index starting at 4096 are for predefined labels (`MetricLabel`)
+enum MetricLabel {
+  AGENTS_LLM_TTFT = 0; // time to first token from LLM
+  AGENTS_STT_TTFT = 1; // time to final transcription
+  AGENTS_TTS_TTFB = 2; // time to first byte
```
paulwe, line 24: we should define `METRICS_LABEL_MAX_VALUE = 4096` or something like that. if we have it in the codegen it's available everywhere and we don't have to keep repeating it in comments.

protobufs/livekit_metrics.proto
```diff
@@ -0,0 +1,62 @@
...
+  repeated TimeSeriesMetric time_series = 2;
+  repeated EventMetric events = 3;
+}
+
+message TimeSeriesMetric {
+  // Metric name e.g "speech_probablity". The string value is not directly stored in the message, but referenced by index
+  // in the `str_data` field of `MetricsBatch`
+  int32 label = 1;
+  int32 participant_identity = 2;
+  optional int32 track_sid = 3;
```
paulwe, line 43: we could use 0 for nil. also these should be uints - we don't need to encode the sign bit

paulwe (approved):

protobufs/livekit_metrics.proto
boks1971: Agreed. Let us reserve smaller numbers for pre-defined ones.

PR #778, opened 2024-08-01
Added ROOM_CLOSED as a DisconnectReason

davidzhao:
When all independent participants leave the room, the room would close and disconnect all dependent participants. They should receive the correct reason. Currently we are sending `UNKNOWN_REASON`


lukasIO (approved):

theomonnom (approved):
ah that's nice, next step is to add it to the ffi for rust

paulwe (approved):

PR #785, opened 2024-08-12
Add dedicated chat message definition

lukasIO:
supersedes #753. 

A protocol level definition for chat is needed in order to get rid of application specific sub-protocols that we currently use for components.

One downside of this approach (having it live on the same level as Transcriptions) in contrast to #753 is that delivery is not explicitly guaranteed when sent via DC.


bcherry:
With this approach it also seems like it wouldn't be possible to have a `ChatMessage` without at least one published `Track`, right? So if you wanted a chat-only participant, you'd still need to publish an empty audio or video track? That seems a little limiting to me but maybe I'm missing something.

The other thing I'm wondering about is how you would implement a mixed type-and-speak app. Would ChatMessage just be for the typed ones, and you'd use Transcription for the spoken ones and the app developer would merge them? If you'd want both to go under ChatMessage, in the Voice Assistant framework, then we'd also want support for progressively-rendered ChatMessage which would require a similar `final` param as Transcription.

lukasIO:
> it wouldn't be possible to have a ChatMessage without at least one published Track

this would be unrelated to media tracks. It needs a data channel to be established, but that happens regardless of any published media.

> support for progressively-rendered ChatMessage

not sure whether this is really necessary. If it was, we could use the `edit` function of chat messages to update them progressively, too.

bcherry:
> > it wouldn't be possible to have a ChatMessage without at least one published Track
> 
> this would be unrelated to media tracks. It needs a data channel to be established, but that happens regardless of any published media.

I'd assumed that since there's no participant field on it but i guess if that's being handled automatically it makes sense. client events would get ChatMessage + Participant notifications right?

lukasIO:
yeah, they would! 
JFYI, the envelope for this is `DataPacket`, which has the participant identity.

protobufs/livekit_models.proto
```diff
@@ -286,6 +287,14 @@ message TranscriptionSegment {
   string language = 6;
 }
 
+message ChatMessage {
+  string id = 1; // populated by server
+  int64 timestamp = 2; // populated by  server
+  int64 edit_timestamp = 3; // populated by server
+  string message = 4;
+  bool deleted = 5; // true to remove message
```
boks1971, line 295: Shouldn't the sending client be populating these fields? What would the sending client message look like? Just the `message`?

And server would populate the rest before forwarding? But, that will cause issues for edits. How would the server know if a message has been edited if it is the one generating the unique message id.

boks1971:
A couple of things

> One downside of this approach (having it live on the same level as Transcriptions) in contrast to https://github.com/livekit/protocol/pull/753 is that delivery is not explicitly guaranteed when sent via DC.

Reliable DC should make this work for most cases reliably.

If we want to send this over signalling channel, would we send the `DataPacket` in a `SignalResponse`? I am thinking of cases where data channel is optional. I think it is fine to have the requirement that if an app is using `Transcriptions` or `ChatMessage`, they will need to have data channels. But, just wanted to do the thought exercise of ensuring it can work over signalling channel also. We already have the example of speaker updates moving from data channels to signalling.

protobufs/livekit_models.proto
lukasIO: ah, not sure why the `id` is also generated by the server. 
Intention was only for the `timestamps` to be populated by the server. 
`id, message and deleted` would be set by the client. 
Does that make more sense?

boks1971: Got it.

I am guessing `deleted` also should be populated by sending client?

A few more questions
1. How many messages (ids) should server hold on to? Feels like sending client should populate the time stamps and server will just normalise it to server clock maybe?
2. If server is holding on to messages, should it replay them on a new participant join?

boks1971: > I am guessing deleted also should be populated by sending client?

Ignore that, thought that also said it was populated by server, but no, the comment does not say that.

davidzhao:
looks good. 

question: if for the agent use case, we are sending chat transcripts as chat, instead of transcriptions. how would we handle the case where Agent is sending a message indicating something the user has said? I.e. the DataPacket envelope would indicate it's coming from the agent. (but we really would like the UI to indicate it's the user)

lukasIO:
> the DataPacket envelope would indicate it's coming from the agent.

current thinking (after last meeting) is that explicit permissions to override the sender identity would be conceivable for this to work

davidzhao:
> current thinking (after last meeting) is that explicit permissions to override the sender identity would be conceivable for this to work

That makes sense. Assuming the permission is there, how would we go about sending such a message with this protocol?

lukasIO:
we could pre-populate the `participant_identity` field on the `DataPacket` envelope?

davidzhao:
> we could pre-populate the `participant_identity` field on the `DataPacket` envelope?

ah that could work. that field is currently being set by the SFU, but with the new permission, we could trust that participant to indicate the data is attributed to someone else.

lukasIO:
yeah, that would be the idea!

davidzhao (approved):

PR #788, opened 2024-08-13
Add `DisconnectReason` to `ParticipantInfo`.

boks1971:
@lukasIO Last time, I did not create a changeset. Is there a specific step to create that? Do I need to run `pnpm changeset` in the repo?


lukasIO:
> Do I need to run pnpm changeset in the repo?

the command to add a changeset would be
```
pnpm changeset add
```

the CLI will guide you through the steps. For this repo it's best to select "all packages" to be affected by protocol changes.

protobufs/livekit_models.proto
```diff
@@ -131,6 +131,7 @@ message ParticipantInfo {
   bool is_publisher = 13;
   Kind kind = 14;
   map<string, string> attributes = 15;
+  DisconnectReason disconnect_reason = 16;
```
lukasIO, line 134: any chance of potentially including insight in to whether the participant is attempting a reconnect ? 
I think users have been asking about this in the past.

boks1971: That is a transient state. Can think about it. This reason would really be valid when `state = DISCONNECTED`. Reconnect indication would be orthogonal I think. What's your thinking on how that should work?

.changeset/great-cougars-buy.md
```diff
@@ -0,0 +1,6 @@
+---
+"github.com/livekit/protocol": major
```
lukasIO, line 2: oh, this should be `patch` instead of major. The CLI is a bit confusing about this. but manually edititing it here will also work

.changeset/great-cougars-buy.md
```diff
@@ -0,0 +1,6 @@
...
+"@livekit/protocol": major
```
lukasIO, line 3: same as above

boks1971: Do I just run the command again?

boks1971: Oops, missed the earlier comment. Will manually edit.

protobufs/livekit_models.proto
boks1971: Thinking more about the reconnecting indication.

`reconnect = 1` if successful should be (mostly) transparent. But, that is the case where server clearly knows there is an attempt to resume.

A full reconnect looks like a new connection to the server. If the participant exists in the room and a full reconnect happens, server could indicate that the participant is attempting another connection (but that would not work in cloud if connecting to different node and that node does not have the old participant's remote participant, would require sending a message to check, but that would impact every join).

Would be good to see what users want and figure out how to do that.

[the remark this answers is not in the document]
boks1971: Sure. If that is `Participant` specific, I think it should be okay. Thoughts on that @davidzhao .

Currently, internally server has these reasons for participant close - https://github.com/livekit/livekit/blob/5d554867102d8b3bee6e8865a85cae5513c19c62/pkg/rtc/types/interfaces.go#L84 and it gets mapped to `DisconnectReason` here - https://github.com/livekit/livekit/blob/5d554867102d8b3bee6e8865a85cae5513c19c62/pkg/rtc/types/interfaces.go#L169.

We can make more participant close reasons and map them to reasons like you mention above.

We do have participant kind now. Maybe, each kind has more specific reasons.

Question: I do not know enough about how SIP participant works. How would the SIP participant be conveyed the reason for its close? Actually, I guess I should first learn how the SIP participant gets closed now.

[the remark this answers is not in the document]
dennwc: Currently it just disconnects from the room as usual (or is remove from the room by API). So I'll need to update the SDK to expose the disconnect reason (or at least some of them).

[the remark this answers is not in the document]
boks1971: Got it. Client leaving currently sends this https://github.com/livekit/protocol/blob/a77139eae2660bc3d88e4440a174d8287b07bcf9/protobufs/livekit_rtc.proto#L39 and server just marks it as client initiated leave. If clients are going to send a reason, we can update server to parse the reason and mark the proper reason (would be mapping from `DisconnectReason` -> `ParticipantCloseReason` and back to `DisconnectReason`, but that is okay 🙂 )

lukasIO (approved):

boks1971:
@dennwc Going to merge this for now. We can add more reasons in a subsequent PR.

@davidzhao Please let us know if you have further thoughts on SIP disconnect reason Denys mentioned and if you have ideas on how to do that? Enhance `DisconnectReason` or a different method. That would involve client sending the reason with `LeaveRequest`, server interpreting it and plumbing that through.

protobufs/livekit_models.proto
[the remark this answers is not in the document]
davidzhao: I think we can add these as new disconnect reasons for SIP.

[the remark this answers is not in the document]
boks1971: Thank you @davidzhao .

@dennwc Please let me know when you add these to `LeaveRequest`. We can then change the server to do the mapping.  Maybe call the reasons with a `SIP_` prefix for `UNAVAILABLE` and `REJECTED`. I do not particularly like prefixing with specific like that, but the un-prefixed version is very generic. So, that could be confusing.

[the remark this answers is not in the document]
davidzhao: yeah hangup makes sense.. but for the rest, how about keeping it the original SIP reasons instead of mapping? this way we wouldn't need to prefix either

No response -> NO_RESPONSE
Busy -> USER_BUSY (not sure about the USER_, just wanted to disambiguate from server busy)

PR #791, opened 2024-08-23
Add EnabledCodecs in JoinReponse

cnderrauber:
Client can use `EnabledCodecs` to select fallback video codec in client side instead of server side so it can negotiate sdp with send `AddTrackRequest` in parallel to speed up the track publication.


boks1971 (approved):

paulwe (approved):

protobufs/livekit_rtc.proto
```diff
@@ -443,3 +444,7 @@ message RequestResponse {
 message TrackSubscribed {
   string track_sid = 1;
 }
+
+message EnabledCodec {
+  repeated Codec videos = 1;
```
paulwe, line 449: does pluralizing this make sense? is this for different videos or is it different video codecs?

cnderrauber: it is all enabled codecs at server side, the plan was using a standalone structure then audio codecs can be added to o but the mimetype already contains kind prefix so removed here  https://github.com/livekit/protocol/pull/791/commits/0e53a6c4d6665706b39e0e1bc3c665c7dd01a55f

lukasIO (approved):

PR #794, opened 2024-08-25
Add published_at field to keep track of published timestamp

davidzhao:
This is useful to determine total track duration


protobufs/livekit_models.proto
```diff
@@ -193,6 +193,8 @@ message TrackInfo {
   string stream = 17;
   TimedVersion version = 18;
   repeated AudioTrackFeature audio_features = 19;
+  // server time when the track was published, in milliseconds
+  int64 published_at = 20;
```
boks1971, line 197: Can we use the `TimedVersion` ☝️

davidzhao: that would certainly be easier.. let me take try that!

davidzhao: that does it, thanks Raja!

PR #795, opened 2024-08-26
Add grant for canSubscribeMetrics.

boks1971:


auth/grants.go
```diff
@@ -78,6 +78,8 @@ type ClaimGrants struct {
 	Metadata string `json:"metadata,omitempty"`
 	// Key/value attributes to attach to the participant
 	Attributes map[string]string `json:"attributes,omitempty"`
+	// if a participant can subscribe to metrics
+	CanSubscribeMetrics *bool `json:"canSubscribeMetrics,omitempty"`
```
davidzhao, line 82: can we put this into `Video` grant? this is where the other permission bits are stored. not named the best, but seems more consistent given where things are today.

boks1971: okay, I put it here because of the naming.

boks1971: Moved it in this [commit](https://github.com/livekit/protocol/pull/795/commits/1add410bf995f5e8e69926accf3e517fd187f908). A few more things
- Grouped struct with its methods (makes it easier on my brain :-) )
- It defaults to `false`, so not adding to things like `BuildSIPToken`, `BuildAgentToken`.

davidzhao (approved):
lgtm

PR #797, opened 2024-08-29
Add SDK values for node and unity-web

bcherry:
We'd like to be able to distinguish these SDKs from their underlying implementations (Rust and JS).

I'm not sure if there is more I'm supposed to do for testing when submitting a protocol PR so lmk.


davidzhao (approved):

lukasIO (approved):

theomonnom (approved):

bcherry:
@lukasIO thank you!

PR #798, opened 2024-08-31
Add callee dispatch rule type.

dennwc:
Add a new Dispatch Rule type that selects a room name based on a callee.

Creating an empty Callee dispatch rule allow mapping `<room>@xyz.sip.livekit.cloud` to the `<room>`. Similarly to Individual rule type (which we should have called "Caller") it also allows setting a prefix for the room name.


davidzhao (approved):
nice!

PR #804, opened 2024-09-05
Add other_sdks field to ClientInfo

bcherry:
https://linear.app/livekit/issue/CLT-605/spec-out-component-usage-analytics
https://linear.app/livekit/issue/CLT-642/add-other-sdks-to-clientinfo


protobufs/livekit_models.proto
```diff
@@ -359,6 +359,9 @@ message ClientInfo {
   string address = 9;
   // wifi, wired, cellular, vpn, empty if not known
   string network = 10;
+  // comma separated list of additional LiveKit SDKs in use of this client, with versions
+  // e.g. "components-js:1.2.3,track-processors-js:1.2.3"
```
lukasIO, line 363: what names should the additional SDKs inherit? 
Currently they're repository names, wondering if package names is more applicable? Not sure though.

lukasIO: Another thing that comes to my mind with `track-processors-js` being mentioned: At join time (when the clientInfo is being sent) we don't necessarily know yet whether or not a client is using such an additional library as it might only be set and/or loaded at a later point in time.

bcherry: i don't have an opinion on the names as long as they're consistent for SDKs that could appear in this list from multiple sources (e.g. the rust SDK being called `rust` when referenced from node, python, and unity)

bcherry: good thought on track processors - not sure there's an easy solution but maybe some way through webpack to detect the existence of the package?

lukasIO: >  as long as they're consistent for SDKs

was thinking it might be good to define them in an enum then.
we might want to change the type to a map with the SDK as a key and the version as a value? 

But looks like even then we cannot use enum values as keys in protobuf :/

lukasIO: > way through webpack to detect the existence of the package?

this would probably require a custom webpack plugin and additionally users would have to include that in their setup. Another way would be to allow client info to be updated during an ongoing session

bcherry: maybe a little naive but I think this field is a pretty in-the-weeds bit of analytics info anyways so maybe we can try the simple thing and if it gets a little messy we can sort it out in queries. it's not gonna be a high-traffic query field IMO

lukasIO: makes sense, the enum idea was just to make the naming consistency a given. 
Converting it to a map<string, string> would potentially make it easier to implement by skipping manual parsing on the field. 
If we're not opting for a map another option would be `repeated string` with each array value representing a sdk version pair. 
Not a strong opinion either way, also happy to keep it as is, just wanted to think about alternative options.

lukasIO (approved):

protobufs/livekit_models.proto
bcherry: i'm just anxious about making things complicated to log - otherwise i agree with you. but i'm not familiar enough so just generally cautious. here's the spot I want it to go:

https://github.com/livekit/livekit/blob/b678ccdd66c4dfa3f256ad0da2b5f340ae5ea176/pkg/service/roommanager.go#L407

I don't know if this supports nested types? I guess it probably does - what do you think @shishirng?

PR #807, opened 2024-09-10
Make messages for RTCPSenderReportState.

boks1971:
To be used for sharing state between nodes on migration.


protobufs/livekit_models.proto
```diff
@@ -486,9 +486,26 @@ message RTPStats {
   google.protobuf.Timestamp last_layer_lock_pli = 36;
 
   RTPDrift packet_drift = 44;
-  RTPDrift report_drift = 45;
+  RTPDrift ntp_report_drift = 45;
```
boks1971, line 489: just a rename.

paulwe (approved):

PR #809, opened 2024-09-18
add utilities for applying options and copying protobufs

paulwe:


utils/options/options.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package options
+
+func Apply[T any, F ~func(T)](o T, opts []F) T {
```
boks1971, line 17: Sorry Paul, I already asked you this once, but I forgot the answer :-(. What does the `~` mean there?

utils/proto.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import "google.golang.org/protobuf/proto"
+
+func CloneProto[T proto.Message](m T) T {
+	return proto.Clone(m).(T)
```
boks1971, line 20: nice!!!

boks1971 (approved):

davidzhao (approved):

utils/options/options.go
paulwe: `~` means the type constraint permits any type with the same underlying type. so for example
```go
type Foo func()

var f0 = func() {}
var f1 Foo = func() {}

func a[T ~func()](f T) {}
func b[T func()](f T)  {}

a(f0) // ok
b(f0) // ok
a(f1) // ok
b(f1) // not ok
```

boks1971: ah, got it. Thank you Paul. Let me commit this to my poor memory and hope it survives till the next time I see this pattern :-)

paulwe: this is needed here because we have functional options like
```go
type FooOption = func(x *FooOptions)

// several implementations of FooOption

func SomeFoo(opts ...FooOption) {
    _ := options.Apply(&FooOptions{}, opts)
    // without ~ this wouldn't work because []FooOption is not []func(*FooOptions)
}
```

PR #814, opened 2024-09-19
RPC types

bcherry:
We’d like to implement an “RPC”-style feature between participants primarily to support agent tool call and other action forwarding to participants, with robust guarantees and error states around delivery. 

Full proposal here: https://www.notion.so/livekit/RPC-Proposal-435567131ce64077aa4466fafc2fb407?pvs=4


protobufs/livekit_models.proto
```diff
@@ -301,6 +304,31 @@ message ChatMessage {
...
+}
+
+message RpcAck {
+  string message_id = 1;
+}
+
+message RpcResponse {
+  string message_id = 1; // ID of the request message
+  string data = 2;
+  int32 error_code = 3; // Can be a value from ErrorCode enum or a custom value
```
lukasIO, line 320:
```suggestion
  uint32 error_code = 3; // Can be a value from ErrorCode enum or a custom value
```

bcherry: dropped in favor of a single string

protobufs/livekit_models.proto
```diff
@@ -302,6 +305,26 @@ message ChatMessage {
   bool generated = 6; // true if the chat message has been generated by an agent from a participant's audio transcription
 }
 
+message RpcRequest {
+  string id = 1;
+  string method = 2;
+  string payload = 3;
+  uint32 response_timeout_ms = 4;
+  uint32 version = 5;
```
boks1971, line 313: How is the version going to be used? Is it SDK level or application level?

If application level, feels like a version in response would be good.

boks1971 (approved):

protobufs/livekit_models.proto
bcherry: SDK level - just thinking ahead that we might change the protocol enough down the road (e.g. adding streaming) that you'd want an outdated client to reject new requests and/or an updated client to reject old ones. I figure we'll add `if version != 1 { error }` to the recipient side right now just in case? I'm not sure it's really needed but it might be good to have.  In that case we might have version 2 send both version 1 & 2 calls, and new recipients would only process the 1s. i'm not sure - probably needs more thought. But I'm anxious about making a mistake in the protocol that we only discover a little down the road and decide we need to update it in a way that's not backwards compatible.

protobufs/livekit_models.proto
```diff
@@ -302,6 +305,26 @@ message ChatMessage {
   bool generated = 6; // true if the chat message has been generated by an agent from a participant's audio transcription
 }
 
+message RpcRequest {
+  string id = 1;
```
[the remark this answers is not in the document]
bcherry: ah this alerted me to a mistake - the others are meant to be `request_id` which should then make it obvious that this one, called request, would just have an `id`? updated

protobufs/livekit_models.proto
boks1971: Got it. We have used a high level protocol version to check for compatibility, but a message level would be good too. I am guessing apps will encode their own versions in the payload if needed and handle backwards compatiblity.

But, the challenge in this case becomes how far do you go. If we are version 10 (hopefully never), would requester send all 10 versions?

With protobuf, new fields can be added and older clients would not be able to read them. Would the protocol change in a way where a v2 would look completely different or if you send the added fields in v2, v1 receiver would read it as v1 and send something back which is completely wrong?

Yeah, something to think about.
It is possible to an `optional` version field later too, but it becomes a bit awkward on receiver side. It would be something like
```
    if versionPresent {
          version = versionField
          if version != myVersion {
               return error
          }
    } else {
        return error
    }
```
So, that extra check to see if version is present or not would be needed if optional version is added later.

protobufs/livekit_models.proto
```diff
@@ -302,6 +305,26 @@ message ChatMessage {
...
+
+message RpcAck {
+  string request_id = 1;
+}
+
+message RpcResponse {
+  string request_id = 1;
+  oneof value {
+    string payload = 2;
+    string error = 3;
```
lukasIO, line 324: in a previous iteration you had `Error` as consisting of both an error_message and an error_code (or similar). What's the incentive of dropping the error_code? I thought that was a neat addition.

lukasIO (approved):

protobufs/livekit_models.proto
bcherry: The ergonomics were messy in TypeScript, although maybe I was just doing it wrong. It's awkward to `reject` a promise with anything but an `Error` object, and populating code and data from that reliably was also awkward. Maybe what I'm doing now where if you throw an error the `message` is automatically brought along is also problematic. We don't really want to bring a whole random error object along the wire, so probably the framework should be catching and translating most errors. I'll think about it some more.

I also found that a developer would have to lookup our codes to know what a given error means, whereas using string identifiers makes it self-evident, whether in the debugger or in logs. (of course we'd still recommend they use our constants to actually write error handling code).

lukasIO: on the JS side we are extending the native `Error` with a `LiveKitError` class that includes a error code, maybe a similar pattern could work here

lukasIO: not sure if you're suggesting the error is _only_ a string message, a string status or both.

bcherry: Yeah right now it's a single string. Another complication is that this has to work across every language: an error thrown in the swift SDK needs to make it across and come out the other side in python, etc. I haven't done a detailed yet but I'm more confident in the portability of a single string rather than a code + message.

I updated the prototype to make this cleaner and also to catch unexpected errors thrown in the handler, which were previously getting sent across the wire and that seemed wrong: https://github.com/livekit/node-sdks/pull/276/files#diff-6b1f6b4d4de8fecd4d8dbf20eb3b46edd692b6fc4681b694172ad43a02862fcfR59

protobufs/livekit_models.proto
```diff
@@ -302,6 +305,32 @@ message ChatMessage {
...
+message RpcResponse {
+  string request_id = 1;
+  oneof value {
+    string payload = 2;
+    RpcError error = 3;
+  }
+}
+
+message RpcError {
+  string name = 1;
```
davidzhao, line 329: do we want error codes instead of name ? for a lot of our RPCs, we return codes similar to HTTP in order to make it easier on users.

davidzhao (commented):
just one comment. otherwise lgtm!

protobufs/livekit_models.proto
bcherry: the complexity here is that RpcError needs to support built-in errors _and_ user definable errors, but we could make them separate types if we wanted.

I think for our errors, number codes are probably fine (esp since there's a message field) but for user-defined errors string is much easier and doesn't have weird issues like reserved number space.

davidzhao: I don't think we necessarily need a "reserved" space.. personal opinion is that numbers make it easier to reason with vs strings. 

psrpc usage is similar.. where psrpc is the protocol, but different services would return different errors, while still adhering to HTTP-like status codes.

bcherry: by reserved space I meant that we currently use name to mean a specific error, not a class of error.  if it's http-like then it's classes, which is different.

 these are the errors the [prototype](https://github.com/livekit/node-sdks/pull/276) has found cause to implement:

```
  CONNECTION_TIMEOUT (no ack)
  RESPONSE_TIMEOUT (no response)
  UNSUPPORTED_METHOD (no registered handler)
  RECIPIENT_DISCONNECTED (recipient dropped while responding)
  UNCAUGHT_ERROR (some random error was thrown in the handler - note that users may throw our error class instead if they want it passed through as-is)
  MALFORMED_RESPONSE (this is likely to go away when we switch from JSON to protobuf)
  PAYLOAD_TOO_LARGE (your request included too much data OR the response included too much data)
  ```
  
  these could be made HTTP-like as:
  
  ```
   503 connection timeout
   504 response timeout, recipient disconnected
   404 method not found
   500 response payload too large
   400 request payload too large
 ```

is that what you mean?

Comment (lukasIO) [reply protobufs/livekit_models.proto:329]: we use an enum with a reserved space for metrics, guess we could do the same thing here?

Comment (bcherry) [reply protobufs/livekit_models.proto:329]: yeah i had it that way originally but I think it comes down to the developer ergonomics:


error handling (swifty pseudo-code)

```
try {
   performRpcRequest(....)
} catch (error: RpcError) {

   if (error.name == LKUnsupportedMethod) {
         log(error) // { name: "lk-unsupported-method", message: "this method has not been registered for RPC"}
      // handling for some livekit built in
   } else if (error.name == "My-Special-Error") {
            log(error) // { name: "My-Special-Error", message: "something terrible has happened"}
      // handling for your own error
   }
}
``` 

error throwing (swifty pseudo-code)

```
registerRpcHandler('my-method', { request in
   throw new RpcError("My-Special-Error", "something terrible has happened"); // forwarded
   throw new Error(....); // converted to LKUncaughtError
}
```

Those feel good to me. I feel like if you convert the names to numbers then it just feels weird throwing your own, because you're required to define a strict code system? And you need to ensure the message is quite explanatory or else the log lines will require consulting your error table to make sense of.

I don't feel too strongly about it and am happy to change if someone else does feel strongly.  Otherwise I think I'll keep it as I have it....

(btw HTTP seems appealing but since RPC doesn't map 1:1 to HTTP it makes it so you still really need error names, at which point you've turned the three-field error into a four-field error and I'm not sure it's particularly useful? I could be wrong)

lukasIO: > personal opinion is that numbers make it easier to reason with vs strings.

I share @davidzhao's opinion, probably because an `errorCode` as number is quite common in general and also because we use numbers already in some other places. 
An advantage, I believe, is the possibility of grouping errors with numbers (that's the only similarity I'd see to HTTP statuses btw). 
But not a hill I'm going to die on.

bcherry: Ok after more reflection and reading opinions on the internet I think it's right to switch back to numeric codes as in the original proposal - one per error. And we'll populate the message as well, based on the code (for our own errors).  We'll reserve the space 1001-1999 for framework-level errors and suggest that developers use the space 101-999 for application-level errors. And I guess I'll store it as a unit32 (wish they had unit16 in proto3!)

protobufs/livekit_models.proto
bcherry: keeping version in - hoping to not need it!

PR #816, opened 2024-09-19
Add TransferParticipant to sip and sip internal services

biglittlebigben:


biglittlebigben:
Current implementation is for the RPC to wait until the transfer succeeds or fails (ie until the new peer picks up the phone). This means that there is no transfer status API/attributes on the participant. The transfer can take longer than the default RPC timeout though, so this may or may not be the way we want to go eventually.

dennwc (approved):

PR #821, opened 2024-09-23
Client metrics protocols

davidliu:


protobufs/livekit_metrics.proto
```diff
@@ -23,6 +23,20 @@ enum MetricLabel {
   AGENTS_LLM_TTFT = 0; // time to first token from LLM
   AGENTS_STT_TTFT = 1; // time to final transcription
   AGENTS_TTS_TTFB = 2; // time to first byte
+  CLIENT_FREEZE_COUNT = 3;                            // Number of video freezes
```
boks1971, line 26: Can we make these a little longer? Sorry, would be good to have these in the name I think.
Something like
`CLIENT_VIDEO_SUBSCRIBER_<property>`
include the media type and direction in the name.

protobufs/livekit_metrics.proto
```diff
@@ -46,6 +62,7 @@ message TimeSeriesMetric {
   uint32 participant_identity = 2; // index into `str_data`
   uint32 track_sid = 3; // index into `str_data`
   repeated MetricSample samples = 4;
+  uint32 rid = 5; // index into 'str_data'
```
boks1971, line 65: I think it is okay to do `string` for this if it makes it simple rather than making a `str_data` entry and indexing into it.

boks1971 (approved):

protobufs/livekit_metrics.proto
davidliu: Hmm, it's not a big deal either way, saves a bit on space as well.

Not familiar with protobufs, but does this need to be optional?

boks1971: Not really necessary I think. It won't get encoded on wire if it has default value.

davidliu: Since default value is zero, would that be an issue knowing whether rid is 0 or null?

davidliu: I guess it might just be better to do rid as string after all.

boks1971: The index starts at 4096 though, right?

davidliu: Ah right, forgot about that.

PR #826, opened 2024-09-26
SIP features

frostbyte73:


protobufs/rpc/io.proto
```diff
@@ -165,7 +165,14 @@ message EvaluateSIPDispatchRulesResponse {
   // Map SIP X-* headers from INVITE to SIP participant attributes.
   // Keys are the names of X-* headers and values are the names of attributes they will be mapped to.
   map<string, string> headers_to_attributes = 14;
-  // NEXT ID: 15
+
+  repeated SIPFeature enabled_features = 15;
```
dennwc, line 169: Should be added to `InternalCreateSIPParticipant` as well to support outbound calls.

Also, I know Krisp is Cloud-only, but other features might not be. Probably worth adding those to `InboundTrunk` and `OutboundTrunk` as well, so that the user can configure other features that we implement.

biglittlebigben: I agree that we'll probably need this in InboundTrunk and OutboundTrunk eventually, but I'd rather wait we wait until we actually need them to add them there.

biglittlebigben: LGTM after addressing @dennwc's comments

davidzhao: `I'd rather wait we wait until we actually need them to add them there.`

wdym? how would a user specify if they want this on vs off?

dennwc: I was actually assuming there's some project-wide setting for Krisp. If there's none, we definitely need this on the Trunk level.

biglittlebigben: If this would only be used for OSS (ie cloud sets up processor as a project level setting), we should wait until we have a OSS processor to add it. If this should be settable at the trunk level for cloud, we should add it.

davidzhao: I believe trunk level is the right thing to do. It would make it possible for users to test the effect of Krisp.

this matches Telnyx's noise cancellation setting as well, it's set at the connection level

frostbyte73: That will be a separate set of PRs that adds it to trunk settings, cloud IO, etc

PR #828, opened 2024-10-02
Add disconnect reasons for SIP.

dennwc:


boks1971 (approved):

protobufs/livekit_models.proto
```diff
@@ -443,6 +443,10 @@ enum DisconnectReason {
   SIGNAL_CLOSE = 9;
   // the room was closed, due to all Standard and Ingress participants having left
   ROOM_CLOSED = 10;
+  // SIP callee did not respond in time
+  USER_UNAVAILABLE = 11;
+  // SIP callee rejected the call (busy)
+  REJECTED_BY_USER = 12;
```
davidzhao, line 449: nit: `USER_REJECTED` would be more consistent with `actor_action` pattern. otherwise lgtm

PR #831, opened 2024-10-03
Add SIP analytics events

biglittlebigben:


protobufs/livekit_analytics.proto
```diff
@@ -110,8 +111,17 @@ enum AnalyticsEventType {
...
   INGRESS_UPDATED = 29;
-
-  // NEXT_ID: 30
+  SIP_INBOUND_TRUNK_CREATED = 30;
+  SIP_INBOUND_TRUNK_DELETED = 31;
+  SIP_OUTBOUND_TRUNK_CREATED = 32;
+  SIP_OUTBOUND_TRUNK_DELETED = 33;
+  SIP_DISPATCH_RULE_CREATED = 34;
+  SIP_DISPATCH_RULE_DELETED = 35;
+  SIP_PARTICIPANT_CREATED = 36;
```
dennwc, line 120: `SIP_PARTICIPANT_CREATED` is basically `SIP_CALL_STARTED` + outbound flag. If we are adding this one, we probably should add `SIP_CALL_ACCEPTED` to reflect the same event for inbound.

protobufs/livekit_analytics.proto
```diff
@@ -110,8 +111,18 @@ enum AnalyticsEventType {
...
+  SIP_CALL_ACCEPTED = 37;
+  SIP_CALL_STARTED = 38;
+  SIP_CALL_ENDED = 39;
```
davidzhao, line 123: does this differentiate between inbound and outbound calls? should we have a separate status for `CONNECTING` or is that what STARTED is? for example the flow for inbound:
* CONNECTING -> when sip picks up, but no user on the other side
* ACCEPTED -> when sip connects to end user.

and for outbound
* CONNECTING -> when we start dialing for outbound
* ACCEPTED -> when the other side picks up

protobufs/livekit_sip.proto
```diff
@@ -416,3 +417,23 @@ message TransferSIPParticipantRequest {
   string room_name = 2;
   string transfer_to = 3;
 }
+
+message SIPCallInfo {
+  string call_id = 1;
+  string trunk_id = 2;
+  string room_id = 3; // ID of the current/previous room published to
```
davidzhao, line 424: should we include the name as well?

Note to @dennwc, room creation is async, and it could take a few seconds before the room ID resolves.

davidzhao (commented):
lg, just a couple of Qs

protobufs/livekit_analytics.proto
biglittlebigben: In the current proposal, PARTICIPANT_CREATED and CALL_ACCEPTED match the "CONNECTING" state you suggest in the outbound and inbound cases respectively. I guess ACCEPTED may be confusing as it means that the SIP server accepted it, not the room. Maybe use "CALL_INCOMING"?

SIP_PARTICIPANT_CREATED has the pro that it relates directly to an API call IMHO.

biglittlebigben: Or CALL_ACCEPTED -> RINGING?

davidzhao: got it. so accepted is for incoming only.. makes sense.. then I would vote for `INCOMING`, since it's a bit more explicit about that fact. wdyt?

protobufs/livekit_sip.proto
```diff
@@ -441,7 +441,8 @@ message SIPUri {
 }
 
 enum SIPCallStatus {
-  SIP_CALL_STARTING = 0;
-  SIP_CALL_ACTIVE = 1;
-  SIP_CALL_DISCONNECTED = 2;
+  SIP_CALL_STATUS_CALL_INCOMING = 0;
```
davidzhao, line 444: kind of a nit.. these enums are pretty long.. wdyt about using abbrevations like `SIP_` or `SCS_`

davidzhao: and since it could be confusing what each status means, wanna just put comments in here?

davidzhao (approved):

PR #844, opened 2024-10-09
Add ringing timeout and max call duration setting for SIP.

dennwc:


protobufs/livekit_sip.proto
```diff
@@ -401,7 +407,12 @@ message CreateSIPParticipantRequest {
   // If true, a random value for identity will be used and numbers will be omitted from attributes.
   bool hide_phone_number = 10;
 
-  // NEXT ID: 11
+  // Max time for the callee to answer the call.
+  google.protobuf.Duration ringing_timeout = 11;
+  // Max call duration.
+  google.protobuf.Duration max_duration = 12;
```
biglittlebigben, line 413: I think `max_call_duration` may be clearer here

dennwc: Done!

davidzhao (approved):
lg

PR #845, opened 2024-10-09
Add disconnect reason for SIP failures.

dennwc:
We can now indicate that LiveKit participant associated with SIP has left particularly because of a SIP protocol failure.


protobufs/livekit_models.proto
```diff
@@ -447,6 +447,8 @@ enum DisconnectReason {
   USER_UNAVAILABLE = 11;
   // SIP callee rejected the call (busy)
   USER_REJECTED = 12;
+  // SIP protocol failure or unexpected response
+  USER_FAILURE = 13;
```
davidzhao, line 451: do you have some examples of how this would fail? `USER_FAILURE` seems too generic.. and it seems like it's not an explicit user action

dennwc: A typical failure would be the SIP server we are trying to dial rejecting our request (e.g. wrong password, wrong number, etc).

Agree, the name is not the best one. I considered `EXTERNAL_FAILURE`, `INTEGRATION_FAILURE` or `CONNECTOR_FAILURE` to indicate that it's not an issue with LiveKit, but with the external service.

davidzhao: `SIP_TRUNK_FAILURE` ?

dennwc: That would work, but we specifically omitted `SIP_` prefix in other cases. Maybe we could also use a name that could be reused for other integrations?

dennwc: Changed to `SIP_TRUNK_FAILURE`.

davidzhao (approved):

PR #860, opened 2024-10-16
add monotonic time util

paulwe:


boks1971 (approved):

utils/mono/mono.go
```diff
@@ -7,8 +7,12 @@ var (
 	epochNano = epoch.UnixNano()
 )
 
+func FromTime(t time.Time) time.Time {
+	return epoch.Add(t.Sub(epoch))
+}
+
 func Now() time.Time {
-	return epoch.Add(time.Since(epoch))
+	return FromTime(time.Now())
```
boks1971, line 15: This could make it non-monotonic if `now` changes? i. e. taking `Now()` is not monotonic.

boks1971: to correct myself here, @paulwe set me right, `time.Now()` has both wall and monotonic time and `t.Sub` in `FromTime` will use the monotonic time.

PR #862, opened 2024-10-17
add manifest location to egress info

frostbyte73:


protobufs/rpc/egress.proto
```diff
@@ -58,6 +58,8 @@ message StartEgressRequest {
   string room_id = 3;
   string token = 8;
   string ws_url = 9;
+
+  bool cloud_backup_enabled = 10; // cloud only
```
biglittlebigben, line 62: Are we ok with cloud only concepts leaking into the TWIRP apis?

biglittlebigben (approved):

PR #863, opened 2024-10-18
Replace protoyaml with fully opensource analog

t43rr7:
This PR proposes replacement for buflint's `protoyaml` library as it's impossible to use LiveKit in some regions due the buflint policy.
The solution I found originally belongs to [this guy](https://github.com/spinnaker/kleat/blob/master/internal/protoyaml/protoyaml.go). 
All tests had passed.


t43rr7:
Note: I noticed that some test failed. This is not related to this PR, all tests had passed locally on my PC with the same command used in CI. Also I checked out the test that failed and quickly run it with debugger, marshaller hasn't even been called.

davidzhao:
Could you elaborate what about buflint's policy that makes it impossible to use?

don't worry about the failing test, one of the test can be a bit flakey :(

t43rr7:
> Could you elaborate what about buflint's policy that makes it impossible to use?

Of course. It's impossible to use buflint in Russia and Belorussia. 

I sure that the professional community should stay out of any political conflicts, but the buflint doesn't think so, saying "No Russian and Belorussian people, you are bad! I forbid to lint your go code!"

It's impossible to access their solutions located in their infrastructure without VPN or GOPROXY. Therefore, it's impossible to use any LiveKit open source tools and libraries due to how go dependencies work.

t43rr7:
@davidzhao ping :) 
What do you think?

PR #864, opened 2024-10-20
Improve usability of room configuration

davidzhao:
- allowing setting configuration parameters directly
- removes dispatch wrapper to simplify agents specification


auth/grants.go
```diff
@@ -19,16 +19,44 @@ import (
 	"strings"
 
 	"golang.org/x/exp/slices"
+	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/livekit/protocol/livekit"
+	"github.com/livekit/protocol/utils"
 )
 
+type RoomConfiguration livekit.RoomConfiguration
```
davidzhao, line 28: wrapper type to ensure it's camelCase in JWT

protobufs/livekit_room.proto
```diff
@@ -66,7 +66,7 @@ message CreateRoomRequest {
   // name of the room
   string name = 1;
   // configuration to use for this room parameters. Setting parameters below override the config defaults.
-  string config_name = 12;
+  string room_preset = 12;
```
davidzhao, line 69: renaming this too to reduce potential confusion with `_name` suffix

protobufs/livekit_room.proto
```diff
@@ -211,16 +212,17 @@ message RoomConfiguration {
...
-  RoomAgent agent = 6;
   // playout delay of subscriber
   uint32 min_playout_delay = 7;
   uint32 max_playout_delay = 8;
   // improves A/V sync when playout_delay set to a value larger than 200ms. It will disables transceiver re-use
   // so not recommended for rooms with frequent subscription changes
   bool sync_streams = 9;
+
+  // Define agents that should be dispatched to this room
+  repeated RoomAgentDispatch agents = 10;
```
davidzhao, line 227: direct specifications of agents

auth/grants.go
```diff
@@ -19,16 +19,44 @@ import (
...
 	Identity string      `json:"-"`
 	Name     string      `json:"name,omitempty"`
 	Kind     string      `json:"kind,omitempty"`
 	Video    *VideoGrant `json:"video,omitempty"`
 	SIP      *SIPGrant   `json:"sip,omitempty"`
+	// Room configuration to use if this participant initiates the room
+	Room *RoomConfiguration `json:"room,omitempty"`
+	// Cloud-only, config preset to use
+	// when both room and roomPreset are set, parameters in room overrides the preset
+	RoomPreset string `json:"roomPreset,omitempty"`
```
boks1971, line 59: are we using `roomPreset` or `room_preset` in config?

auth/grants.go
```diff
@@ -69,8 +105,6 @@ type VideoGrant struct {
 	RoomAdmin bool   `json:"roomAdmin,omitempty"`
 	RoomJoin  bool   `json:"roomJoin,omitempty"`
 	Room      string `json:"room,omitempty"`
```
boks1971, line 107: There is already `room` in `VideoGrant`. Would top-level `room` cause confusion?

boks1971 (approved):

auth/grants.go
davidzhao: protojson serialization serializes to camelCase by default

auth/grants.go
davidzhao: good point.. i'll rename to `roomConfig`

auth/grants.go
```diff
@@ -19,16 +19,44 @@ import (
...
+
 type ClaimGrants struct {
 	Identity string      `json:"-"`
 	Name     string      `json:"name,omitempty"`
 	Kind     string      `json:"kind,omitempty"`
 	Video    *VideoGrant `json:"video,omitempty"`
 	SIP      *SIPGrant   `json:"sip,omitempty"`
+	// Room configuration to use if this participant initiates the room
+	RoomConfig *RoomConfiguration `json:"roomConfig,omitempty"`
+	// Cloud-only, config preset to use
```
paulwe, line 57: in oss this is loaded from `RoomConfig.RoomConfigurations`

paulwe (approved):

auth/grants.go
davidzhao: yeah. we could change that as well, to be a bit more concise.

biglittlebigben: These can be quite large. Are we ok with long tokens?

biglittlebigben (approved):

auth/grants.go
davidzhao: yeah I think that's fine.. if their use case requires it.

for egress usage though, we'll need to advise users to not put their egress keys in there, but instead to rely on "default storage configurations" that's specific to the project.

PR #865, opened 2024-10-21
Allow requesting a ringtone during SIP call transfers

biglittlebigben:


protobufs/rpc/sip.proto
```diff
@@ -99,4 +99,7 @@ message InternalCreateSIPParticipantResponse {
 message InternalTransferSIPParticipantRequest {
   string sip_call_id = 1;
   string transfer_to = 2;
+ 
+  // Optionally play ringtone to the SIP participant as an audible indicator of being transferred
+  bool play_ringtone = 3;
```
davidzhao, line 104: I guess this terminology is already there.. but technically this is the `dialtone`, not ringtone

davidzhao: should we make a backwards compatible change now? wdyt @biglittlebigben @dennwc ?

biglittlebigben: Yes, it's wrong, but everybody calls it that way, including our customers. I'm all for changing it, but by backward compatible change, do you mean "wire protocol compatible", ie rename the field with the same id, or code compatible (ie new field and play the dial tone if either of the boolean is true)?

davidzhao: I was thinking code.. because we use twirp APIs.. and there isn't a clean way to change the column names there. but don't feel strongly that we need to change it.

biglittlebigben: True for twirp. Folks will need to set both for some time, until we deploy the new flag everywhere, which is not ideal...

biglittlebigben: Just renaming the field ought to be safe in the internal RPC protobuf though?

davidzhao: yeah internal RPC is totally fine.

davidzhao (approved):

PR #866, opened 2024-10-21
Krisp inbound trunk setting

frostbyte73:


protobufs/livekit_sip.proto
```diff
@@ -195,6 +195,8 @@ message SIPInboundTrunkInfo {
   google.protobuf.Duration ringing_timeout = 11;
   // Max call duration.
   google.protobuf.Duration max_call_duration = 12;
+
+  bool krisp_enabled = 13;
```
biglittlebigben, line 199: What is the motivation to use boolean here rather than with a `repeated` feature field like for the rpc?

frostbyte73: I did it here because it's specific to SIPInboundTrunkInfo, while future features might not apply

biglittlebigben: supported vs unsupported features can be dealt with request validation early in the request processing. Unique booleans for each feature will require us to keep adding fields to the messages, which may become quite long. It would work from a functionality perspective though, of course. Curious what other folks think.

davidzhao: it seems clearer to have the Trunk definition to be explicit and flat.

davidzhao (approved):

PR #868, opened 2024-10-24
krisp for outbound trunks

frostbyte73:


protobufs/livekit_sip.proto
```diff
@@ -232,6 +232,8 @@ message SIPOutboundTrunkInfo {
   // Map SIP X-* headers from 200 OK to SIP participant attributes.
   // Keys are the names of X-* headers and values are the names of attributes they will be mapped to.
   map<string, string> headers_to_attributes = 10;
+
+  bool krisp_enabled = 11;
```
davidzhao, line 236: can we set this on the CreateSIPParticipant API instead? For outbound calls, it would give better control if it could be enabled or disabled for each call.

for inbound calls, since there's no explicit APIs, we'd have to leave any customizable options on the trunk

protobufs/livekit_sip.proto
```diff
@@ -418,6 +416,8 @@ message CreateSIPParticipantRequest {
   // Max call duration.
   google.protobuf.Duration max_call_duration = 12;
 
+  // Enable voice isolation for the callee.
+  bool enable_krisp = 14;
   // NEXT ID: 14
```
davidzhao, line 421: nit: 
```suggestion
  // NEXT ID: 15
```

davidzhao (approved):

PR #871, opened 2024-10-25
Add DataStream support

lukasIO:


protobufs/livekit_models.proto
```diff
@@ -601,3 +601,60 @@ enum AudioTrackFeature {
...
+  
+    // oneof to choose between specific header types
+    oneof content_header {
+      TextHeader textHeader = 9;
+      FileHeader fileHeader = 10;
+    }
+  }
+
+  message Packet {
+    string messageId = 1;  // unique identifier for this data stream to map it to the correct header
```
boks1971, line 648: not calling this `streamId`?

protobufs/livekit_models.proto
```diff
@@ -601,3 +601,60 @@ enum AudioTrackFeature {
...
+    uint64 chunkIndex = 2; // chunkId as an unsigned 32-bit integer
```
boks1971, line 649: comment says 32-bit, type here is 64-bit. Guessing 32-bit should be enough for this?

protobufs/livekit_models.proto
```diff
@@ -601,3 +601,60 @@ enum AudioTrackFeature {
...
+    bytes content = 3; // content as binary (bytes)
+    uint32 contentLength = 4; // content length in bytes
```
boks1971, line 651: what is the purpose of this? can this not be derived from `content` as length of it? Are there cases where content is padded?  (maybe for the encryption case?)

protobufs/livekit_models.proto
boks1971: oh, looks like you are making both length and chunkId 64-bit, comment needs update then.

protobufs/livekit_models.proto
```diff
@@ -601,3 +601,60 @@ enum AudioTrackFeature {
   TF_ENHANCED_NOISE_CANCELLATION = 5;
 }
 
+message DataStream {
+
+  // enum for operation types (specific to TextStreamHeader)
```
boks1971, line 606: :nit: there is no message called `TextStreamHeader`. When it is written like that, it makes it look like a type and I would want to search for it to look at the shape of it.

protobufs/livekit_models.proto
```diff
@@ -601,3 +601,60 @@ enum AudioTrackFeature {
...
+    bool generated = 5; // true if the text has been generated by an agent from a participant's audio transcription
+
+  }
+  
+  // header properties specific to file or image streams
+  message FileHeader {
+    string fileName = 1;  // optional: name of the file
+  }
+
+  // main DataStreamHeader that contains a oneof for specific headers
```
boks1971, line 629: :nit: maybe change comment to say `DataStream.Header` to signify that it is a not a type of its own.

protobufs/livekit_models.proto
```diff
@@ -601,3 +601,60 @@ enum AudioTrackFeature {
...
+    int32 version = 2;  // Optional: Version for updates/edits
+    string replyToStreamId = 3;  // Optional: Reply to specific message
+    repeated string attachedStreamIds = 4;
+    bool generated = 5; // true if the text has been generated by an agent from a participant's audio transcription
+
+  }
+  
+  // header properties specific to file or image streams
+  message FileHeader {
+    string fileName = 1;  // optional: name of the file
```
boks1971, line 626: It says optional, but shouldn't this be mandatory for type file?

boks1971 (approved):

protobufs/livekit_models.proto
lukasIO: yeah, although I'm not sure anymore what we'd need it for, even in the encryption case. Will remove and we can re-add if it turns out we need it

protobufs/livekit_models.proto
lukasIO: done!

protobufs/livekit_models.proto
lukasIO: thanks for catching this, was an oversight

cnderrauber (approved):

theomonnom (approved):

nbsp (approved):

PR #880, opened 2024-11-08
EgressInfo helpers

frostbyte73:


livekit/livekit_egress.go
```diff
@@ -0,0 +1,85 @@
...
+func (e *EgressInfo) UpdateStatus(status EgressStatus) {
+	e.Status = status
+	e.UpdatedAt = time.Now().UnixNano()
+}
+
+func (e *EgressInfo) SetBackupUsed() {
+	e.Status = EgressStatus_EGRESS_FAILED
+	e.UpdatedAt = time.Now().UnixNano()
+	e.Error = MsgBackupUsed
+	e.ErrorCode = int32(http.StatusPermanentRedirect)
```
biglittlebigben, line 36: How will we communicate the case where a segmented file egress eventually ended as failure, but where some segments have failed to upload as well? Customers may want to download whatever segments are available, but will need a way to see that some are only available on the backup storage.

What I'm getting at is that I'm wondering if backup status should be orthogonal to the egress status/error state.

frostbyte73: True, I think we'll need to add something else because even a separate backup_used state wouldn't cover that

PR #883, opened 2024-11-13
chore: support CJS

nbsp:
this PR changes the CI to run two protobuf generators, one for each, plus another for dts. currently untested.


packages/javascript/src/index.d.ts
```diff
@@ -1,12 +1,12 @@
...
-export * from "./gen/livekit_egress_pb.js";
-export * from "./gen/livekit_ingress_pb.js";
-export * from "./gen/livekit_metrics_pb.js";
-export * from "./gen/livekit_models_pb.js";
-export * from "./gen/livekit_room_pb.js";
-export * from "./gen/livekit_rtc_pb.js";
-export * from "./gen/livekit_sip_pb.js";
-export * from "./gen/livekit_webhook_pb.js";
-export * from "./gen/version.js";
+export type * from "./gen/dts/livekit_agent_dispatch_pb.d.ts";
```
lukasIO, line 2: hm, don't remember ever seeing this import pattern 🤔

packages/javascript/src/index.js
```diff
@@ -1,13 +1,28 @@
...
-export * from "./gen/livekit_ingress_pb.js";
-export * from "./gen/livekit_metrics_pb.js";
-export * from "./gen/livekit_models_pb.js";
-export * from "./gen/livekit_room_pb.js";
-export * from "./gen/livekit_rtc_pb.js";
-export * from "./gen/livekit_sip_pb.js";
-export * from "./gen/livekit_webhook_pb.js";
-export * from "./gen/version.js";
+const { protoInt64 } = require("@bufbuild/protobuf");
+const agentDispatch = require("./gen/esm/livekit_agent_dispatch_pb.js");
```
lukasIO, line 3: requiring ESM will fail though?

lukasIO: this shouldnimport from cjs I think?

packages/javascript/package.json
```diff
@@ -2,25 +2,32 @@
   "name": "@livekit/protocol",
   "version": "1.27.1",
   "description": "",
-  "type": "module",
```
lukasIO, line 5: why not leave that in and make the ending for cjs .cjs like in the node repo?

nbsp: we could do that, yeah. i was just keeping in line with components-js which doesnt use type:module either. your call

packages/javascript/src/index.d.ts
nbsp: the LSP didn't complain, but i'll double check to make sure this works

packages/javascript/package.json
lukasIO: My vote goes to consistency, we can set type module here and in node sdks and adapt it in components or use the existing components pattern everywhere. No strong opinion either way, just that we use the same pattern would be nice

packages/javascript/src/index.d.ts
nbsp: works great, except the files under `cjs` need to be renamed to `.cjs` in order to be required as CommonJS. i don't know if GH Actions comes built in with any rename utility

nbsp: ok so i did a bulkrename but now i need to sed through all the files and replace the names as well

nbsp:
tested working on livekit-server-sdk (which breaks for a different reason, will be handled there, this part is fine)

packages/javascript/package.json
```diff
@@ -4,23 +4,31 @@
...
-      "default": "./src/index.js"
+      "import": "./src/index.js",
+      "require": "./src/index.cjs"
     },
     "./*": {
-      "types": "./src/gen/livekit_*_pb.d.ts",
-      "default": "./src/gen/livekit_*_pb.js"
+      "types": "./src/gen/dts/livekit_*_pb.d.ts",
+      "import": "./src/gen/esm/livekit_*_pb.js",
+      "require": "./src/gen/cjs/livekit_*_pb.js"
```
lukasIO, line 21: do these need a .cjs ending?

packages/javascript/src/index.d.ts
lukasIO: would be nice if we could configure that as part of the protocol es plugin, I think file extension is an option there?

nbsp: only able to set js/ts/d.ts on there unfortunately

nbsp:
tested building downstream with esbuild (agents + tsup)

packages/javascript/src/index.js
```diff
@@ -10,4 +10,4 @@ export * from "./gen/livekit_room_pb.js";
 export * from "./gen/livekit_rtc_pb.js";
 export * from "./gen/livekit_sip_pb.js";
 export * from "./gen/livekit_webhook_pb.js";
-export * from "./gen/version.js";
+export * from "./gen/version.mjs";
```
lukasIO, line 13: why would we mix .js and .mjs imports here?

nbsp: the order of operations for the build is genversion → js (as CJS) → rename to cjs → js (as ESM).
if we genversion as `js` and not `mjs`, it'll get translated to cjs and overwrite the other one.
another approach would be to run genversion&js (CJS) → rename → genversion&js (ESM), but i feel like that's worse

lukasIO: I see, but there's a separate `index.cjs` file, what's the relation to that then?

nbsp: that's because the only files getting renamed from js to cjs are inside gen.
basically:

- `genversion` (creates version.{m,c}js, that won't be renamed)
- `protoc` generates CJS files
- we rename all of those newly-generated files to `.cjs` from `.js` (this is the part where we can't genversion as js, because it'll override the cjs file)
- `protoc` as ESM.

PR #889, opened 2024-11-24
Utility to build livekit.MetricsBatch

boks1971:
Builder for livekit.MetricsBatch. As that message uses a dictionary, this helper allows adding by values and build the dictionary.

Also adding a few server side metric labels.


utils/metrics_batch_builder.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	return false
+}
+
+func (m *MetricsBatchBuilder) getStrDataIndex(s string) uint32 {
+	idx, ok := m.stringData[s]
+	if !ok {
+		m.MetricsBatch.StrData = append(m.MetricsBatch.StrData, s)
+		idx = uint32(int(livekit.MetricLabel_METRIC_LABEL_PREDEFINED_MAX_VALUE) + len(m.MetricsBatch.StrData) - 1)
```
boks1971, line 300: Using  the pre-defined offset of 4096 for all entities pointing to `StrData`. Technically, the offset is needed only for the label. If we do that, for the other fields that index into `StrData` (like `ParticipantIdentity` etc.) should either
- become `optional` so that `0` value can be distinguished from `not present` OR
- use offset of `+1` to indicate index into `StrData` and `0` can be interpreted as `not present`.

Tagging @davidliu for any changes needed in Android if we decide to change this. I told him when he was working on this several weeks back to do `4096` as base offset for all.

paulwe (approved):

PR #892, opened 2024-11-25
SIP media encryption setting.

dennwc:
Add SIP media encryption setting to support SRTP. Note that the default is to require encryption if TLS transport is used.

Requires https://github.com/livekit/protocol/pull/891.


protobufs/livekit_sip.proto
```diff
@@ -197,6 +205,7 @@ message SIPInboundTrunkInfo {
   google.protobuf.Duration max_call_duration = 12;
 
   bool krisp_enabled = 13;
+  SIPMediaEncryption media_encryption = 14;
```
biglittlebigben, line 208: Similarly to krisp, this allows to set the encryption setting for each call (participant) for outbound, but makes it a trunk global setting for inbound. Is there any reason not to put these 2 settings on the dispatch? This seems it would be more consistent.

protobufs/livekit_sip.proto
```diff
@@ -110,6 +110,13 @@ enum SIPTransport {
   SIP_TRANSPORT_TLS = 3;
 }
 
+enum SIPMediaEncryption {
+  SIP_MEDIA_ENCRYPT_WITH_TLS = 0; // require encryption with TLS
```
davidzhao, line 114: I would vote for a more lenient default. Especially since it's already possible to use TLS but without SRTP. Could we change the default to _ALLOW?

dennwc: I agree that this will be a sensible default in general.

My thinking was that we should keep the current behavior to not accidentally break someone's setup (since we may discover bugs initially). TLS was introduced only recently, and I assume that customers who explicitly enabled it require full encryption anyway. Thus the "require is TLS" is set as a default here.

Also, I think that setting it as a default for UDP might break some existing trunks, because we will include all supported encryption algorithms and keys for them in SDP, which will most likely push the packet size above MTU. So I think if we want the `ALLOW` option to be a safe default, we should enable it only for TCP and TLS at least.

davidzhao: > I assume that customers who explicitly enabled it require full encryption anyway. Thus the "require is TLS" is set as a default here.

I think this assumption could break existing apps. In practice we have no way of knowing what users of TLS are assuming. Also, even if that might be their desire, the current setup on the trunk provider side might not be configured to support SRTP. For both of those reasons picking SRTP-required default would seem risky.

From a compatibility standpoint, I'm of the opinion that the "default" should be the most compatible. Those requiring tighter security should state that explicitly.

dennwc: OK, in that case we should set `DISABLE` as a default.

protobufs/livekit_sip.proto
```diff
@@ -154,7 +160,8 @@ message SIPTrunkInfo {
   // User-defined metadata for the Trunk.
   string metadata = 12;
 
-  // NEXT ID: 15
+  SIPMediaEncryption media_encryption = 15;
```
biglittlebigben, line 163: Isn't SIPTrunkInfo deprecated? If so, do we want to add new options to it?

biglittlebigben (approved):

protobufs/livekit_sip.proto
dennwc: Oops! Good catch! Fixed.

davidzhao (approved):
🚀

PR #894, opened 2024-11-26
Add EgressSourceType to EgressInfo 

biglittlebigben:
This adds the source type used in an egress to the EgressInfo in preparation for adding a path where the SDK handles some room composite use cases natively.

The `Web` and `SDK` constants reflect what we use internally in Egress. Other values may be clearer?


biglittlebigben:
This is needed for proper billing

frostbyte73 (approved):

PR #897, opened 2024-11-27
fix(javascript): fix CJS require importing .js

nbsp:


lukasIO (approved):
all the hoops one has to jump through 🤯

PR #899, opened 2024-11-28
upgrade to webrtc v4

cnderrauber:


boks1971 (approved):
Would this work with SIP also? Are you making necessary changes there also?

paulwe (approved):

cnderrauber:
> Would this work with SIP also? Are you making necessary changes there also?

@boks1971 Will pion/webrtc/v4 break sip?

boks1971:
> > Would this work with SIP also? Are you making necessary changes there also?
> 
> @boks1971 Will pion/webrtc/v4 break sip?

Don't think so. But, I am not aware of changes needed. Just wanted to make sure that other parts of the system do not need to maintain an alt branch for long.

For example, if SIP requires a bunch of changes for v4 (same with Go Server SDK) and porting over is going to take long, it will have to be on `alt` branch of this repo for that period of time if it needs protocol changes. That could be a bit of a pain. So, wanted to understand the impact.

boks1971:
I guess ingress/egress all use pion. So, they will have to port or work off of an alt branch till it is ported.

boks1971:
> > > > Would this work with SIP also? Are you making necessary changes there also?
> > > 
> > > 
> > > @boks1971 Will pion/webrtc/v4 break sip?
> > 
> > 
> > Don't think so. But, I am not aware of changes needed. Just wanted to make sure that other parts of the system do not need to maintain an alt branch for long.
> > For example, if SIP requires a bunch of changes for v4 (same with Go Server SDK) and porting over is going to take long, it will have to be on `alt` branch of this repo for that period of time if it needs protocol changes. That could be a bit of a pain. So, wanted to understand the impact.
> 
> It depends on how it use the dependency, will check it. I have a list of repo to upgrade includes `protocol` `mediatransportutils` `backend-common` `go-sdk` `livekit` and `cloud` to update them together, if there is other repo not included then tell me, I'll update them.

Thank you. I think `ingress`, `egress` and `sip` are the other three.

PR #905, opened 2024-11-28
Revert "chore(deps): bump go to 1.23"

davidzhao:
Reverts livekit/protocol#885

Since this package is depended on by others (including external users). we should not raise Go version requirements too aggressively.


paulwe (approved):

rektdeckard:
I believe the bump was to fix build issues in some part of closed-source due to dependency on 1.23 features (indirectly). It may be a dependency we can replace, though.

davidzhao:
> I believe the bump was to fix build issues in some part of closed-source due to dependency on 1.23 features (indirectly). It may be a dependency we can replace, though.

I think it's a dependency for counterfeiter/v6, it's compiling now after I lowered the version.. let's see if downstream deps would error again

PR #911, opened 2024-12-04
Add headers for CreateSipParticipant

s-hamdananwar:


protobufs/livekit_sip.proto
```diff
@@ -424,14 +424,22 @@ message CreateSIPParticipantRequest {
   // If true, a random value for identity will be used and numbers will be omitted from attributes.
   bool hide_phone_number = 10;
 
+  // These headers are sent as-is and may help identify this call as coming from LiveKit for the other SIP endpoint.
```
davidzhao, line 427: I think we should just keep headers, and remove the mappings fields. the mappings are a level of indirection away and it's a bit difficult to understand.

s-hamdananwar: Makes sense! Made a new commit based on this comment

davidzhao (commented):
lg, but will leave final approval to @biglittlebigben and @dennwc as we'll need to implement this in the SIP service

biglittlebigben:
LGTM, but dennwc may have more opinions than me on this.

dennwc (changes requested):
Look good, but you also need to update the [NewCreateSIPParticipantRequest](https://github.com/livekit/protocol/blob/main/rpc/sip.go#L14) function. It should now merge headers from the trunk and the request.

rpc/sip.go
```diff
@@ -54,6 +54,11 @@ func NewCreateSIPParticipantRequest(
 		features = append(features, livekit.SIPFeature_KRISP_ENABLED)
 	}
 
+	headers := req.Headers
+	if headers == nil {
+		headers = trunk.Headers
```
dennwc, line 59: You'll have to iterate over key-values and add them to merge maps properly. Here you're discarding one of the maps.

s-hamdananwar: I see. My assumption was that user would be having headers in either `CreateSIPParticipantRequest` or `SIPOutboundTrunkInfo`, and if they had in both it would be the same- hence just overriding in case one is nil. Will make the requested change

dennwc (approved):
Done! I fixed headers merging, fixed changeset and added header validation that I missed during previous review. Should be good to go now!

s-hamdananwar:
@dennwc thanks for getting this up!

PR #915, opened 2024-12-06
feat(livekit): add sip invite message to sip participant attributes

mohsenbostan:


dennwc (changes requested):
Full INVITE message could be pretty large and may be larger than LiveKit's attribute size limit.

What's the exact information you want to extract from the INVITE?

PR #920, opened 2024-12-10
Allow mapping all SIP headers to attributes.

dennwc:


protobufs/livekit_sip.proto
```diff
@@ -194,6 +200,7 @@ message SIPInboundTrunkInfo {
   // Map LiveKit attributes to SIP X-* headers when sending BYE or REFER requests.
   // Keys are the names of attributes and values are the names of X-* headers they will be mapped to.
   map<string, string> attributes_to_headers = 14;
+  SIPHeaderOptions include_headers = 15;
```
biglittlebigben, line 203: I think we may need to add some documentation here, as the interaction of attributes_to_headers and include_headers may not be obvious.

dennwc: Done! Please let me know if there's anything else missing.

biglittlebigben (approved):

PR #923, opened 2024-12-15
adding room configuration in SIP dispatch rule

davidzhao:


theomonnom (approved):

lukasIO (approved):

dennwc (commented):
Looking into it a bit more, if we want SIP to set this config on a room, we will need to pass it from IO service to SIP.

This can be done by adding the same parameter to [EvaluateSIPDispatchRulesResponse](https://github.com/livekit/protocol/blob/main/protobufs/rpc/io.proto#L134) and then passing it along in [EvaluateDispatchRule](https://github.com/livekit/protocol/blob/main/sip/sip.go#L505).

davidzhao:
got it, i'll add them before merging

PR #925, opened 2024-12-17
Relax header validation for SIP.

dennwc:
Since https://github.com/livekit/protocol/pull/920 is merged, we now allow mapping any SIP headers to attributes. Thus, do not enforce `headers_to_attributes` map to only contain `X-*` headers. While at it - properly validate header names.


livekit/sip.go
```diff
@@ -103,9 +103,24 @@ func (p *SIPOutboundTrunkInfo) AsTrunkInfo() *SIPTrunkInfo {
...
+		} else if r >= '0' && r <= '9' {
+			return false
+		}
+		switch r {
+		case '-':
+			return false
+		}
+		return true
+	}) {
+		return fmt.Errorf("invalid header name: %q", header)
```
kuba--, line 123: Can we replace it by sth. like:
```
if !regexp.MustCompile(`^[a-zA-Z0-9\-]*$`).MatchString(word) {
 return fmt.Errorf("invalid header name: %q", header)
}
```

kuba--: btw. I suppose the number full of `-` is also incorrect

dennwc: Done, switched to regexp. Full header validation is out of scope - SIP itself will drop invalid headers during parsing. This is just a best-effort validation to prevent typos.

davidzhao (approved):

PR #928, opened 2024-12-19
add agent client to retrieve agent status, metadata, load, and job count

anvh2:
### 🛠 Changes being made
- Added agent client to check if enabled/check availability for agent registration to livekit-server.
- Added metadata for workers registered to livekit-server to manage the context of the agent in the dispatch service.
- Added worker_info to retrieve agent status, metadata, load, and job count.

### 🧠 Rationale behind the change
- In the dispatch service, it is necessary to manage agents registered to livekit-server to determine which agent can be dispatched to a room.


anvh2:
Hi Team,

Could you please let me know if this PR is necessary for implementation? In my context, I want to manage the agents registered to the LiveKit server. Before starting a new room, I would like to check:
- Which script/scenario an agent can serve based on the metadata registered with the LiveKit server.
- Agent availability for handling conversations in this room. Through load testing, we can determine how many rooms a single agent worker instance can handle (including STT, TTS, VAD, and LLM).

Before dispatching, I may need to check:
- job_count per worker
- load / threshold for agent availability.

Please provide more information and context for scaling to production.

Thank you!

PR #931, opened 2024-12-22
Remove total_chunks from data streams

lukasIO:
Turns out in a lot of cases (e.g. reading files from disk) total_chunks is not yet determined when sending the header. 
So removing it entirely and instead relying on total byte size seems like the right move


davidzhao (approved):

davidzhao:
merging this one!

PR #937, opened 2025-01-07
Make IngressInfo.enabled optional

biglittlebigben:
To provide backward compatibility with already existing Ingresses


davidzhao (approved):
lg

PR #940, opened 2025-01-10
Add DataStream.Trailer for finalizing streams

lukasIO:
Instead of sending `Chunk.complete` with every data stream chunk, it makes more sense to have a dedicated closing message.

This allows us to communicate more insights on reasons why the stream stopped (esp. if it stopped unexpectedly) and also allows us to finalize some data of the stream info, e.g. the `endTime` for a transcription.


davidzhao (approved):
is the closing message required? do we have mechanisms of retrying in case it's not delivered?

lukasIO:
closing message is required for the stream to close, we do not have mechanisms today to retry it.

This isn't worse as before (the `complete: true`) message had the same restrictions, we can just be more explicit about things with a dedicated message for closing

PR #942, opened 2025-01-11
add GetAndSet to bitmap

paulwe:
precomputing the mask seems to be a very minor pessimization but we can skip one call to `getSlotAndOffset` in `(*RTPStatsReceiver).Update` by combining get/set into one method.

also some other driveby cleanup

goos: darwin
goarch: arm64
pkg: github.com/livekit/protocol/utils
cpu: Apple M1 Pro
BenchmarkBitmap
BenchmarkBitmap/Set
BenchmarkBitmap/Set-10 584310650 2.055 ns/op
BenchmarkBitmap/Set-10 586064010 2.049 ns/op
BenchmarkBitmap/Set-10 582722875 2.057 ns/op
BenchmarkBitmap/Set-10 585069157 2.063 ns/op
BenchmarkBitmap/Set-10 566389279 2.063 ns/op
PASS
ok github.com/livekit/protocol/utils 15.692s

goos: darwin
goarch: arm64
pkg: github.com/livekit/protocol/utils
cpu: Apple M1 Pro
BenchmarkBitmap
BenchmarkBitmap/Set
BenchmarkBitmap/Set-10 591135066 2.030 ns/op
BenchmarkBitmap/Set-10 592524807 2.055 ns/op
BenchmarkBitmap/Set-10 585467359 2.022 ns/op
BenchmarkBitmap/Set-10 591682300 2.024 ns/op
BenchmarkBitmap/Set-10 593558163 2.027 ns/op


boks1971 (approved):
ouch, that was a pessimisation, did not expect that.

paulwe:
i'm not sure this will have any effect either way. `Set` has higher cpu use than `IsSet` which runs unconditionally. maybe the cost is dominated by some cache operation...
Screenshot 2025-01-11 at 10 53 06 AM

Screenshot 2025-01-11 at 10 52 08 AM

in a microbenchmark the extra call has no impact

BenchmarkBitmap
BenchmarkBitmap/IsSet/Set
BenchmarkBitmap/IsSet/Set-10         	544463373	         2.092 ns/op
BenchmarkBitmap/IsSet/Set-10         	583395589	         2.140 ns/op
BenchmarkBitmap/IsSet/Set-10         	590237834	         2.048 ns/op
BenchmarkBitmap/IsSet/Set-10         	589507502	         2.046 ns/op
BenchmarkBitmap/IsSet/Set-10         	587347552	         2.043 ns/op
BenchmarkBitmap/GetAndSet
BenchmarkBitmap/GetAndSet-10         	589057398	         2.048 ns/op
BenchmarkBitmap/GetAndSet-10         	592168935	         2.044 ns/op
BenchmarkBitmap/GetAndSet-10         	591168433	         2.042 ns/op
BenchmarkBitmap/GetAndSet-10         	592299975	         2.034 ns/op
BenchmarkBitmap/GetAndSet-10         	592091506	         2.040 ns/op

boks1971:
>i'm not sure this will have any effect either way. Set has higher cpu use than IsSet which runs unconditionally. maybe the cost is dominated by some cache operation...

`IsSet` runs only for out-of-order packets. That's why it must be smaller.

boks1971:
`ClearRangeAndSet` might have a bigger impact if it can be optimised as it runs for all in-order packets. `ClearRange` will be a no-op unless there are gaps - https://github.com/livekit/livekit/blob/28c39efa06c980d9cfa9c110b26b7b721d2503fe/pkg/sfu/rtpstats/rtpstats_receiver.go#L325

paulwe:
inlining doesn't seem to have an effect either 🫤

paulwe:
>ClearRangeAndSet might have a bigger impact if it can be optimised as it runs for all in-order packets. ClearRange will be a no-op unless there are gaps - https://github.com/livekit/livekit/blob/28c39efa06c980d9cfa9c110b26b7b721d2503fe/pkg/sfu/rtpstats/rtpstats_receiver.go#L325

these don't use the same bitmap indices so i don't see anything to eliminate by merging them

PR #949, opened 2025-01-22
Participant joined_at_ms to get higher resolution of join time.

boks1971:


cnderrauber (approved):

bcherry:
would it make sense to add a `Room.creation_time_ms` while we're at? the client SDKs consume these in the same way and I also imagine there are developers doing math on room creation time vs participant join time so they should have the same resolution...

boks1971:
Will look at adding that @bcherry . One note though. room creation time and participant join time could be using different clocks in cloud as they could be on different nodes. But, it should be within the clock skew of servers which is probably okay for the kind of comparisons you have mentioned.

boks1971:
Protocol change PR - https://github.com/livekit/protocol/pull/950

PR #952, opened 2025-01-24
Add flag to make CreateSIPParticipant synchronous.

dennwc:


davidzhao (approved):
lgtm

protobufs/rpc/sip.proto
```diff
@@ -101,13 +101,19 @@ message InternalCreateSIPParticipantRequest {
 
   livekit.SIPMediaEncryption media_encryption = 28;
 
-  // NEXT ID: 29
+  // Wait for the call to dial before returning.
+  bool wait_until_dialed = 29;
```
davidzhao, line 105: on second thought. I think the name of this parameter could be confusing.. it reads as if the function will return as soon as we've dialed it..

I think the intention is to say, wait until the creation is complete? (i.e. either success or failure).. so should we call this `wait_until_complete` instead?

or `block`?

davidzhao (changes requested):

protobufs/rpc/sip.proto
dennwc: The current behavior is "wait until LK participant is created" and with the flag set it will be "wait until call is dialed".

So I was thinking that `wait_until_dialed` reflects it quite well, but could rename to `wait_until_complete` as well.

`block` on the other hand is misleading - we already blocks until participant is created. I think we need to specify exactly what is the event we'll wait for.

Would `wait_until_call_pickup` be better maybe?

dennwc: Changed to `wait_until_answered`

davidzhao (approved):

protobufs/rpc/sip.proto
```diff
@@ -101,13 +101,19 @@ message InternalCreateSIPParticipantRequest {
...
+
+  // NEXT ID: 30
 }
 
 message InternalCreateSIPParticipantResponse {
   string participant_id = 1;
   string participant_identity = 2;
   string sip_call_id = 3;
+
+  uint32 sip_failure_code = 4;
```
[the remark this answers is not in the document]
biglittlebigben: What is the context for that new convention?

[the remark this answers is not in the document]
dennwc: That's mostly to make SDKs work out of the box. Users would expect an errors in case the call fails, but for internal services we'd rather know that communication between services was successful, even though the call was not.

PR #959, opened 2025-02-03
Dynamic SIP dispatch rule

davidzhao:


dennwc (commented):
Nice! It needs a code path in `DispatchRulePriority` and also in `DispatchRulePriority` (which won't work, unless SIP will send a webhook). I assume we want to keep the current webhook delivery mechanism, right?

alfredjophy:
Any updates on this ?

alfredjophy:
Would this be open for contribution @dennwc

PR #961, opened 2025-02-03
Allow passing all SIP trunk options in CreateSIPParticipant.

dennwc:


protobufs/livekit_sip.proto
```diff
@@ -521,6 +521,7 @@ message DeleteSIPDispatchRuleRequest {
 message CreateSIPParticipantRequest {
   // What SIP Trunk should be used to dial the user
   string sip_trunk_id = 1;
+  SIPOutboundTrunkInfo sip_trunk = 20;
```
davidzhao, line 524: IMO it seems rather confusing to embed *all* of the options from a SIP Trunk here. I would vote for just including the fields that are needed explicitly
- url
- username
- password

dennwc:
@davidzhao Done, added the remaining trunk fields in a separate config proto. Or do you prefer unwrapping it into a top-level message?

Also note that it doesn't accept a raw SIP URI, instead it specifies hostname and transport separately.

davidzhao (approved):

PR #962, opened 2025-02-04
Proto to "report" different kinds of data.

boks1971:
Starting off with `FeatureUsage` with the one feature od
`KRISP_NOISE_CANCELLATION`.


protobufs/livekit_analytics.proto
```diff
@@ -202,3 +204,24 @@ message AnalyticsNodeRooms {
...
+  }
+  Feature feature = 1;
+  string project_id = 2;
+  string room_name = 3;
+  string room_id = 4;
+  string participant_identity = 5;
+  string participant_id = 6;
+  string track_id = 7;
+  google.protobuf.Timestamp started_at = 8;
+  google.protobuf.Timestamp ended_at = 9;
```
boks1971, line 226: A couple of notes/questions
1. `AnalyticsEvent` message has a bunch of common fields like `room_id`, `track_id`, but I am making this standalone message fully self-contained. Is that okay?
2. Talking about self-contained, this only has `room_name`, `room_id`, `participant_identity` and `participant_id` and it does not have full `Room` or `ParticipantInfo`. Is that okay? If full objects are needed, will have to get clients to submit it as the controller handling the report will not have all context.

protobufs/livekit_analytics.proto
```diff
@@ -121,8 +121,9 @@ enum AnalyticsEventType {
   SIP_CALL_INCOMING = 37;
   SIP_CALL_STARTED = 38;
   SIP_CALL_ENDED = 39;
+  REPORT = 40;
```
boks1971, line 124: Adding an `AnalyticsEvent` to report different kinds of data to analytics. Open to a better name.

protobufs/livekit_analytics.proto
```diff
@@ -202,3 +204,24 @@ message AnalyticsNodeRooms {
   google.protobuf.Timestamp timestamp = 3;
   repeated AnalyticsRoom rooms = 4;
 }
+
+message ReportInfo {
```
boks1971, line 208: Container for `Report` event data. Starting off a `FeatureUsage` to report Krisp usage.

.changeset/six-humans-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
```
lukasIO, line 2: this is a bug in the changeset bot :( 

see https://github.com/changesets/bot/issues/82

```suggestion
"@livekit/protocol": patch
```

protobufs/livekit_analytics.proto
shishirng: this looks good

paulwe (approved):

protobufs/livekit_analytics.proto
```diff
@@ -202,3 +204,28 @@ message AnalyticsNodeRooms {
...
+}
+
+message TimeRange {
+  google.protobuf.Timestamp started_at = 1;
+  google.protobuf.Timestamp ended_at = 2;
+}
+
+message FeatureUsageInfo {
+  enum Feature {
+    KRISP_NOISE_CANCELLATION = 0;
```
bcherry, line 221: we're going to want to meter krisp BVC and krisp NC separately as we may have different pricing arrangements for the two models, so we probably need either two features or a separate sub-feature property

boks1971: Will add another feature enum. That's the reason why I named the first one noise cancellation explicitly instead of just calling it Krisp.

boks1971: Added in this [commit](https://github.com/livekit/protocol/pull/962/commits/003d347f8ba5ba56ab959a19fd3cbee1b3ba3535).

PR #963, opened 2025-02-05
Add Twirp options to preserve error details and client timeouts

dennwc:
This change enables the following:
- PSRPC propagation of protobuf-encoded error details.
- Twirp propagation of client timeouts.
- Twirp propagation of protobuf error details (both binary and custom metadata mappings).

There are a few changes required to make it work:
- PSRPC version must be updated on the servers/clients.
- Twirp client will need to add `DefaultClientOptions()...` to the options.
- Twirp server will need to add `DefaultServerOptions()...` and additionally call `RegisterServer` when registering Twirp servers into HTTP mux.

Some side-effects may occur when merging this. Notably, any error that implements `GRPCStatus` will be propagated through PSRPC and Twirp. Thus, error codes may change from `Unknown` to a more specific one.


biglittlebigben (approved):
There seems to be some overlap with the twirp/grpc bridging already implemented in psrpc: https://github.com/livekit/psrpc/blob/main/errors.go. Is there any plan to merge these 2 implementations?

dennwc:
@biglittlebigben There's some overlap indeed, but PSRPC implementation only does PSRPC-gRPC and PSRPC-Twirp. Going through gRPC-PSRPC-Twirp may be lossy, this is why it's not currently calling these functions.

I know that a few of our project do internal conversion between error codes of different kind. So I'd like to first merge this, then see if existing project can migrate and incorporate these changes to protocol. And then see if we can also merge it somehow with PSRPC helpers.

paulwe:
@dennwc there's a grpc middleware to make the grpc-psrpc-twirp path work here - https://github.com/livekit/cloud-protocol/blob/main/grpcutil/psrpcerror.go - we should add support for the new detailed error responses and move it to protocol. this seems preferable to using grpc status as the common error format since psrpc already supports the twirp error interface and should avoid the less tidy bits of these error helpers.

PR #977, opened 2025-02-10
fix(Outbound Sip Creation): allow null numbers

chasemcdo:


dennwc:
@chasemcdo I think you might be looking for [this](https://github.com/livekit/protocol/pull/961)? It allows making outbound calls without trunk, so that all settings can be passed directly to `CreateSIPParticipant` API.

PR #979, opened 2025-02-12
Webhook analytics event.

boks1971:
NOTE: Need a frostbyte73/core PR to merge and tag a release
PR pending: https://github.com/frostbyte73/core/pull/5


go.mod
```diff
@@ -86,3 +86,5 @@ require (
 	golang.org/x/tools v0.29.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20241202173237-19429a94021a // indirect
 )
+
+replace github.com/frostbyte73/core => ../frostbyte73/core/core
```
paulwe, line 90: should we fork these? (this and `iters`)

boks1971: core - we have @frostbyte73 to ping, so maybe can remain there?
iters - are we making changes, if up streaming is delaying things makes sense to fork

boks1971: oh, is iters from @dennwc , did not realise that, sorry

yeah, would be good to not have that dependency, forking and upstreaming without blocking makes sense.

@frostbyte73 @dennwc @davidzhao okay to fork those into livekit org?

PR #989, opened 2025-02-26
Move SIP call dispatch info into a separate type.

dennwc:


protobufs/livekit_sip.proto
```diff
@@ -671,3 +671,12 @@ enum SIPCallDirection {
   SCD_INBOUND = 1;
   SCD_OUTBOUND = 2;
 }
+
+message SIPCall {
```
biglittlebigben, line 675: How does this compare to the existing SIPCallInfo type? Can we also try to keep the fields names consistent between the different types? For instance, SIPCallInfo has a `call_id` field that, I believe` contains the same value.

biglittlebigben (commented):
Can you comment on the motivation of this change? I agree it may make for a cleaner internal RPC API, but it involves an API transition.

PR #1000, opened 2025-03-07
Per resource webhook queue.

boks1971:
With a worker based approach, an event on a resource (like a room) not being handled could affect other resources (other rooms or ingress or egress).

To prevent that, do a per resource queue with two kinds of limits
- age of event: when de-queuing, if the event queued time exceeds a certain limit (default: 5 seconds), the event is dropped
- depth of queue: drop event if the number of queued events exceeds some level (default: 200)

To reap the per resource queue, it is reaped if the queue is idle for a configurable timeout (default: 5 min).

Also, deprecating `num_dropped` from proto.

Some other bits of notes/challenges:
- Keeping the old URL notifier also. Also, the default notifier is that. Can change to the new one before merging this PR if this approach looks good.
- Storing `context.Context` in resource queue so that it can be used during posting to webhook from resource queue go routine. Storing that is not great, but we need it for analytics key.


webhook/notifier.go
```diff
@@ -71,3 +73,151 @@ func (n *DefaultNotifier) RegisterProcessedHook(hook func(ctx context.Context, w
 		u.RegisterProcessedHook(hook)
 	}
 }
+
+// ---------------------------------
```
boks1971, line 77: consolidating common functions/structures between different notifiers into this file.

webhook/notifier.go
```diff
@@ -71,3 +73,151 @@ func (n *DefaultNotifier) RegisterProcessedHook(hook func(ctx context.Context, w
...
+
+type logAdapter struct{}
+
+func (l *logAdapter) Printf(string, ...interface{}) {}
+
+// ---------------------------------
+
+func eventKey(event *livekit.WebhookEvent) (string, bool) {
```
boks1971, line 94: This is one change for existing URL notifier, returning an indication of end.

webhook/resource_queue.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/livekit/protocol/livekit"
+)
+
+var (
+	errQueueBackedUp = errors.New("queue is backed up")
+	errQueueFull     = errors.New("queue is full")
+)
+
+type item struct {
+	ctx      context.Context
```
boks1971, line 35: storing `ctx` here is not great, but need analytics key

webhook/resource_queue.go
```diff
@@ -0,0 +1,145 @@
...
+		default:
+			r.mu.Lock()
+			if r.items.Len() == 0 {
+				r.cond.Wait()
+			}
+
+			if r.closed.IsBroken() {
+				r.mu.Unlock()
+				continue
+			}
```
paulwe, line 132: this is racey and confusing because we're mixing too many synchronization primitives. `r.mu` should guard closed and drain mutation (they can be a plain bools). this can be simplified to something like
```go
func Stop() {
    r.mu.Lock()
    defer r.mu.Unlock()
    if !r.closed {
        r.closed = true
        r.drain = !force
    }
}
```
```go
for {
    r.mu.Lock()
    for {
        if r.closed{
            r.mu.Unlock()
            if r.drain { // drain is only updated once atomically with closed so it's safe to read
                r.flush()
            }
            return
        }
        if r.items.Len() != 0 {
            break
        }
        r.cond.Wait()
    }
    
    item := r.items.PopFront() // we only get here if the queue isn't empty
    r.mu.Unlock()
    // ...
}
```

paulwe (approved):

webhook/resource_queue.go
```diff
@@ -0,0 +1,145 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if r.items.Len() >= r.params.MaxDepth {
+		return errQueueFull
+	}
+
+	if r.items.Len() > 0 {
+		front := r.items.Front()
+		if time.Since(front.queuedAt) > r.params.MaxAge {
```
paulwe, line 89: we should prune expired items in the worker loop rather than rejecting new submissions

boks1971: I was debating between the two. Cool. Will move it to the processing queue.

webhook/resource_queue.go
boks1971: Thank you. Will make the change.

webhook/resource_queue.go
boks1971: Implemented changes in these commits
- https://github.com/livekit/protocol/pull/1000/commits/4b5ccabd128bc4af664de3fa86a8bddbabb66ada
- https://github.com/livekit/protocol/pull/1000/commits/ac35fa74c9c2fe9d9e52a2e45dae2f8243ca0ac9
- https://github.com/livekit/protocol/pull/1000/commits/7b248aa039ded5440acf4e1152f13f296346f24d
(sorry, had some test errors, so multiple commits).

Please have a look when you get a chance. If these look good, I will switch the default notifier to this before merging.

webhook/resource_queue.go
```diff
@@ -0,0 +1,157 @@
...
+	return nil
+}
+
+func (r *resourceQueue) flush() {
+	r.mu.Lock()
+	for r.items.Len() > 0 {
+		item := r.items.PopFront()
+		r.mu.Unlock()
+
+		if r.params.Poster != nil && item != nil {
```
paulwe, line 123: i don't see code paths where this could be true?

webhook/resource_url_notifier.go
```diff
@@ -0,0 +1,348 @@
...
+		rqi = &resourceQueueInfo{resourceQueue: rq, key: key}
+		tqi = &utils.TimeoutQueueItem[*resourceQueueInfo]{Value: rqi}
+		rqi.tqi = tqi
+		r.resourceQueues[key] = rqi
+	}
+	r.mu.Unlock()
+
+	err := rqi.resourceQueue.Enqueue(ctx, event)
+	for {
+		if err == nil || err != errQueueClosed {
```
paulwe, line 167: this condition won't be possible after fixing the queue race

webhook/resource_url_notifier.go
```diff
@@ -0,0 +1,348 @@
...
+	if r.closed.IsBroken() {
+		return errClosed
+	}
+
+	key := eventKey(event)
+
+	r.mu.Lock()
+	var tqi *utils.TimeoutQueueItem[*resourceQueueInfo]
+	rqi := r.resourceQueues[key]
+	if rqi == nil {
```
paulwe, line 153: we have to call `Reset` on the timeout queue on every access to bump the deadline. if we do it before releasing this lock the rest of the code can be simpler

we can check that the item hasn't expired like
```go
if rqi == nil || !r.resourceQueueTimeoutQueue.Reset(rqi.tqi) {
```

webhook/resource_url_notifier.go
```diff
@@ -0,0 +1,348 @@
...
+		case <-r.closed.Watch():
+			return
+
+		case <-ticker.C:
+			for it := r.resourceQueueTimeoutQueue.IterateAfter(r.params.Timeout); it.Next(); {
+				rqi := it.Item().Value
+				if !rqi.resourceQueue.StopIfIdle(r.params.Timeout) && rqi.tqi != nil {
+					r.resourceQueueTimeoutQueue.Reset(rqi.tqi)
+					continue
+				}
```
paulwe, line 338: using `IterateRemoveAfter`atomically marks the expired items deleted to sync the return value from `Reset`

if we got here the items in the queue have all either been delivered or expired so we can unconditionally stop the queue

webhook/resource_queue.go
boks1971: Yeah true, was just being paranoid. Will remove those checks.

webhook/resource_url_notifier.go
boks1971: Thank you Paul. Try 3 here in this [commit](https://github.com/livekit/protocol/pull/1000/commits/81bbfcae42288cbdf1cd5844b51c498338ad470b) with feedback incorporated.

paulwe (approved):
Thank you Raja! lgtm 🚀

boks1971:
> Thank you Raja! lgtm 🚀

Thank you Paul for your patience and great review! Going to make this the default notifier and finish up this PR. Will do a separate one for event filtering.

PR #1002, opened 2025-03-10
Add ForwardParticipant method to RoomService

cnderrauber:


protobufs/livekit_room.proto
```diff
@@ -228,3 +233,17 @@ message RoomConfiguration {
...
   repeated RoomAgentDispatch agents = 10;
 }
+
+message ForwardParticipantRequest {
+  // room to forward participant from
+  string room = 1;
+  // identity of the participant to forward
+  string identity = 2;
+  // room to forward participant to
+  string destination_room = 3;
```
boks1971, line 243: Should this be a list if a participant can be forwarded to many rooms?

boks1971 (approved):

protobufs/livekit_room.proto
cnderrauber: it can be called multiple times to forward to many rooms and I think it is better than a list for reasons:

cnderrauber: it can be called multiple times to do that. it would be better than a list since it is simple to process error or retry when the list version is partial succeed.

boks1971: Sounds good.

PR #1003, opened 2025-03-10
downgrade all turn, ice, and sdk errors

frostbyte73:


boks1971 (approved):

logger/medialogutils/cmd.go
```diff
@@ -58,8 +64,8 @@ func (l *HandlerLogger) Write(p []byte) (n int, err error) {
 		case strings.HasPrefix(line, "0:00:"):
 			// ignore cuda and template not mapped gstreamer warnings
 			continue
-		case strings.HasPrefix(line, "turnc"):
-			// warn on turnc error
+		case len(line) > 6 && downgrade[line[:6]]:
```
boks1971, line 67: Case insensitive check?

frostbyte73: case sensitive is fine, they're always the same logs

PR #1008, opened 2025-03-12
ForwardParticipant changes

cnderrauber:
remove stop flag in the request
add new `forwarded` participant type


boks1971 (approved):

davidzhao (commented):
@cnderrauber do you mind adding changesets here so we can generate the appropriate changelog? this should be a minor version

PR #1019, opened 2025-03-24
Add KindDetails to ParticipantInfo

lukasIO:
also removes the FORWARDED and CLOUD_AGENT from Participant.Kind

please suggest a better name instead of `CloudFeature`


protobufs/livekit_models.proto
```diff
@@ -137,10 +137,10 @@ message ParticipantInfo {
     SIP = 3;
     // LiveKit agents
     AGENT = 4;
-    // forwarded participant from another room
-    FORWARDED = 5;
-    // Cloud agents
-    CLOUD_AGENT = 6;
+  }
+  enum CloudFeature {
```
boks1971, line 143: My 2c: keep it specific and not a wide net like `CloudFeature`. I would call it `KindMetadata` (or something like that) to indicate that this is metadata related to `ParticipantKind`. There may be other features down the line which can be their own enum.

In this particular case, these are are mutually exclusive, so having `repeated` may not be necessary, but as a general practice, I am thinking leaving it as `repeated` is good as we may have more `kind` related metadata down the line.

boks1971 (approved):
LGTM, but want to get 👀 from @cnderrauber and @real-danm to make sure that these changes are handled appropriately in related places.

real-danm (approved):
this looks good to me, thanks for fixing @lukasIO

protobufs/livekit_models.proto
lukasIO: Yeah, I think that makes sense. 
Are they mutually exclusive? 
So a cloud_agent could not be forwarded?

boks1971: My bad, it can be. So, not mutually exclusive.

cnderrauber (approved):
looks good!

protobufs/livekit_models.proto
lukasIO: @boks1971 updated the name to `KindDetails` which isn't as overloaded in meaning with the rest of our `metadata` stuff. let me know what you think

PR #1022, opened 2025-03-26
Refine backup codec policy

cnderrauber:
Make the default behavior to be a `preferred` value that the sfu will determine proper policy by room information.

### DIFF
--- a/protobufs/livekit_models.proto
+++ b/protobufs/livekit_models.proto
@@ -78,10 +78,13 @@ enum ImageCodec {
 
 // Policy for publisher to handle subscribers that are unable to support the primary codec of a track
 enum BackupCodecPolicy {
-  // default behavior, regress to backup codec and all subscribers will receive the backup codec
+  // default behavior, the track prefer to regress to backup codec and all subscribers will receive the backup codec,
+  // the sfu will try to regress codec if possible but not assured.
   REGRESSION = 0;
   // encoding/send the primary and backup codec simultaneously
   SIMULCAST = 1;
+  // force the track to regress to backup codec, this option can be used in video conference or the publisher has limited bandwidth/encoding power
+  FORCE_REGRESSION = 2;
 }
 
 message PlayoutDelay {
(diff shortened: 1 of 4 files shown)


protobufs/livekit_models.proto
```diff
@@ -78,10 +78,13 @@ enum ImageCodec {
 
 // Policy for publisher to handle subscribers that are unable to support the primary codec of a track
 enum BackupCodecPolicy {
-  // default behavior, regress to backup codec and all subscribers will receive the backup codec
+  // default behavior, the track prefer to regress to backup codec and all subscribers will receive the backup codec,
+  // the sfu will try to regress codec if possible but not assured.
   REGRESSION = 0;
```
boks1971, line 83: maybe rename this to `PREFER_REGRESSION` or `AUTO_REGRESSION` and change the `FORCE_REGRESSION` to just `REGRESSION`. When not using `FORCE_` for `SIMULCAST`, but using for `REGRESSION` could be confusing?

cnderrauber: Thought about this but concerning renaming the existed field will break sdks (maybe js only) when they update the protocol version.. is it ok to do that?

boks1971: Good point, I think it should be okay, but it might require an app change for devs.

cc: @lukasIO for opinion on that.

davidzhao: it's totally fine to rename enums. protobuf values are sent as ints in the wire protocol.

only Twirp/webhooks uses JSON serialization, and this value isn't used anywhere there

paulwe (approved):

protobufs/livekit_models.proto
paulwe: we would avoid this if we changed this to `PREFER_REGRESSION: 0, REGRESSION: 2` - legacy clients will still get the same behavior most of the time... no one is setting this yet so it doesn't really matter. and when they update it will change to force regression mode silently when their client starts sending 2

boks1971: IIUC, I think @cnderrauber is asking about apps using the enum name in publish settings and not necessarily the wire protocol.

But like @paulwe , if somebody had set it to `REGRESSION` enum by name, they wanted to force a regression right? So, it will now become protobuf value 2 which still has the intended behaviour of forcing regression. So, renaming sounds good.

PR #1030, opened 2025-04-02
Add rate for memory use and total.

boks1971:
Load and overage for memory is calculated differently in cloud.
Removing that distinction caused memory based load to show higher load
consistently which was unintentional.


paulwe (approved):

protobufs/livekit_internal.proto
```diff
@@ -151,6 +151,8 @@ message NodeStatsRate {
   // time weighted averages across stats windows forming part of a rate measurement interval
   float cpu_load = 20;
   float memory_load = 21;
+  float memory_used = 22;
+  float memory_total = 23;
 
   // NEXT ID: 22
```
paulwe, line 157: 24?

boks1971: Good catch, sorry, still waling up :-)

PR #1031, opened 2025-04-03
Use cgroup for memory stats.

boks1971:
This is not going to work well when limit is not set. It falls back to go-osstat if limits are not set.


paulwe (approved):
> This is not going to work well when limit is not set. It falls back to go-osstat if limits are not set.

we only set requests... i'm not sure this will work

boks1971:
> > This is not going to work well when limit is not set. It falls back to go-osstat if limits are not set.
> 
> we only set requests... i'm not sure this will work

yeah, this is probably not going to work at all till we set limits, but should not be any different to what we have currently. Will wait for more opinions to merge this.

utils/hwstats/memory_linux.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+const (
+	memStatsPathV1 = "/sys/fs/memory/memory.usage_in_bytes"
+	memStatsPathV2 = "/sys/fs/memory/memory.current"
+
+	memUsagePathV1 = "/sys/fs/cgroup/memory/memory.usage_in_bytes"
+	memLimitPathV1 = "/sys/fs/cgroup/memory/memory.limit_in_bytes"
+
+	memCurrentPathV2 = "sys/fs/cgroup/memory/memory.current"
```
davidzhao, line 33: should this have a leading `/` as well?

davidzhao (approved):

utils/hwstats/memory_linux.go
boks1971: Yes, missed it. Thank you for catching.

PR #1035, opened 2025-04-05
allow calling promise.Resolve more than once

paulwe:


.changeset/blue-birds-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
```
boks1971, line 2: can you change this to
```
"github.com/livekit/protocol": patch
"@livekit/protocol": patch
```

boks1971 (approved):

PR #1039, opened 2025-04-08
Helpers to deal with SDP fragments.

boks1971:
WHIP uses SDP fragments which fail parsing using full-fledged SDP libraries like pion/sdp, go-sdp.

Add utlities to deal with SDP fragments.


sdp/sdp.go
```diff
@@ -191,3 +193,512 @@ func CodecsFromMediaDescription(m *sdp.MediaDescription) (out []sdp.Codec, err e
 
 	return out, nil
 }
+
+func GetBundleMid(parsed *sdp.SessionDescription) (string, bool) {
```
paulwe, line 197: there's a lot of new code here, would any of it benefit from tests?

paulwe (approved):

sdp/sdp.go
boks1971: Yes, will write them next.

PR #1050, opened 2025-04-17
fix changeset

boks1971:


anunaym14 (approved):
Should there be an workflow for this? To check which files are modified and check changeset against them?

boks1971:
> Should there be an workflow for this? To check which files are modified and check changeset against them?

The changeset generator does get the commit note and add it. But, setting the package name is manual. I am guessing there are more automated ways, but I have not checked it. It has been manual so far.

PR #1051, opened 2025-04-17
feat: MoveParticipant API

davidzhao:
This new API would move a participant from one cloud room to another. It will support the following use cases:
- warm transfer in voice AI
- fast switching between different live streams
- prejoin room in meetings


cnderrauber (approved):

paulwe:
could we do this with subscription policies instead? this api breaks load balancing and analytics

protobufs/livekit_room.proto
```diff
@@ -61,10 +61,17 @@ service RoomService {
...
   rpc UpdateRoomMetadata (UpdateRoomMetadataRequest) returns (Room);
 
-  // Forward a participant's track(s) to another room. Requires `roomAdmin` and `destinationRoom`. The forwarding will 
+  // Cloud-only
+  // a connected participant's track(s) to another room. Requires `roomAdmin` and `destinationRoom`. The forwarding will 
   // stop when the participant leaves the room or `RemoveParticipant` has been called in the destination room. 
   // A participant can be forwarded to multiple rooms. The destination room will be created if it does not exist.
   rpc ForwardParticipant(ForwardParticipantRequest) returns (ForwardParticipantResponse);
+
+  // Cloud-only
```
dennwc, line 70: Why it's cloud-only, though? It should be even easier to implement in OSS.

dennwc (approved):
That would be really useful for SIP. LGTM!

protobufs/livekit_room.proto
davidzhao: rooms are not guaranteed to exist on the same server participant is connected to. this requires the mesh

dennwc: Makes sense to have it for single server setups at least.

davidzhao: would not want to have features that only work under specific configuration & deployment. this would be a nightmare to support.

PR #1052, opened 2025-04-18
Add IMAGE_SUFFIX_NONE_OVERWRITE

biglittlebigben:
Add a ImageFileSuffix enum value to disable appending any image suffix and overwrite the generated image each time a new one is uploaded instead. Overwriting the image seems exclusive to having any suffix (particularly a timestamp suffix), so this seemed the best place to add it. `OVERWRITE` was added as part of the enum name to make it clear that data loss will occur with this option.


davidzhao (approved):
lgtm

PR #1053, opened 2025-04-24
allow rewriting deferred logger values

paulwe:


cnderrauber (approved):
Great, thanks for the work!!

PR #1056, opened 2025-04-30
Add support for spatial video metadata

ladvoc:
To allow clients to properly handle spatial video (stereoscopic and/or spherical), this PR proposes adding a subset of the fields defined by the [Spherical Video V2 RFC](https://github.com/google/spatial-media/blob/master/docs/spherical-video-v2-rfc.md) to `TrackInfo`.

### DIFF
--- a/protobufs/livekit_models.proto
+++ b/protobufs/livekit_models.proto
@@ -229,6 +229,8 @@ message TrackInfo {
   TimedVersion version = 18;
   repeated AudioTrackFeature audio_features = 19;
   BackupCodecPolicy backup_codec_policy = 20;
+  // metadata for stereoscopic and spherical video features
+  SpatialVideoFeatures spatial_video_features = 21;
 }
 
 enum VideoQuality {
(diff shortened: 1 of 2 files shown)


protobufs/livekit_models.proto
```diff
@@ -249,6 +251,43 @@ message VideoLayer {
   uint32 ssrc = 5;
 }
 
+// Metadata for stereoscopic and spherical video features. Includes
+// a subset of the fields defined by the Google Spherical Video V2 RFC.
+message SpatialVideoFeatures {
```
lukasIO, line 256: maybe let's go for a more general`VideoTrackFeatures` ?

lukasIO (commented):
I think in theory for now we could comment out everything apart from `stereo_mode` ? 

All fields make sense to me, I'm just not sure if defining all of them on the protocol level is the right choice.

Would love to hear some additional opinions from the rest of the team.

bcherry:
I think it's a great idea to add something like this to the track.  I think we should only add explicit detailed support for things we're actually planning to support at a framework level (e.g. the sdks + components libraries have special rendering support). i'm definitely in favor of a basic enum for very well understood formats that don't require extra metadata (e.g. stereo l/r) but the others i'm not so sure although i'm not an expert. I think that customers implementing more complex formats are also probably comfortable implementing their own rendering too, at least for the medium term. and i wouldn't want us to add a complex field that no one really uses because it's not quite comprehensive enough. 

it would be great if track had a freeform attributes field or something but for now we put that on participant and i think its probably fine for a developer to use that to share detailed track  rendering information across clients if needed.

wdyt @chenosaurus?

chenosaurus:
> I think it's a great idea to add something like this to the track. I think we should only add explicit detailed support for things we're actually planning to support at a framework level (e.g. the sdks + components libraries have special rendering support). i'm definitely in favor of a basic enum for very well understood formats that don't require extra metadata (e.g. stereo l/r) but the others i'm not so sure although i'm not an expert. I think that customers implementing more complex formats are also probably comfortable implementing their own rendering too, at least for the medium term. and i wouldn't want us to add a complex field that no one really uses because it's not quite comprehensive enough.
> 
> it would be great if track had a freeform attributes field or something but for now we put that on participant and i think its probably fine for a developer to use that to share detailed track rendering information across clients if needed.
> 
> wdyt @chenosaurus?

I agree,  let's have only `stereo_mode` to start and add more of the parameters when we have a need for it.

PR #1063, opened 2025-05-05
Add SIPTransferInfo to analytics

biglittlebigben:


shishirng (approved):

protobufs/livekit_sip.proto
```diff
@@ -711,6 +711,17 @@ message SIPCallInfo {
   // NEXT ID: 25
 }
 
+message SIPTransferInfo {
+  string call_id = 1;
+  uint32 refer_seq_num = 2;
+  int64 transfer_initiated_at_ns = 3;
+  int64 transfer_completed_at_ns = 4;
```
dennwc, line 718: Maybe lets use the `google.protobuf.Timestamp` for the new time fields?

dennwc (approved):

protobufs/livekit_sip.proto
biglittlebigben: No strong opinion from me here: consistency vs better type. @shishirng, what do you think?

PR #1066, opened 2025-05-08
add KindDetail tests and required funcs

real-danm:
allows participant tracking for cloud agents.

### DIFF
--- a/auth/grants.go
+++ b/auth/grants.go
@@ -49,12 +49,13 @@ func (c *RoomConfiguration) UnmarshalJSON(data []byte) error {
 }
 
 type ClaimGrants struct {
-	Identity string      `json:"-"`
-	Name     string      `json:"name,omitempty"`
-	Kind     string      `json:"kind,omitempty"`
-	Video    *VideoGrant `json:"video,omitempty"`
-	SIP      *SIPGrant   `json:"sip,omitempty"`
-	Agent    *AgentGrant `json:"agent,omitempty"`
+	Identity   string      `json:"-"`
+	Name       string      `json:"name,omitempty"`
+	Kind       string      `json:"kind,omitempty"`
+	KindDetail string      `json:"kindDetail,omitempty"`
+	Video      *VideoGrant `json:"video,omitempty"`
+	SIP        *SIPGrant   `json:"sip,omitempty"`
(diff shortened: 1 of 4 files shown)


auth/grants.go
```diff
@@ -49,12 +49,13 @@ func (c *RoomConfiguration) UnmarshalJSON(data []byte) error {
...
-	Identity string      `json:"-"`
-	Name     string      `json:"name,omitempty"`
-	Kind     string      `json:"kind,omitempty"`
-	Video    *VideoGrant `json:"video,omitempty"`
-	SIP      *SIPGrant   `json:"sip,omitempty"`
-	Agent    *AgentGrant `json:"agent,omitempty"`
+	Identity   string      `json:"-"`
+	Name       string      `json:"name,omitempty"`
+	Kind       string      `json:"kind,omitempty"`
+	KindDetail string      `json:"kindDetail,omitempty"`
```
boks1971, line 55: don't remember another example, do we use camel case for JSON or snake_case?

boks1971 (approved):
A small comment in line.

Other than that, thinking aloud if this should be in the token, but I cannot think of a better place to put this especially for cloud agent use case. Makes sense to put it in the token.

auth/grants.go
paulwe: should this be a repeated field? we might have forwarded hosted agent participants

boks1971: I had the same comment and then the redacted comment it as it felt like a token could be only one. 

ParticipantInfo does have it as repeated, but that is added by server.

I could not convince myself that this should be repeated in the token, but I agree it is probably better to make it repeated to keep the flexibility.

davidzhao: i was discussing with dan.. I think we should take hosted attributes out of this.. that shouldn't be a settable field by the end user.

boks1971: > Other than that, thinking aloud if this should be in the token, but I cannot think of a better place to put this especially for cloud agent use case. Makes sense to put it in the token.

Had the same question too. Where would this get set then? I was not sure where that could happen.

boks1971: Guess, it can be set in the `ParticipantInfo` by the server side. Guess it can be detected using other mechanisms by the server.

PR #1073, opened 2025-05-14
Deffered logger control interface

boks1971:


paulwe (approved):
this should work - could you write a test to confirm, please?

logger/logger.go
```diff
@@ -483,7 +483,7 @@ func (l LogRLogger) WithoutSampler() Logger {
 }
 
 func (l LogRLogger) WithDeferredValues() (Logger, DeferredFieldResolver) {
-	return l, func(args ...any) {}
+	return l, nil
```
paulwe, line 486: should we return a noop interface here to avoid the nil pointer?

boks1971: I was thinking about it too. Will change it.

Also, do we need to create a new `core` logger on reset? I think if you zap logger does not replace fields IIRC. So, after `Reset()`, if we add different fields, the old fields will still be there. If we add same fields, I think the zap logger does duplicate fields?

paulwe: deferrer doesn't modify. the core so the way this is implemented works. your solution is the same one i was going to implement after lunch

boks1971: Updated. Also, added a test. Please re-review when you get a chance. Want to make sure I am testing the right thing.

boks1971: For test, I added two more fields to the `testLogger` (`C` and `D`) and set them after `Reset`. But, then decided to change the existing value. Maybe, should also add new values to the test in addition to modifying `A` and `B` to make the test better?

paulwe: the test looks good. this is ready to merge

PR #1075, opened 2025-05-15
add agent jwt helper

paulwe:


agent/token.go
```diff
@@ -35,3 +41,79 @@ func BuildAgentToken(
...
+	}
+
+	return &WorkerTokenProvider{
+		nodeID:  nodeID,
+		keys:    keys,
+		timeout: config.Timeout,
+	}
+}
+
+func (t *WorkerTokenProvider) Encode(claims WorkerClaims) (string, error) {
```
boks1971, line 77: don't understand the multiple keys use case, but is there a need to have an API with a key index other than 0 to encode with?

agent/token_test.go
```diff
@@ -0,0 +1,60 @@
...
+				WorkerTokenConfig{
+					Secret:  c.keys1,
+					Timeout: time.Hour,
+				},
+			)
+
+			claims, err := wt1.Decode(token)
+			if c.success {
+				require.NoError(t, err)
+				require.Equal(t, workerID, claims.Subject)
```
boks1971, line 54: Worth checking `Issuer`? I guess that is testing `jose` actually, so probably not needed 🙂

boks1971 (approved):

agent/token.go
paulwe: it allows key rotation

PR #1094, opened 2025-06-04
Signalling v2 draft proposal

boks1971:
**UPDATE:** The notes below the line are the original notes. The shape of this remains the same, but reducing scope of this PR to make smaller PRs and also collaboratively design more messages as we go along. Some changes from the below description
1. Reducing messages to only a few types. Will add messages one or a few at a time as we implement more. Also, want to collaboratively/interactively design messages in smaller chunks.
2. `ConnectRequest` has been pared down from the start of this PR. Mainly, do not have the publisher side. The idea is that the first message can contain tracks to publish and SDP offer of publisher peer connection. Not handling it here as there is a need for an RPC style message for track publish (i. e. client gets feedback about the publish request). So, need to discuss how to design it.
3. Not doing chunked response yet. We may not need to do this.

@reviewers - please share opinion about naming (like I mentioned below, I am not a fan of `v2` naming) also along with any other feedback.

Will leave more remarks inline.

-------------------------------------------------------

A few bits of notes before you review
1. Will be reviewing all messages and cleaning up/re-defining as necessary. For now, have defined a few new messages and included a bunch of existing messages.
5. Don't really like the name `Signalv2` in the message naming. Please let me know if we can do something better. One option, I was thinking about was just to include a `version` field in message, but I could not settle on what level it should be added at (i. e. should it be added to the top level message which has `OneOf` OR each message).

The connect sequence would be
1. Client sends `ConnectRequest` message in `HTTP POST` body to start the connection process. JWT is the `Bearer` authorisation in the `POST`. The request would be to `https://<connection-url>/rtc/v2`.
2. Client should use a  long timeout (something like 30s) on this request.
6. Server responds with `ConnectResponse`. This will have all ICE candidates from server which will allow clients to start the ICE connectivity process.
7. Server will send chunked response. This is done so that server can keep sending events (like new participant joining, track publish, etc.) using the same request.
8. Client can continue to send further messages to server using `HTTP PATCH` to `https://<connection-url>/rtc/v2/{participant_sid}`. `participant_sid` is retrieved from `ParticipantInfo` in `ConnectResponse`. Server will continue to use the initial connection and send chunked responses to messages in the `PATCH`.
9. Client can send `HTTP DELETE` to `https://<connection-url>/rtc/v2/{participant_sid}` in case the client has to close participant session before data channel is established. Once data channel is established, client can use either `HTTP DELETE` or `LeaveRequest` via data channel (NOTE TO SELF: probably should just standardise on one way to end a session, i. e. `HTTP DELETE` which works for both cases).
10. Client + Server switch to data channel for signalling once the peer connections are established and data channel opened.
11. Client should open a reliable/ordered data channel on primary peer connection dedicated for signalling message use only. Client should send a `ReplayRequest` via the newly opened data channel. Server will replay messages starting from the message id after the `last_processed_remote_message_id` message in `ReplayRequest`.
12. Either side can send `Acknowledgement` at any time to (i) indicate last processed remote message id and (ii) request replay from a certain message. This is useful for cases like client detecting temporary disconnection of peer connection. If the receiving side detects a gap in the received sequence, it can send an `Acknowledgement` with last processed so that the sending side will replay from that point.  (NOTE TO SELF: need to think of a mechanism to ensure that this does not get stuck permanently)
13. Client sends `ReconnectRequest` to resume a session (the session could be resumed on a different server in `cloud`, i. e. migrated). Server responds with `ReconnectResponse`. (NOTE: need to enhance this to include publisher SDP answer and subscriber SDP offer to cover moving to new server)


protobufs/livekit_rtc_v2.proto
```diff
@@ -0,0 +1,156 @@
...
+  bool disable_red = 6;
+}
+
+message VideoTrack {
+  TrackSettings track_settings = 1;
+  repeated SimulcastCodec simulcast_codecs = 2;
+  BackupCodecPolicy backup_codec_policy = 3;
+}
+
+message ConnectRequest {
```
boks1971, line 70: This is the first message from client -> server to connect to a room.

It should be an HTTP POST with JWT as Bearer token. And this message as POST body.

Server will `ConnectResponse` for this.

protobufs/livekit_rtc_v2.proto
```diff
@@ -0,0 +1,156 @@
...
+  TrackSettings track_settings = 1;
+  repeated AudioTrackFeature audio_features = 5;
+  bool disable_red = 6;
+}
+
+message VideoTrack {
+  TrackSettings track_settings = 1;
+  repeated SimulcastCodec simulcast_codecs = 2;
+  BackupCodecPolicy backup_codec_policy = 3;
+}
```
boks1971: Making specific messages for audio track and video track.

protobufs/rpc/signal_v2.proto
```diff
@@ -0,0 +1,47 @@
...
+
+package rpc;
+
+option go_package = "github.com/livekit/protocol/rpc";
+
+import "options.proto";
+import "livekit_internal.proto";
+import "livekit_rtc_v2.proto";
+
+service Signalv2 {
```
boks1971: This is the internal protocol between controller & media. Very similar to `signal.proto`, but for v2 signalling.

protobufs/livekit_rtc_v2.proto
```diff
@@ -0,0 +1,156 @@
...
+    Ping ping = 17;
+    UpdateLocalAudioTrack update_local_audio_track = 18;
+    UpdateLocalVideoTrack update_local_video_track = 19;
+  }
+}
+
+message Signalv2ClientEnvelope {
+  // collection of client -> server messages
+  repeated Signalv2ClientMessage client_messages = 1;
+}
```
boks1971, line 42: Making an envelope type to send a collection of messages. Useful for replaying messages from a certain point in time.

protobufs/livekit_rtc_v2.proto
```diff
@@ -0,0 +1,156 @@
...
+
+import "livekit_rtc.proto";
+import "livekit_models.proto";
+
+message Signalv2ClientMessage {
+  // monotonically increasing number starting at some non-zero number (should not rollover)
+  uint32 message_id = 1;
+
+  // last remote message processed
+  uint32 last_processed_remote_message_id = 2;
```
boks1971, line 30: Introducing these two fields for all messages to indicate sender side message sequence and what remote message was last processed.

protobufs/livekit_rtc_v2.proto
```diff
@@ -0,0 +1,156 @@
...
+}
+
+message Signalv2ServerMessage {
+  // monotonically increasing number starting at some non-zero number (should not rollover)
+  uint32 message_id = 1;
+
+  // last remote message processed
+  uint32 last_processed_remote_message_id = 2;
+
+  repeated ErrorResponse errors = 3;
```
boks1971: Introducing this to send errors in processing specific messages from client. Informational only (maybe we can make it actionable down the line). Thinking about this to give clients a better view of any errors on server side.

NOTE: Individual messages can include `request_id` to get a response (irrespective of an error). That mechanism will continue to be supported. This is a generic mechanism to just send down errors processing any message.

protobufs/livekit_rtc_v2.proto
```diff
@@ -0,0 +1,157 @@
...
+    SessionDescription subscriber_sdp = 8;
+    TrickleRequest ice_candidate = 9;
+    MuteTrackRequest mute_track_request = 10;
+    UpdateSubscription update_subscription = 11;
+    UpdateTrackSettings update_track_settings = 12;
+    LeaveRequest leave_request = 13;
+    SubscriptionPermission update_subscription_permission = 14;
+    SimulateScenario simulate_scenario = 15;
+    UpdateParticipantMetadata update_participant_metadata = 16;
+    Ping ping = 17;
```
cnderrauber: Is the `ping/pong` required in the v2 protocol? It is used by the client to detect signal disconnect but we can rely on the ice connection state now

boks1971: Good point @cnderrauber . I forgot to remove it. It should not be needed.

auth/grants.go
```diff
@@ -162,7 +162,7 @@ func checkOutputForCredentials(output any) error {
 }
 
 type ClaimGrants struct {
-	Identity string      `json:"-"`
```
boks1971, line 165: Could not find reason for this. But, exporting this to JSON and using `Identity` and `Room` name from the grants itself rather than passing them around separately.

protobufs/livekit_internal.proto
```diff
@@ -164,9 +164,9 @@ message StartSession {
   // if a client is reconnecting (i.e. resume instead of restart)
   bool reconnect = 4;
   bool auto_subscribe = 9;
-  bool hidden = 10;
+  bool hidden = 10 [deprecated=true];
   ClientInfo client = 11;
-  bool recorder = 12;
```
boks1971, line 169: marking some of these unused stuff as I find them.

protobufs/livekit_models.proto
```diff
@@ -207,28 +207,29 @@ message TrackInfo {
...
+  bool disable_dtx = 8 [deprecated=true];   // deprecated in favor of audio_features
   // source of media
   TrackSource source = 9;
-  repeated VideoLayer layers = 10;
+  repeated VideoLayer layers = 10 [deprecated=true]; // see `codecs` for layers of individual codec
   // mime type of codec
   string mime_type = 11;
   string mid = 12;
   repeated SimulcastCodecInfo codecs = 13;
-  bool stereo = 14;
```
boks1971, line 224: same here, marking unused or deprecated stuff.

protobufs/livekit_models.proto
```diff
@@ -240,6 +241,13 @@ enum VideoQuality {
 
 // provide information about available spatial layers
 message VideoLayer {
+  enum Mode {
+    UNUSED = 0;
+    STANDARD = 1;
+    SIMULCAST = 2;
+    SVC = 3;
+  }
```
boks1971, line 249: one of the things we want to do where clients explicitly say SIMULCAST vs SVC. Also, starting to use the pattern of marking the default value as `UNUSED` to force explicit use.

protobufs/livekit_models.proto
```diff
@@ -715,3 +723,13 @@ message WebhookConfig {
...
+
+message Fragment {
+  uint32 packet_id = 1;  // all fragments of a packet will have the same id
+  uint32 fragment_number = 2; // should start at 1
+  uint32 num_fragments = 3;
+  uint32 fragment_size = 4;
+  uint32 total_size = 5;
+  bytes data = 6;
+}
+
```
boks1971, line 735: Fragmentation that could be needed when a message is too big for data channel limits. This might look like a bit of an overkill, but trying to make it robust.

To explain a bit
- `fragment_size` should match `len(data)`, but `fragment_size` is explicit field just in case the data is cut off. protobuf decode should fail in that case, but made is self-checking.
- `num_fragments` will be the same in all fragments, but that allows allocating a fragment array upon receiving any fragment and populating the correct slot in the re-assembled message.
- The above also enables receiving fragments out-of-order (although the usage is meant to be over reliable/ordered data channels, made the design such that out-of-order can be supported).
- `total_size` is same is all fragments, but that is another self-check mechanism after all the fragments have been assembled to check if the size matches the sender's full packet size.

protobufs/livekit_rtc_v2.proto
```diff
@@ -0,0 +1,95 @@
...
+
+  // last remote message processed
+  uint32 last_processed_remote_message_id = 2;
+
+  repeated ErrorResponse errors = 3;
+
+  oneof message {
+    Fragment fragment = 4;
+    Signalv2ServerEnvelope envelope = 5;
+    ConnectResponse connect_response = 6;
```
boks1971: - `ConnectResponse` will be a standalone message. This will not have `messageId`.
- After that, typical operation would be,
    o sending side groups all messages to send into envelope
    o if size of the envelope needs fragmentation, fragment it and send individual fragments.
     o If not, send envelope as is.
     o If fragmented, receiver should re-assemble and unmarshal an envelope type and handle the messages inside the envelope.

protobufs/livekit_rtc_v2.proto
boks1971: Having second thoughts about this. Not sure if this would be useful, i. e. sending back error description for specific messages (keyed by message id).

I am on the wall between deciding to give it a slot as a common entity and not using it  OR removing it now and later adding a common entity with a slot number that is high. I think I am overthinking this. Leaning towards removing it and add it later if needed.

Please let me know if you have thoughts.

protobufs/livekit_models.proto
lukasIO: just for clarification what message type does a fragments' data field represent?
In clients we'll have to parse the data to that message type and if it's ambigious (e.g. could be ConnectResponse or Signalv2ServerEnvelope) I'm not sure how we'd do this cleanly

protobufs/livekit_rtc_v2.proto
lukasIO: I'm not a fan of this field. 
We wanted to make rpc style requests more explicit anyways and we, also, have the envelope now that allows multiple responses being sent down, couldn't we just have the error as part of the oneof message of `Signalv2ServerMessage` then?

protobufs/livekit_rtc_v2.proto
lukasIO: what's the main incentive of having ConnectResponse omit the `messageId` ? 
Sounds to me like we would make it more or less transport specific that way by requiring the req/resp matching to be handled on the transport level (i.e. http req in our case)

protobufs/livekit_rtc_v2.proto
```diff
@@ -0,0 +1,95 @@
...
+  bool auto_subscribe = 1;
+  bool adaptive_stream = 2;
+  optional bool subscriber_allow_pause = 3;
+  bool disable_ice_lite = 4;;
+}
+
+message ConnectRequest {
+  ClientInfo client_info = 1;
+  ConnectionSettings connection_settings = 2;
+  map<string, string> participant_attributes = 3;
```
lukasIO: are we consciously deciding to allow the connection request to only include attributes and not metadata? 

How would we handle a connection request that tries to set attributes but the permission is missing? It would be weird to deny the connection, but it would still be good to inform clients about that part of the request not being successful.

protobufs/livekit_rtc_v2.proto
boks1971: Sounds good. We can add it as `oneOf` later. Phew, that clears my mind 🙂 Thank you. Will get rid of it.

protobufs/livekit_models.proto
boks1971: It will always be `Signalv2ServerEnvelope`. `ConnectResponse` is always over HTTP response, so there is no need for fragmentation.

lukasIO: ok, that makes sense, that also somewhat answers my concern [here](https://github.com/livekit/protocol/pull/1094#discussion_r2209711891) as long as we're happy with tying the connect request + response to a specific transport (http)

protobufs/livekit_rtc_v2.proto
boks1971: This is an implementation challenge on server side. I need to spend more time trying to make this better.

`ConnectResponse` is synchronous now. Actually, if we keep doing HTTP till data channel forms, those responses will be synchronous too.

The way I have it on server is to cache responses. Asynchronously, whatever server needs to send, I am thinking of give it to the caching layer which will stamp the message id and then send it out. If the client sends a `replay` message from a specific sequence number, it will pull out messages from cache, put it in envelope and send it all.

`ConnectResponse` is a special one as there are more pieces of server code is involved and the sync/async is a bit messier. And also, I want `ConnectRequest` to get `ConnectResponse` as response (or at least has to be the first message). I am apprehensive of putting it into cache and pulling it as other messages may get ahead. 

I will think about this more.

protobufs/livekit_models.proto
boks1971: Just replied to that message. I would like to make this as transport agnostic as possible. But, some implementation details made it challenging. But, as I write more code and think more about this and as more scenarios get addressed, will look at adding message id to `ConnectResponse` also.

protobufs/livekit_rtc_v2.proto
boks1971: This is a bit of carrying over existing stuff. `Attributes` were specifically added to handle some SIP cases so that the webhook has all these IIRC. So, I just copied it. Good idea to add `Metadata` too. Will look at adding it.

Also, if there is no permission, it will get a `401 Unauthorized` response. So, yes, there will be an error response if there is no permissions to update own metadata.

lukasIO: oh, and also, how would we handle conflicting entries in attributes that are stored in the token vs in the request here?

when you say clients would get a 401 response in case `canUpdateOwnMetadata` is false, would that in turn mean no connection can get established?
Would that differ from a user's token not being valid?

boks1971: Adding comments to code now. This will add keys to what exists in the token. This will not delete keys from token even there is a key collision and this one sets the key to empty value.

Going to do the same with `metadata`. Metadata provided via this request will be used only if the one from the token is not set/empty.

Regarding errors, it is mapped to HTTP error code. We will send description of error as JSON in the body and we can make it more descriptive. We should design those error descriptions so that it is clear where exactly things are failing.

boks1971: This [commit ](https://github.com/livekit/protocol/pull/1094/commits/e27eb0acd7330b9aae2cc04e289ccc5a325ae4bd)has the changes. Please let me know if this looks okay.

lukasIO: thanks for the clarification, looks good to me!

boks1971: I read the code wrong. It actually replaces key value if there is a key collision and the value is not empty value. Sorry, will update the comment.

In that light, I am thinking if `metadata` is provided here, it should replace the one in the token? What do you think?

lukasIO: I think consistency is good here, so mirroring what we do for attributes makes sense to me

boks1971: Sounds good. Will update comments and commit. Will also change metadata overwrite behaviour in server.

boks1971: Corrected comments in this [commit](https://github.com/livekit/protocol/pull/1094/commits/7ad507f811c9b1451ca8d9cc16ca27edfa1c85b2).

lukasIO:
could you add 

```
export * from "./gen/livekit_rtc_v2_pb.js";
```

in the two `.js` and `d.ts` files in `packages/javascript/src`. That way the JS package will get the stuff from the new file as exports

boks1971:
> could you add
> 
> ```
> export * from "./gen/livekit_rtc_v2_pb.js";
> ```
> 
> in the two `.js` and `d.ts` files in `packages/javascript/src`. That way the JS package will get the stuff from the new file as exports

Will do.

boks1971:
> could you add
> 
> ```
> export * from "./gen/livekit_rtc_v2_pb.js";
> ```
> 
> in the two `.js` and `d.ts` files in `packages/javascript/src`. That way the JS package will get the stuff from the new file as exports

Done in this [commit](https://github.com/livekit/protocol/pull/1094/commits/17287b15e6f58fd74297ad2070ee56ff60373467).

PR #1095, opened 2025-06-04
Add prom metrics for webhook dispatch.

boks1971:
1. success/failure/drop counter.
2. queue length histogram.
   - doing this only for resource_url_notifier which does a per room queue as queue length is easily available. This is the one currently in use.
   - worker pool based notifier will need a change in `core` package to return a queue length when enqueuing.

### DIFF
--- a/webhook/stats.go
+++ b/webhook/stats.go
@@ -18,6 +18,7 @@ import (
 	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


webhook/stats.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func initWebhookStats(constLabels prometheus.Labels) {
+	promWebhookDispatchTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   "livekit",
+		Subsystem:   "webhook",
+		Name:        "dispatch_total",
+		ConstLabels: constLabels,
+	}, []string{"status", "reason"})
+	prometheus.MustRegister(promWebhookDispatchTotal)
```
paulwe: `promauto.New*` creates and registers metrics with the default registry in the same function

boks1971: oh nice! will do

PR #1097, opened 2025-06-07
ensure access token do not contain sensitive credentials

davidzhao:


boks1971 (approved):

paulwe (approved):

auth/grants.go
```diff
@@ -48,6 +51,116 @@ func (c *RoomConfiguration) UnmarshalJSON(data []byte) error {
...
+// egress service.
+func (c *RoomConfiguration) CheckCredentials() error {
+	if c.Egress == nil {
+		return nil
+	}
+
+	if c.Egress.Participant != nil {
+		for _, output := range c.Egress.Participant.FileOutputs {
+			if err := checkOutputForCredentials(output.Output); err != nil {
+				return err
```
anunaym14, line 68: https://pkg.go.dev/slices#ContainsFunc

Cleaner to use this instead of the for loop

davidzhao: I'm not seeing that.. I need to capture the error, not just checking for if it passes a boolean check

anunaym14: Ah yes good point, let me try to come up with something

anunaym14: https://github.com/livekit/protocol/commit/03e2559a4b7ec09c706dc2bafc442f51e31a8afb

How does this look? All tests pass.

anunaym14: The diff in GH UI looks all over the place, probably worth it to open the file directly

davidzhao: personally, I'm not a huge fan of using reflections when alternatives exists. it's less code, but much harder to read.

PR #1101, opened 2025-06-10
Detailed responses in trunk iter

nishadmusthafa:
This will help discern between cases when
1. No trunks are configured
2. Trunks are configured but none matched.(This can happen when there is no default trunk)
3. There is a default trunk that got applied
4. A specific trunk matched

### DIFF
--- a/sip/sip.go
+++ b/sip/sip.go
@@ -483,8 +483,8 @@ func MatchTrunk(trunks []*livekit.SIPInboundTrunkInfo, call *rpc.SIPCall, opts .
 	return MatchTrunkIter(iters.Slice(trunks), call, opts...)
 }
 
-// MatchTrunkIterDetailed is like MatchTrunkIter but returns detailed match information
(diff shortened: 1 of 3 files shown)


sip/sip.go
```diff
@@ -459,6 +483,99 @@ func MatchTrunk(trunks []*livekit.SIPInboundTrunkInfo, call *rpc.SIPCall, opts .
 	return MatchTrunkIter(iters.Slice(trunks), call, opts...)
 }
 
+// MatchTrunkIterDetailed is like MatchTrunkIter but returns detailed match information
+func MatchTrunkIterDetailed(it iters.Iter[*livekit.SIPInboundTrunkInfo], call *rpc.SIPCall, opts ...MatchTrunkOpt) (*TrunkMatchResult, error) {
```
dennwc: Just a nit, but you could drop `Iter` from the name, if you want.

PR #1108, opened 2025-06-14
Move uploader code from Egress

biglittlebigben:
This is needed both for the pcap upload and cloud-io support for deleting backups


storage/storage.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/livekit/psrpc"
+)
+
+const (
+	maxRetries = 5
+	minDelay   = time.Millisecond * 100
+	maxDelay   = time.Second * 5
+)
+
```
nishadmusthafa: Do we need the StorageConfig and getStorageConfig as well here ? Would a better name for this file be config.go ?

biglittlebigben: The StorageConfig object in egress has fairly egress specific semantics in my opinion. getStorageComfig particularly is about creating an egress configuration from an egress request.

PR #1123, opened 2025-06-30
Adding support for dynamic dispatch rules

nishadmusthafa:
1. Adding the capability for dispatch rule to have a url and http method
2. Allowing the user to respond with a dynamic response for the url
3. Allowing three types of dispatch actions(answer, hangup and transfer)

### DIFF
--- a/livekit/sip.go
+++ b/livekit/sip.go
@@ -579,6 +579,17 @@ func (p *SIPDispatchRuleInfo) Validate() error {
 	if p.Rule == nil {
 		return errors.New("missing rule")
 	}
+
+	// Validate dynamic dispatch rule URL must be HTTPS
+	if dynamicRule := p.Rule.GetDispatchRuleDynamic(); dynamicRule != nil {
+		if dynamicRule.Url == "" {
+			return errors.New("dynamic dispatch rule URL cannot be empty")
+		}
+		if !strings.HasPrefix(dynamicRule.Url, "https://") {
+			return errors.New("dynamic dispatch rule URL must use HTTPS protocol")
+		}
+	}
+
 	return nil
 }
 
--- a/protobufs/livekit_sip.proto
+++ b/protobufs/livekit_sip.proto
@@ -470,6 +470,68 @@ message SIPDispatchRuleCallee {
   bool randomize = 3;
 }
 
+message SIPDispatchRuleDynamic {
+  string url = 1;
+  string method = 2;
+}
+
+// Defines the type of action for a SIP call
+enum DispatchAction {
+  // Answer the call and connect to a LiveKit room
+  ACTION_ANSWER = 0;
+
+  // Reject the call
+  ACTION_REJECT = 1;
+
+  // Transfer the call to another destination
+  ACTION_TRANSFER = 2;
+}
+
+// SIPDispatchRuleDynamicResponse is the response from the webhook URL.
+message SIPDispatchRuleDynamicResponse {
+  // Type of action to take
+  DispatchAction action = 1;
+
+  // Fields used when action is ACTION_ANSWER
+  // -----------------------------------
+  // Room name to connect to
+  string room_name = 2;
+  // Optional pin required to enter room
+  string pin = 3;
+  // Cloud-only, config preset to use
+  string room_preset = 4;
+  // RoomConfiguration to use if the participant initiates the room
+  RoomConfiguration room_config = 5;
+  //valid for only cloud
+  bool krisp_enabled = 6;
+  SIPMediaEncryption media_encryption = 7;
+
+  // Fields used when action is ACTION_REJECT
+  // -----------------------------------
+  // Status code to reject with
+  SIPStatusCode reject_status_code = 8;
+  // Optional reason message for rejection
+  string reject_reason = 9;
+
+  // Fields used when action is ACTION_TRANSFER
+  // -----------------------------------
+  // Destination to transfer to e.g. phone number
+  string transfer_to = 10;
+  // Optional human-readable name or department for transfer
+  string destination_name = 11;
+  // Max time to wait for transfer to be answered
+  google.protobuf.Duration transfer_ringing_timeout = 12;
+
+  // Common fields that apply to any action type
+  // -----------------------------------
+  // User-defined metadata.
+  // Participants created by this rule will inherit this metadata.
+  string metadata = 13;
+  // User-defined attributes.
+  // Participants created by this rule will inherit these attributes.
+  map<string, string> attributes = 14;
+}
+
 message SIPDispatchRule {
   oneof rule {
     // SIPDispatchRuleDirect is a `SIP Dispatch Rule` that puts a user directly into a room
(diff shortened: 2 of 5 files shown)


nishadmusthafa:
This is the cli change -> https://github.com/livekit/livekit-cli/pull/606

livekit/sip.go
```diff
@@ -579,6 +579,17 @@ func (p *SIPDispatchRuleInfo) Validate() error {
 	if p.Rule == nil {
 		return errors.New("missing rule")
 	}
+
+	// Validate dynamic dispatch rule URL must be HTTPS
+	if dynamicRule := p.Rule.GetDispatchRuleDynamic(); dynamicRule != nil {
+		if dynamicRule.Url == "" {
+			return errors.New("dynamic dispatch rule URL cannot be empty")
+		}
+		if !strings.HasPrefix(dynamicRule.Url, "https://") {
```
dennwc, line 588: Plain HTTP should probably be allowed for OSS.

protobufs/livekit_sip.proto
```diff
@@ -470,6 +470,68 @@ message SIPDispatchRuleCallee {
...
+  // Transfer the call to another destination
+  ACTION_TRANSFER = 2;
+}
+
+// SIPDispatchRuleDynamicResponse is the response from the webhook URL.
+message SIPDispatchRuleDynamicResponse {
+  // Type of action to take
+  DispatchAction action = 1;
+
+  // Fields used when action is ACTION_ANSWER
```
dennwc, line 495: This should be a separate message and a `oneof` field. In that case we no longer need a `DispatchAction` enum.

protobufs/livekit_sip.proto
```diff
@@ -470,6 +470,68 @@ message SIPDispatchRuleCallee {
...
+  string room_name = 2;
+  // Optional pin required to enter room
+  string pin = 3;
+  // Cloud-only, config preset to use
+  string room_preset = 4;
+  // RoomConfiguration to use if the participant initiates the room
+  RoomConfiguration room_config = 5;
+  //valid for only cloud
+  bool krisp_enabled = 6;
+  SIPMediaEncryption media_encryption = 7;
```
dennwc, line 507: We should probably put these common fields in some message type and share in for inbound and outbound calls. Could be done later, of course.

protobufs/livekit_sip.proto
```diff
@@ -470,6 +470,68 @@ message SIPDispatchRuleCallee {
...
+message SIPDispatchRuleDynamicResponse {
+  // Type of action to take
+  DispatchAction action = 1;
+
+  // Fields used when action is ACTION_ANSWER
+  // -----------------------------------
+  // Room name to connect to
+  string room_name = 2;
+  // Optional pin required to enter room
+  string pin = 3;
```
dennwc, line 500: I believe this would change the dispatch logic significantly. I wonder if an entered pin should be sent as a new webhook event.

protobufs/livekit_sip.proto
```diff
@@ -552,7 +619,26 @@ message SIPDispatchRuleInfo {
 
   bool krisp_enabled = 11;
   SIPMediaEncryption media_encryption = 12;
-  // NEXT ID: 13
+
+  // Fields from SIPDispatchRuleDynamicResponse for dynamic rules
+  // Type of action to take
+  DispatchAction action = 13;
```
dennwc, line 625: Shouldn't it already be in the `rule` field?

sip/sip.go
```diff
@@ -810,6 +822,8 @@ func MatchDispatchRuleIter(trunk *livekit.SIPInboundTrunkInfo, rules iters.Iter[
 
 // EvaluateDispatchRule checks a selected Dispatch Rule against the provided request.
 func EvaluateDispatchRule(projectID string, trunk *livekit.SIPInboundTrunkInfo, rule *livekit.SIPDispatchRuleInfo, req *rpc.EvaluateSIPDispatchRulesRequest) (*rpc.EvaluateSIPDispatchRulesResponse, error) {
+	log := logger.GetLogger()
+	log.Debugw("Evaluating dispatch rule dailamooooo", "rule", rule)
```
dennwc, line 826: Should be removed?

sip/sip.go
```diff
@@ -831,6 +845,125 @@ func EvaluateDispatchRule(projectID string, trunk *livekit.SIPInboundTrunkInfo,
...
+	)
+
+	switch dynamicRule := rule.GetRule().GetRule().(type) {
+	default:
+		room, rulePin, err = GetPinAndRoom(rule)
+		if err != nil {
+			return nil, twirp.WrapError(twirp.NewError(twirp.Internal, "Failed to get room and pin"), err)
+		}
+	case *livekit.SIPDispatchRule_DispatchRuleDynamic:
+		type dynamicDispatchRequest struct {
```
dennwc, line 861: We should define these as protobufs and use protojson to encode/decode.

sip/sip.go
```diff
@@ -831,6 +845,125 @@ func EvaluateDispatchRule(projectID string, trunk *livekit.SIPInboundTrunkInfo,
...
+			ToUser:    call.To.User,
+			ToHost:    call.To.Host,
+		}
+
+		reqBytes, err := json.Marshal(reqBody)
+		if err != nil {
+			return nil, twirp.WrapError(twirp.NewError(twirp.Internal, "Failed to marshal request"), err)
+		}
+
+		resp, err := http.Post(dynamicRule.DispatchRuleDynamic.GetUrl(), "application/json", bytes.NewReader(reqBytes))
```
dennwc, line 899: We might want an option to pass a custom `http.Client` here.

sip/sip.go
```diff
@@ -831,6 +845,125 @@ func EvaluateDispatchRule(projectID string, trunk *livekit.SIPInboundTrunkInfo,
...
+		err     error
+	)
+
+	switch dynamicRule := rule.GetRule().GetRule().(type) {
+	default:
+		room, rulePin, err = GetPinAndRoom(rule)
+		if err != nil {
+			return nil, twirp.WrapError(twirp.NewError(twirp.Internal, "Failed to get room and pin"), err)
+		}
+	case *livekit.SIPDispatchRule_DispatchRuleDynamic:
```
dennwc, line 860: Worth moving to a separate function.

dennwc (changes requested):

protobufs/livekit_sip.proto
nishadmusthafa: Yeah, DZ was also mentioning webhooks for another category of events. This would be if the agent fails to connect. Are we talking about adding more events here -> https://github.com/livekit/protocol/blob/main/webhook/consts.go#L27 ?

protobufs/livekit_sip.proto
nishadmusthafa: Same point on flat json as above

protobufs/livekit_sip.proto
nishadmusthafa: Not really, action is only in the SIPDispatchRuleDynamicResponse and only when it's a dynamic rule. So we have to explicitly set it once a response comes back.

sip/sip.go
nishadmusthafa: Yes 😅. thanks for catching that.

protobufs/livekit_sip.proto
nishadmusthafa: hah.. that's exactly what I'd done earlier. But @davidzhao and I had a discussion on this. My initial version had a bunch of nesting and using messages in messages. The idea here was to make it simple and give users the ability to use as flat a structure as possible when returning a response.

nishadmusthafa: Just to add more context here, the flat json for response as well as for payload is so that anyone who wants to use this doesn't need to have the sdk to be able to do this.

PR #1124, opened 2025-06-30
Add fields to support AssumeRole in egress S3 uploads

biglittlebigben:


protobufs/livekit_egress.proto
```diff
@@ -235,6 +235,8 @@ message S3Upload {
   string access_key = 1;
   string secret = 2;
   string session_token = 11;
+  string assumed_role_arn = 12;         // ARN of the role to assume for file upload. Egress will make an AssumeRole API call using the provided access_key and secret to assume that role
```
frostbyte73: nit: assume instead of assumed

PR #1126, opened 2025-07-03
Normalize trunk filtering

nishadmusthafa:


utils/sip/sip_processing.go
```diff
@@ -0,0 +1,47 @@
...
+		"-", "",
+		"(", "",
+		")", "",
+	)
+)
+
+// NormalizeNumber normalizes a phone number by removing formatting characters and ensuring it starts with a "+".
+// If the input is empty, it returns an empty string.
+// If the input doesn't match the expected number pattern, it returns the original input unchanged.
+func NormalizeNumber(num string) string {
```
dennwc: It could be in `livekit/sip.go` as well, instead of a new utils package.

livekit/sip.go
```diff
@@ -721,8 +722,14 @@ func (p *ListSIPInboundTrunkRequest) Filter(info *SIPInboundTrunkInfo) bool {
 	if len(p.Numbers) != 0 && len(info.Numbers) != 0 {
 		ok := false
 		for _, num := range info.Numbers {
-			if slices.Contains(p.Numbers, num) {
-				ok = true
+			normalizedNum := siputils.NormalizeNumber(num)
```
dennwc: Could you also keep a direct `slices.Contains(p.Numbers, num)` check as well? We don't need to call regexp if the numbers match exactly.

dennwc (approved):

PR #1127, opened 2025-07-09
Add encryption metadata for data packets

lukasIO:


protobufs/livekit_models.proto
```diff
@@ -315,8 +315,9 @@ message UserPacket {
   optional uint64 end_time = 10;
   // added by SDK to enable de-duping of messages, for INTERNAL USE ONLY
   bytes nonce = 11;
-
-  // NEXT_ID: 12
+  Encryption.Type encryption_type = 12;  // defaults to NONE
+  bytes iv = 13;
```
bcherry: what is `iv` for?

lukasIO: https://developer.mozilla.org/en-US/docs/Web/API/AesGcmParams#iv

a necessary unique initialisation vector for each encryption operation

PR #1133, opened 2025-07-18
Use Wrapper format for wire format messages.

boks1971:
Please see inline for comments.


protobufs/livekit_rtc_v2.proto
```diff
@@ -22,42 +22,43 @@ option ruby_package = "LiveKit::Proto";
...
 import "livekit_models.proto";
 
-message Signalv2ClientMessage {
-  // monotonically increasing number (by 1) starting at some non-zero number (should not rollover)
-  uint32 message_id = 1;
-
-  // last remote message processed
-  uint32 last_processed_remote_message_id = 2;
-
+message Signalv2WireMessage {
```
boks1971, line 25: Maybe needs a better name, but the naming made it clear in my head that this is what comes over the wire and will be decoded as.

Send side processing:
1. Wrap messages in `Envelope` even if it is a single message.
2. Serialise `Envelope`.
3. If size of the serialised output is less than the transport size limit, send as is on the wire.
4. If not, create `Fragment`s and send them.

Receive side processing:
1. Deserialise as `Signalv2WireMessage`.
2. If type is `Envelope`, process messages (`client` should process `Signalv2ServerMessage`s and `server` should process `Signalv2ClientMessage`s) 
3. If type is `Fragment`, reassemble the fragments and go back to Step 1. The reassembled packet should deserialise as `Envelope`.

Some implementation needs:
All `Fragment`s of a packet should be sent together, i. e. there should not be an `Envelope` interspersed. While an interspersed `Envelope` will not break function, it will make it inefficient depending on how the receive side handles messages.

For example, let us assume that messages 4, 5, 6, 7, 8 are in an `Envelope` that needs to be broken up into `Fragment`s.  Let us assume there are three `Fragment`s (fragment1 has messages 4, 5, fragment2: 6, fragment3: 7, 8). Now, let us assume that message 9 gets sent between fragment2 and fragment3. For the receive side, the last full message it was able to process is 3. Now there is a big jump. So, the receive may ask for replay from 4. In a pathological case, this could keep happening.

protobufs/livekit_rtc_v2.proto
```diff
@@ -22,42 +22,43 @@ option ruby_package = "LiveKit::Proto";
...
-    Fragment fragment = 3;
-    Signalv2ClientEnvelope envelope = 4;
-    ConnectRequest connect_request = 5;
+    Envelope envelope = 1;
+    Fragment fragment = 2;
   }
 }
 
-message Signalv2ClientEnvelope {
+message Envelope {
```
boks1971, line 32: making this generic, i. e. not making a server envelope and client envelope. Each side can just process whatever is right for them. The generic `Envelope` and `Fragment` is cleaner in my head.

protobufs/livekit_rtc_v2.proto
```diff
@@ -22,42 +22,43 @@ option ruby_package = "LiveKit::Proto";
...
   // collection of client -> server messages
   repeated Signalv2ClientMessage client_messages = 1;
+ 
+  // collection of server -> client messages
+  repeated Signalv2ServerMessage server_messages = 2;
 }
 
-message Signalv2ServerMessage {
+message Sequencer {
```
boks1971, line 40: splitting out this common part.

cnderrauber (approved):

PR #1139, opened 2025-07-21
Adding the inference grant to the Claim

adrian-cowham:
This grant will be used by the inference gateway for authentication.

### DIFF
--- a/auth/grants.go
+++ b/auth/grants.go
@@ -561,7 +561,7 @@ func (s *AgentGrant) MarshalLogObject(e zapcore.ObjectEncoder) error {
 
 type InferenceGrant struct {
 	// Admin grants to all inference features (LLM, STT, TTS)
-	Admin bool `json:"admin,omitempty"`
+	Perform bool `json:"perform,omitempty"`
 }
 
 func (s *InferenceGrant) Clone() *InferenceGrant {
@@ -579,7 +579,7 @@ func (s *InferenceGrant) MarshalLogObject(e zapcore.ObjectEncoder) error {
 		return nil
 	}
 
-	e.AddBool("Admin", s.Admin)
+	e.AddBool("Perform", s.Perform)
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


real-danm (approved):

adrian-cowham:
I added Gateway specific guid IDs. These are used as options when connecting to the psql db. A psql connection is required for when calling `backend-common/auth/accesskey.NewAccessKeyProvider`

auth/grants.go
```diff
@@ -554,6 +559,32 @@ func (s *AgentGrant) MarshalLogObject(e zapcore.ObjectEncoder) error {
 
 // ------------------------------------------------------------------
 
+type InferenceGrant struct {
+	// Admin grants to all inference features (LLM, STT, TTS)
+	Admin bool `json:"admin,omitempty"`
```
davidzhao: should we make this more nuanced vs having to rely on the "admin" permission?

i.e. does it make sense to have separate permission tokens for diff services? i.e. either:

```
llm: bool
tts: bool
stt: bool
```

or maybe a single "perform" permission to allow performing an inference? (as opposed to admin could mean update settings or other features)

utils/guid/id.go
```diff
@@ -56,6 +56,7 @@ const (
 	CloudAgentVersionPrefix = "CAV_"
 	CloudAgentSecretPrefix  = "CAS_"
 	CloudAgentWorkerPrefix  = "CAW_"
+	CloudAgentGatewayPrefix = "CAG_"
```
davidzhao: is the idea that for each inference request we'd have an ID?

if so could we just call it `CloudInference` ?

auth/grants.go
adrian-cowham: I was going to add the finer grain permissions but decided to go with the simpler approach since we can add those permissions later. Happy to add them now though. Also, I like the name `perform` better, will change it.

utils/guid/id.go
adrian-cowham: These values are prefixes for node ids. The node ids are used as labels for the DB prometheus metrics. Happy to rename it but this seems to be the convention I noticed.

adrian-cowham: I removed these values since the gateway no longer needs a DB connection.

PR #1146, opened 2025-07-28
Lk phone numbers

nishadmusthafa:


protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// PhoneNumberInventoryItem - Represents an available phone number for purchase
+message PhoneNumberInventoryItem {
+  string phone_number = 1;        // Phone number in E.164 format (e.g., "+14155552671")
+  string country_code = 2;        // Country code (e.g., "US")
+  string area_code = 3;           // Area code (e.g., "415")
+  string locality = 4;            // City/locality (e.g., "San Francisco")
+  string region = 5;              // State/region (e.g., "CA")
+  string capabilities = 6;        // Comma-separated capabilities (e.g., "voice,sms")
```
dennwc:
```suggestion
  repeated string capabilities = 6;        // a list of capabilities (e.g., ["voice", "sms"])
```

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+  int64 monthly_cost_cents = 7;   // Monthly cost in cents
```
dennwc: Do we expect other currencies going forward?

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+  rpc ListPurchasedPhoneNumbers(ListPurchasedPhoneNumbersRequest)
+      returns (ListPurchasedPhoneNumbersResponse) {}
+  
+  // Release a purchased phone number
+  rpc ReleasePhoneNumber(ReleasePhoneNumberRequest)
+      returns (google.protobuf.Empty) {}
+}
+
+// ListPhoneNumberInventoryRequest - Request to list available phone numbers
+message ListPhoneNumberInventoryRequest {
```
dennwc, line 50: Do we need a filter for capabilities here? And a provider, maybe?

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+  string area_code = 2;           // Optional: Filter by area code (e.g., "415")
+  string pattern = 3;             // Optional: Filter by pattern (e.g., "***-***-1234")
+  int32 limit = 4;                // Optional: Maximum number of results (default: 50)
+  int32 offset = 5;               // Optional: Number of results to skip (default: 0)
+}
+
+// ListPhoneNumberInventoryResponse - Response containing available phone numbers
+message ListPhoneNumberInventoryResponse {
+  repeated PhoneNumberInventoryItem items = 1;  // List of available phone numbers
+  int32 total_count = 2;                        // Total number of available numbers
```
dennwc: Is it practical to return the total number of all the numbers available for purchase? I assume that's a very long list and some providers may not even return the count.

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+  int32 limit = 3;                              // Limit used in the request
+  int32 offset = 4;                             // Offset used in the request
```
dennwc: Offset would work for most providers, I guess, but I was thinking maybe a more generic `string token` or `string next` would make sense? This way we could still pass the next page number for certain providers and possibly a different kind of  pagination token for others.

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+  int32 limit = 1;                // Optional: Maximum number of results (default: 50)
+  int32 offset = 2;               // Optional: Number of results to skip (default: 0)
+}
+
+// ListPurchasedPhoneNumbersResponse - Response containing purchased phone numbers
+message ListPurchasedPhoneNumbersResponse {
+  repeated PurchasedPhoneNumber items = 1;  // List of purchased phone numbers
+  int32 total_count = 2;                   // Total number of purchased numbers
+  int32 limit = 3;                         // Limit used in the request
+  int32 offset = 4;                        // Offset used in the request
```
dennwc: It's okay to use page number here if it makes sense, but for SIP we used a bit different pagination strategy. We sort the items by ID and pass the `string afterID` for pagination. This way even in case items were added/removed, the pagination never produces duplicate items. Do you think the same approach makes sense here?

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+  string phone_number = 1;        // Phone number to release (e.g., "+1234567890")
+}
+
+// PhoneNumberInventoryItem - Represents an available phone number for purchase
+message PhoneNumberInventoryItem {
+  string phone_number = 1;        // Phone number in E.164 format (e.g., "+14155552671")
+  string country_code = 2;        // Country code (e.g., "US")
+  string area_code = 3;           // Area code (e.g., "415")
+  string locality = 4;            // City/locality (e.g., "San Francisco")
+  string region = 5;              // State/region (e.g., "CA")
```
dennwc: These fields are very similar to ones in `PurchasedPhoneNumber`, maybe we could make a `PhoneNumberInfo` with all the common ones? We did this mistake with original SIP in the past by duplicating fields, adding anew new common field becomes very error-prone. So if we put it into a common struct the code can just pass it along without the need for explicit conversion of fields.

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
+syntax = "proto3";
+
+package livekit;
+
+option go_package = "github.com/livekit/protocol/livekit";
+
+import "google/protobuf/empty.proto";
+
+// Public Phone Number Service - External API for phone number management
+service PhoneNumberService {
```
dennwc, line 31: If this makes it to OSS, do you think we could adapt it to Twilio API for example?

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+message ListPhoneNumberInventoryResponse {
+  repeated PhoneNumberInventoryItem items = 1;  // List of available phone numbers
+  int32 total_count = 2;                        // Total number of available numbers
+  int32 limit = 3;                              // Limit used in the request
+  int32 offset = 4;                             // Offset used in the request
+}
+
+// PurchasePhoneNumberRequest - Request to purchase a phone number
+message PurchasePhoneNumberRequest {
+  string phone_number = 1;        // Phone number to purchase (e.g., "+1234567890")
```
dennwc: Do you think making it an array makes sense to allow batch orders?

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// PurchasedPhoneNumber - Represents a phone number owned by a LiveKit project
+message PurchasedPhoneNumber {
+  string id = 1;                  // Unique identifier for the phone number
+  string phone_number = 2;        // Phone number in E.164 format (e.g., "+14155552671")
+  string country_code = 3;        // Country code (e.g., "US")
+  string area_code = 4;           // Area code (e.g., "415")
+  string national_number = 5;     // National number without country code (e.g., "4155552671")
+  bool is_mobile = 6;             // Whether this is a mobile number
```
dennwc: Are there any other kinds? I've seen "toll-free" as a separate category as well, for example. If so, we might need an enum here.

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,98 @@
...
+  string status = 7;              // Current status ("active", "pending", "released")
```
dennwc: The number of these states seem fixed, we probably could define most of them upfront as an enum.

protobufs/livekit_phone_number.proto
nishadmusthafa: Yeah, I actually gave this some thought. In this iteration, I didn't really look into batching as much. This variation you suggested is one kind of batch implementation. There's also another where you can just give an area code and say you want 10 numbers and you don't care about what the numbers are. But I think definitely makes sense to make this an array.

protobufs/livekit_phone_number.proto
nishadmusthafa: I didn't want to add capability in the request from the get go as we wanted to mostly focus on voice now. SMS certainly is something that should be supported when we support that as a channel.

For provider, my thought was to keep it separate from this request and be able to make config changes like which provider for which country in a different place. Right now, that is more in the cloud. My gut tells me that the implementations will look quite different between cloud and OSS for that.

protobufs/livekit_phone_number.proto
nishadmusthafa: yeah, I think that's a good idea. Would take a stab at it in a later iteration. Also as I mentioned in the comment below, we might have variations in how that is implemented in cloud/oss for things like what is supported. Billing probably is not something we might support in OSS as it would be something very custom to implementation.

protobufs/livekit_phone_number.proto
nishadmusthafa: Good point. I think it might make sense as an optional field in case we receive it.

protobufs/livekit_phone_number.proto
nishadmusthafa: Hmm.. sometimes, people might just want the numbers numerically sorted for presentation/order/etc. Will need to see how much the ID sorting makes those kinds of things easier/harder.

protobufs/livekit_phone_number.proto
nishadmusthafa: I actually had a different thought here. Will make changes.

protobufs/livekit_phone_number.proto
nishadmusthafa: will change

protobufs/livekit_phone_number.proto
dennwc: ID could be the actual number, btw. Or the ID could have a prefix with a number in it.

protobufs/livekit_phone_number.proto
nishadmusthafa: Decided to leave it out for now.

protobufs/livekit_phone_number.proto
nishadmusthafa: Only thing to keep in mind here is what would happen if the provider doesn't have a bulk api. We'd need to implement that additional logic. maybe we could validate to only do one in the beginning and add that capability in the future. In the immediate term, this is not an issue as bulk apis exist.

protobufs/livekit_phone_number.proto
nishadmusthafa: actually I went with more or less what you suggested. I added some changes to make the pricing data better.

protobufs/livekit_phone_number.proto
```diff
@@ -29,70 +48,80 @@ service PhoneNumberService {
...
+// GlobalPhoneNumber represents a phone number with standardized format
+message GlobalPhoneNumber {
+  string id = 1;                  // unique identifier
+  string e164_format = 2;         // e.g., "+14155552671"
+  string country_code = 3;        // e.g., "US"
+  string area_code = 4;           // e.g., "415"
+  string number_type  = 5;        // e.g., local, toll-free, national, mobile
+  string locality = 6;            // City/locality (e.g., "San Francisco")
+  string region = 7;              // State/region (e.g., "CA")
+  int64 spam_score = 8;           // can be used later for fraud detection
```
dennwc:
```suggestion
  double spam_score = 8;           // can be used later for fraud detection
```

protobufs/livekit_phone_number.proto
```diff
@@ -29,70 +48,80 @@ service PhoneNumberService {
...
+  int64 created_at = 9;           // timestamp when created
+  int64 updated_at = 10;          // timestamp when updated
```
dennwc:
```suggestion
  google.protobuf.Timestamp created_at = 9;           // timestamp when created
  google.protobuf.Timestamp updated_at = 10;          // timestamp when updated
```
We should use `Timestamp` type in the new protos, since it's always unclear in which units the int timestamp is.

protobufs/livekit_phone_number.proto
```diff
@@ -29,70 +48,80 @@ service PhoneNumberService {
...
-  bool is_mobile = 6;             // Whether this is a mobile number
-  string status = 7;              // Current status ("active", "pending", "released")
-  int64 assigned_at = 8;          // Timestamp when the number was assigned
-  int64 released_at = 9;          // Timestamp when the number was released (if applicable)
-} 
\ No newline at end of file
+  GlobalPhoneNumber phone_number = 1;  // Common phone number fields
+  PhoneNumberStatus status = 2;        // Current status of the phone number
+  int64 assigned_at = 3;               // Timestamp when the number was assigned
+  int64 released_at = 4;               // Timestamp when the number was released (if applicable)
```
dennwc:
```suggestion
  google.protobuf.Timestamp assigned_at = 3;               // Timestamp when the number was assigned
  google.protobuf.Timestamp released_at = 4;               // Timestamp when the number was released (if applicable)
```

protobufs/livekit_phone_number.proto
```diff
@@ -29,70 +48,80 @@ service PhoneNumberService {
...
+  string page_token = 4;          // Optional: Token for pagination (empty for first page)
 }
 
 // ListPhoneNumberInventoryResponse - Response containing available phone numbers
 message ListPhoneNumberInventoryResponse {
   repeated PhoneNumberInventoryItem items = 1;  // List of available phone numbers
-  int32 total_count = 2;                        // Total number of available numbers
-  int32 limit = 3;                              // Limit used in the request
-  int32 offset = 4;                             // Offset used in the request
+  string next_page_token = 2;                   // Token for next page (empty if no more pages)
```
dennwc: Maybe we should make a new `TokenPagination` in the `livekit_models.proto`, similar to existing `Pagination`. It will allow us to write a generic pagination code an reuse it for all models.

protobufs/livekit_phone_number.proto
dennwc: Do you think making spam score a `double` makes sense here?

dennwc (approved):
Looks good now, thanks a lot! One small item remains only.

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,129 @@
...
+  string number_type  = 5;        // e.g., local, toll-free, national, mobile
+  string locality = 6;            // City/locality (e.g., "San Francisco")
+  string region = 7;              // State/region (e.g., "CA")
+  double spam_score = 8;          // can be used later for fraud detection
+  google.protobuf.Timestamp created_at = 9;           // timestamp when created
+  google.protobuf.Timestamp updated_at = 10;          // timestamp when updated
+}
+
+// TelephonyCost represents the pricing structure for a specific telephony service
+message TelephonyCost {
```
davidzhao, line 106: I don't think this should be exposed. our other APIs don't expose cost/pricing information currently.

when we do, it should not be dependent on a particular resource. but rather types of resources.

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,129 @@
...
+  // List available phone numbers in inventory
+  rpc ListPhoneNumberInventory(ListPhoneNumberInventoryRequest)
+      returns (ListPhoneNumberInventoryResponse) {}
+  
+  // Purchase a phone number from inventory
+  rpc PurchasePhoneNumber(PurchasePhoneNumberRequest)
+      returns (PurchasePhoneNumberResponse) {}
+  
+  // List purchased phone numbers for a project
+  rpc ListPurchasedPhoneNumbers(ListPurchasedPhoneNumbersRequest)
```
davidzhao, line 41:
```suggestion
  rpc ListPhoneNumbers(ListPhoneNumbersRequest)
```

davidzhao: IMO by default list is listing resources that you own

protobufs/livekit_phone_number.proto
```diff
@@ -0,0 +1,129 @@
...
+  repeated TelephonyCost costs = 3;       // Array of costs for different services
+}
+
+// PurchasedPhoneNumber - Represents a phone number owned by a LiveKit project
+message PurchasedPhoneNumber {
+  GlobalPhoneNumber phone_number = 1;  // Common phone number fields
+  PhoneNumberStatus status = 2;        // Current status of the phone number
+  google.protobuf.Timestamp assigned_at = 3;               // Timestamp when the number was assigned
+  google.protobuf.Timestamp released_at = 4;               // Timestamp when the number was released (if applicable)
+  SIPDispatchRuleInfo sip_dispatch_rule = 5;  // Optional: Associated SIP dispatch rule
```
davidzhao, line 128: should this be set on the dispatch rule itself? we already have the "inbound numbers" field in there

davidzhao (commented):
it'd be great if we can simplify the PurchasedPhoneNumber, GlobalPhoneNumber, and PhoneNumberInventoryItem.

can they be the same object?

PR #1151, opened 2025-08-06
join request as proto message

boks1971:


davidzhao (approved):

protobufs/livekit_rtc.proto
```diff
@@ -471,3 +475,30 @@ message RequestResponse {
...
+  // will overwrite if the same key is in the token
+  // will not delete keys from token if there is a key collision and this sets that key to empty value
+  map<string, string> participant_attributes = 4;
+
+  repeated AddTrackRequest add_track_requests = 5;
+  SessionDescription publisher_offer = 6;
+
+  bool reconnect = 7;
+  ReconnectReason reconnect_reason = 8;
+  string participant_sid = 9;
```
boks1971, line 498: not adding `SyncState` as it contains the subscriber PC offer/answer and those could get large for a large room and that would cause issues with query param size.

protobufs/livekit_rtc.proto
```diff
@@ -471,3 +475,30 @@ message RequestResponse {
...
+  ConnectionSettings connection_settings = 2;
+
+  string metadata = 3; // if not empty, will overwrite `metadata` in token
+
+  // will set keys provided via this
+  // will overwrite if the same key is in the token
+  // will not delete keys from token if there is a key collision and this sets that key to empty value
+  map<string, string> participant_attributes = 4;
+
+  repeated AddTrackRequest add_track_requests = 5;
```
davidzhao, line 493: I think adding these should be ok, but there is a browser limit of 64KB. are we going to fit ok within the limits?

davidzhao: if there's a chance we could go over, then the client would need logic to filter certain ones out and send after the fact.

davidzhao (approved):

protobufs/livekit_rtc.proto
boks1971: true, attributes and metadata could add up too. So, clients would have to check the size and adjust accordingly.

protobufs/livekit_rtc.proto
```diff
@@ -471,3 +475,30 @@ message RequestResponse {
...
 }
+
+message ConnectionSettings {
+  bool auto_subscribe = 1;
+  bool adaptive_stream = 2;
+  optional bool subscriber_allow_pause = 3;
+  bool disable_ice_lite = 4;
+}
+
+message JoinRequest {
```
boks1971, line 482: clients should protobuf marshal this, base64 encode it and send it as query param `join_request`.

davidzhao: should we require compression as well? i.e. marshal, gzip, base64

boks1971: yep, that would be good.

lukasIO: as a requirement I'm not sure that's ideal when considering embedded use cases

boks1971: Got it @lukasIO. I can remove it on server side.

Wondering if we should also have `compressed_join_request` query param to have gzip'ed version which other clients can use.

lukasIO: alternatively we could consider an protobuf envelope that defines the compression (if any)

PR #1153, opened 2025-08-07
Wrapped join request to indicate compression.

boks1971:


protobufs/livekit_rtc.proto
```diff
@@ -498,3 +498,12 @@ message JoinRequest {
...
 
+message WrappedJoinRequest {
+  enum Compression {
+    NONE = 0;
+    GZIP = 1;
+  }
+
+  Compression compression = 1;
+  bytes join_request = 2;   // marshalled JoinRequest + potentially compressed
+}
```
boks1971, line 509: Steps would be
1. Create `JoinRequest`.
2. Proto marshal.
3. Maybe apply compress on marshalled proto.
4. Create `WrappedJoinRequest` with (2) or (3) and set the `Compression` field appropriately.
5. Proto marshal `WrappedJoinRequest`.
6. base64 encode (5).
7. Send as query param `join_request`.

PR #1158, opened 2025-08-09
Allow clients to specify video layers mode.

boks1971:
The main issue is distinguishing between one RTP streaming carry multiple spatial layers (example SVC codecs) vs one spatial layer (codecs that are typically simulcast). The enum names are long, but wanted them to be clear.


protobufs/livekit_rtc.proto
```diff
@@ -119,16 +119,16 @@ message SimulcastCodec {
...
-  // NEXT-ID: 5
+  // NEXT-ID: 6
 }
 
 message AddTrackRequest {
   // client ID of track, to match it when RTC track is received
   string cid = 1;
   string name = 2;
   TrackType type = 3;
-  // to be deprecated in favor of layers
```
boks1971, line 131: removing this comment as this is used in adaptive streaming and also useful to have this as capture size or max size. The width and height are also in `UpdateLocalVideoTrack` message. So, this comment is misleading.

davidzhao (approved):

PR #1163, opened 2025-08-11
add hash function to generate common guid

paulwe:


.changeset/nine-melons-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
```
boks1971, line 2: This is probably going to fail CI in the next PR?

PR #1166, opened 2025-08-12
Add webhook for aborted participant connection.

boks1971:
Will be sent when the participant session is closed without participant
becoming active.


protobufs/livekit_analytics.proto
```diff
@@ -88,6 +88,7 @@ enum AnalyticsEventType {
   ROOM_ENDED = 1;
   PARTICIPANT_JOINED = 2;
   PARTICIPANT_LEFT = 3;
+  PARTICIPANT_CONNECTION_ABORTED = 45;
```
boks1971, line 91: @shishirng adding a new analytics type event for cases where participant joins, but never becomes active, i. e. participant session is closed without participant becoming active. Is it okay to add and send new events without changes on analytics side?

cnderrauber (approved):

davidzhao (approved):
lgtm!

PR #1167, opened 2025-08-12
Support additional fields for agent versions

yoonsio:
This is required to support new features:
* track deployed timestamp for agent versions
* support custom labels for agent versions


protobufs/livekit_cloud_agent.proto
```diff
@@ -82,6 +82,8 @@ message AgentVersion {
     string version = 1;
     bool current = 2;
     google.protobuf.Timestamp created_at = 3;
+    google.protobuf.Timestamp deployed_at = 4;
+    map<string, string> labels = 5;
```
rektdeckard: Just curious, what are we using a map instead of a string slice for? Will there be different classes of labels (`user`, `system`, etc)?

yoonsio: Labels are usually key-value pair. (eg `env=production`)

rektdeckard: So these are like attributes then? And they're meant to be set by end-users, correct?

davidzhao: should we standardize on the terminology of `attributes` then?

yoonsio: @davidzhao If that is the terminology we use across the stack, I'm fine with that.
GCP uses `labels` for their resources.
Screenshot 2025-08-12 at 3 14 13 PM

yoonsio: Kubernetes also use `labels` and `annotations` for their resources.

yoonsio: synced offline. Will use `attributes` to keep things consistent within [our ecosystem](https://github.com/search?q=repo%3Alivekit%2Fprotocol%20attributes&type=code).

PR #1168, opened 2025-08-13
Fix changeset

boks1971:


paulwe (approved):

boks1971:
Oh, the bot is reporting this
```
Some errors occurred when validating the changesets config:
The package or glob expression "github.com/livekit/protocol" specified in the `fixed` option does not match any package in the project. You may have misspelled the package name or provided an invalid glob expression. Note that glob expressions must be defined according to https://www.npmjs.com/package/micromatch.
```

And I did not create a changeset for this. Looks like DZ did and it got the default fake name in there. Guess this works only for `@livekit` package names or maybe it needs to be configured for the other pattern also.

@lukasIO is this a config issue?

Issue #856, opened 2024-10-15
Make UnlikelyLogger as interface instead of struct


xalanq:
https://github.com/livekit/protocol/blob/c0c2e6e03dcc9e8b5d75568b78b91f73d8fd0d6e/logger/logger.go#L103C1-L103C29

Interface `Logger` has an method `WithUnlikelyValues` which returns a `UnlikelyLogger` with two private fields `logger` and `keysAndValues`. But the `UnlikelyLogger` is not an interface, there is no way to `mock` or `reimplement` the `Logger` because of this.

zesun96:
hi, did there any solution, I want to wirte log to file.

PR #1169, opened 2025-08-13
Added the UnlikelyLogger constructor

zesun96:
fix https://github.com/livekit/protocol/issues/856


davidzhao:
you can simply use any existing logger and do `WithUnlikelyValues`

PR #1174, opened 2025-08-24
Protocol changes to potentially support single peer connection mode.

boks1971:
Please see inline for detailed notes.


protobufs/livekit_rtc.proto
```diff
@@ -509,3 +513,8 @@ message WrappedJoinRequest {
   Compression compression = 1;
   bytes join_request = 2;   // marshalled JoinRequest + potentially compressed
 }
+
+message MediaSectionsRequirement {
+  uint32 num_audios = 1;
+  uint32 num_videos = 2;
+}
```
boks1971, line 522: This is a `SignalResponse` sent by server when it needs a renegotiation. This happens when server is adding or removing subscribed tracks.

As this can happen when server is removing a track, the number of audios/videos in this message be 0.

The client should add number of audio/video sections (using `addTransceiver` using `audio|video` kind and `recvonly` direction) and do a negotiation exchange.

A couple of other bits of notes
1. Clients can add more than what is specified here. This is the minimum number it should add. Clients can also add some sections gratuitously in the initial offer on publisher peer connection too.
2. This is a bit inefficient now as server may send this more than once and create more media sections than necessary as there is no good interlock mechanism. Please let me know if you have any protocol based ideas to prevent extra sections from getting created as SDPs are large.

protobufs/livekit_rtc.proto
```diff
@@ -378,14 +380,16 @@ message RoomMovedResponse {
...
   UpdateSubscription subscription = 2;
   repeated TrackPublishedResponse publish_tracks = 3;
   repeated DataChannelInfo data_channels = 4;
   // last received server side offer before reconnecting
-  SessionDescription offer = 5;
+  SessionDescription subscriber_offer = 5;
   repeated string track_sids_disabled = 6;
   repeated DataChannelReceiveState datachannel_receive_states = 7;
+  SessionDescription publisher_offer = 8;
+  SessionDescription publisher_answer = 9;
```
boks1971: For the single peer connection case, the publisher side offer/answer is needed.

There are two possible approaches
1. Rename old offer/answer fields like ☝️ and create new fields for publisher offer/answer.
2. Leave this message unchanged and use implicit knowledge that client will send offer/answer from the correct peer connection.

Thoughts?

protobufs/livekit_internal.proto
```diff
@@ -181,6 +181,7 @@ message StartSession {
   repeated livekit.AddTrackRequest add_track_requests = 21;
   SessionDescription publisher_offer = 22;
   SyncState sync_state = 23;
+  bool use_single_peer_connection = 24;
```
boks1971, line 184: This will be set to `true` when client connects using `WrappedJoinRequest` .

protobufs/livekit_rtc.proto
cnderrauber: is it ok to change the field name of protocol message? i would prefer `2` as `1` will cause compile error if sdks update the protocol but don't want to support single peer connection now (may be in future).

cnderrauber (approved):
looks good!

protobufs/livekit_rtc.proto
boks1971: Renaming field is fine. Only the field ids go over the wire.

Yeah clients will need to change the two field names if they update protocol. It should be fine to not include the new fields if they do not want to support single peer connection.

boks1971: Will revert changes here and use the implicit assumption. If somebody has a differing opinion, will change it then.

PR #1179, opened 2025-08-27
Lk phone numbers update get

nishadmusthafa:
Adding update and get apis for purchased phone numbers

### DIFF
--- a/protobufs/livekit_phone_number.proto
+++ b/protobufs/livekit_phone_number.proto
@@ -4,7 +4,6 @@ package livekit;
 
 option go_package = "github.com/livekit/protocol/livekit";
 
-import "google/protobuf/empty.proto";
 import "google/protobuf/timestamp.proto";
 import "livekit_models.proto";
 import "livekit_sip.proto";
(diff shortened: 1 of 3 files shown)


protobufs/livekit_phone_number.proto
```diff
@@ -63,6 +70,7 @@ message ListPhoneNumberInventoryResponse {
 // PurchasePhoneNumberRequest - Request to purchase phone numbers
 message PurchasePhoneNumberRequest {
   repeated string phone_numbers = 1;  // Phone numbers to purchase (e.g., ["+1234567890", "+1234567891"])
+  string sip_dispatch_rule_id = 2;    // Optional: SIP dispatch rule ID to apply to all purchased numbers
```
dennwc, line 72: This might be misleading, since it's not linking that rule ID to the number, it's cloning the rule, right?

So to update a rule for all numbers, one would update the rule and then update numbers with that rule ID again.

Worth documenting it or adding rule id to the number model directly, if we want to link them persistently.

nishadmusthafa: It isn't meant to be a clone. I've made it consistent by using `sip_dispatch_rule_id` in `PurchasedPhoneNumber`

dennwc (approved):

PR #1181, opened 2025-08-28
Fix javascript changelog

lukasIO:
JS package has its own changelog file

### DIFF
--- a/packages/javascript/CHANGELOG.md
+++ b/packages/javascript/CHANGELOG.md
@@ -2,9 +2,11 @@
 
 ## 1.41.0
 
-### Patch Changes
+### Minor Changes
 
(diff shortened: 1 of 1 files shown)


packages/javascript/CHANGELOG.md
```diff
@@ -1,13 +1,11 @@
 # @livekit/protocol
 
-## 2.0.0
+## 1.41.0
 
-### Minor Changes
+### Patch Changes
```
boks1971: Minor? or Patch?

boks1971 (approved):

packages/javascript/CHANGELOG.md
lukasIO: thanks for catching that, fixed!

PR #1184, opened 2025-09-02
Protocol error handling

boks1971:
For signalling responses,
- `RequestResponse` will be used
- If request has a `request_id` field (currently only in `UpdateParticipantMetadata`), a success response is also expected by sender.
- For errors, the request is included in the response so that senders have context of the request which caused the error.

This PR adds a few requests to the response. We can add more over time.


boks1971:
@lukasIO @hiroshihorie please let me know if you have feedback. Going to merge this on @anunaym14 's review.

PR #1188, opened 2025-09-05
Adding a call context field to CallInfo for dynamically setting info …

nishadmusthafa:
…about the call


protobufs/livekit_sip.proto
```diff
@@ -719,8 +720,9 @@ message SIPCallInfo {
   string audio_codec = 20;
   string media_encryption = 21;
   string pcap_file_link = 25;
+  map<string, google.protobuf.Any> call_context = 26;
```
dennwc: `repeated` maybe?

dennwc (approved):

PR #1194, opened 2025-09-08
RPC changes for ingress proxy WHIP

cnderrauber:
* Sfu notify ingress about WHIP webrtc connection closed 
* Sweeper delete whip session on sfu
* Add `sub_skip_reference_ts` in participant info if the publisher's rtcp sr is incomplete for simulcast (will be set for obs whip session)


protobufs/livekit_models.proto
```diff
@@ -174,8 +174,10 @@ message ParticipantInfo {
   map<string, string> attributes = 15;
   DisconnectReason disconnect_reason = 16;
   repeated KindDetail kind_details = 18;
+  // indicates that subscriber subscribes to track of the participant can skip reference timestamps
```
boks1971: I think this should not be in `ParticipantInfo`. This is sent as update. Other participants need not know this. Can we move this to the request where WHIP is created from ingress?

cnderrauber: The field is set in the WHIP handle but we need to broadcast it to all the subscribers. It is not the best place that the client sdk doesn't need to know this,  but I can't find a better way to put the `internal` property so put it here. Do you have a suggestion place?

boks1971: ah got it, sorry my bad, yeah this needs to be known to subscribers, let me think

boks1971: Can we add a mode [here](https://github.com/livekit/protocol/blob/f6072128d2f5bb469b75598ed55524d57e2aaa02/protobufs/livekit_models.proto#L256) so that it gets carried with track info? TrackInfo also goes to client and clients need not know about it, but I feel keeping it at track level is a bit better and we can make it an enum so that it is not a separate field.

Something like `ONE_SPATIAL_LAYER_PER_STREAM_NO_RTCP_SENDER_REPORT` or something like that (sorry, that name is horrible :-) )

cnderrauber: will look at it

PR #1198, opened 2025-09-08
feat: initial connectors api

anunaym14:


protobufs/connectors/whatsapp_api.proto
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package livekit.connectors;
+option go_package = "github.com/livekit/protocol/connectors";
+option csharp_namespace = "LiveKit.Proto.Connectors";
+option ruby_package = "LiveKit::Proto::Connectors";
+
+import "connectors/whatsapp_common.proto";
```
biglittlebigben: It's not immediately clear to me what the messages in this file are used for. It's likely to be similarly confusing for customers looking at these files for the first time. Could you provide some background here and add some inline docs?

protobufs/connectors/whatsapp_api.proto
```diff
@@ -0,0 +1,53 @@
...
+
+message WhatsAppCallRequest {
+  string messaging_product = 1; // always "whatsapp"
```
biglittlebigben: Why is this field needed if it's always "whatsapp"? Is it to identify it when serialized as something else than protrobuf?

protobufs/connectors/whatsapp_api.proto
```diff
@@ -0,0 +1,53 @@
...
+  optional string to = 2;
+  optional string call_id = 4;
```
biglittlebigben: This is skipping `3`

protobufs/connectors/whatsapp_api.proto
```diff
@@ -0,0 +1,53 @@
...
+  WhatsAppCallAction action = 5; // action to perform
+  optional WhatsAppSession session = 6;
+  optional string biz_opaque_callback_data = 7;
```
biglittlebigben: What is in this field?

protobufs/connectors/whatsapp_api.proto
```diff
@@ -0,0 +1,53 @@
...
+
+message WhatsAppCallResponse {
+  optional bool success = 1;
+  optional string messaging_product = 2;
+  optional WhatsAppApiError error = 3;
+  repeated WhatsAppCall calls = 4;
+}
+
+message WhatsAppApiError {
+  uint32 code = 1;
```
biglittlebigben: What is this code namespace?

protobufs/connectors/whatsapp_api.proto
```diff
@@ -0,0 +1,53 @@
...
+}
+
+message WhatsAppCallResponse {
+  optional bool success = 1;
+  optional string messaging_product = 2;
+  optional WhatsAppApiError error = 3;
+  repeated WhatsAppCall calls = 4;
+}
+
+message WhatsAppApiError {
```
biglittlebigben: While it sometimes make sense to nest messages to avoid structures becoming too long, nested structures are also more cumbersome to use. What is the incentive for a separate error message here. Is this expected to be reused?

protobufs/connectors/whatsapp_api.proto
```diff
@@ -0,0 +1,53 @@
...
+enum WhatsAppCallAction {
+  WHATSAPP_CALL_ACTION_CONNECT = 0;
+  WHATSAPP_CALL_ACTION_PRE_ACCEPT = 1;
+  WHATSAPP_CALL_ACTION_ACCEPT = 2;
+  WHATSAPP_CALL_ACTION_REJECT = 3;
+  WHATSAPP_CALL_ACTION_TERMINATE = 4;
+}
+
+message WhatsAppCallResponse {
+  optional bool success = 1;
```
biglittlebigben: Should this be a status enum instead of a boolean (WhatsAppCallStatus maybe)?

protobufs/connectors/whatsapp_api.proto
```diff
@@ -0,0 +1,53 @@
...
+  optional string messaging_product = 2;
```
biglittlebigben: Any reason to make this field optional? It will default to an empty string otherwise, which seems like a fine default unless we plan to have a messaging product with no name. optionals are useful, but they tend to be more cumbersome to use.

protobufs/connectors/whatsapp_common.proto
```diff
@@ -0,0 +1,63 @@
...
+}
+
+enum WhatsAppSdpType {
+  WHATSAPP_SDP_TYPE_OFFER = 0;
+  WHATSAPP_SDP_TYPE_ANSWER = 1;
+}
+
+message WhatsAppCall {
+  string id = 1;
+  optional string to = 2;
```
biglittlebigben: Any strong reason to use optionals here again?

protobufs/connectors/whatsapp_webhook.proto
```diff
@@ -0,0 +1,87 @@
...
+
+package livekit.connectors;
+option go_package = "github.com/livekit/protocol/connectors";
+option csharp_namespace = "LiveKit.Proto.Connectors";
+option ruby_package = "LiveKit::Proto::Connectors";
+
+import "google/protobuf/timestamp.proto";
+import "connectors/whatsapp_common.proto";
+
+message WhatsAppCallWebhook {
```
biglittlebigben: There are 4 levels of nesting before we get to the payload here: Webhook->WebhookEntry->WebhookChange->WebhookValue. Could you detail why these are needed?

protobufs/connectors/whatsapp_common.proto
biglittlebigben: How do "to" and "from" relate to WhatappContact below?

protobufs/connectors/whatsapp_webhook.proto
```diff
@@ -0,0 +1,87 @@
...
+}
+
+message WhatsAppWebhookChange {
+  WhatsAppWebhookValue value = 1;
+  string field = 2;  // "calls" for call events
+}
+
+message WhatsAppWebhookValue {
+  string messaging_product = 1;  // Always "whatsapp"
+  WhatsAppMetadata metadata = 2;
```
biglittlebigben: Do we need the extra WhatsAppMetadata structure here or should this be flattened?

protobufs/connectors/whatsapp_webhook.proto
```diff
@@ -0,0 +1,87 @@
...
+message WhatsAppContacts {
+  WhatsAppProfile profile = 1;
+  string wa_id = 2;
+}
+
+message WhatsAppProfile {
+  string name = 1;
+}
+
+message WhatsAppError {
```
biglittlebigben: How does this relate to WhatsAppApiError?

protobufs/connectors/whatsapp_webhook.proto
```diff
@@ -0,0 +1,87 @@
...
+  string phone_number_id = 2;
+}
+
+message WhatsAppContacts {
+  WhatsAppProfile profile = 1;
+  string wa_id = 2;
+}
+
+message WhatsAppProfile {
+  string name = 1;
```
biglittlebigben: Do we need an extra message it it only contains a `name` field? Do we expect new fields to be added?

protobufs/connectors/whatsapp_webhook.proto
```diff
@@ -0,0 +1,87 @@
...
+}
+
+message WhatsAppError {
+  uint32 code = 1;
+  string message = 2;
+  string href = 3;
```
biglittlebigben: We usually use "url" in our APIs

protobufs/connectors/whatsapp_webhook.proto
```diff
@@ -0,0 +1,87 @@
...
+  WhatsAppErrorData error_data = 4;
+}
+
+message WhatsAppErrorData {
```
biglittlebigben: Do we need an extra message for a single string field? Do we expect this message to grow?

protobufs/connectors/whatsapp_webhook.proto
```diff
@@ -0,0 +1,87 @@
...
+  string details = 1;
+}
+
+message WhatsAppStatuses {
+  string id = 1;
+  google.protobuf.Timestamp timestamp = 2;
+  string type = 3; // always "call"
```
biglittlebigben: Why do we need this field?

protobufs/connectors/whatsapp_webhook.proto
```diff
@@ -0,0 +1,87 @@
...
+  string message = 2;
+  string href = 3;
+  WhatsAppErrorData error_data = 4;
+}
+
+message WhatsAppErrorData {
+  string details = 1;
+}
+
+message WhatsAppStatuses {
```
biglittlebigben: What is this message for?

protobufs/livekit_connectors.proto
```diff
@@ -0,0 +1,88 @@
...
+  
+  rpc ConnectWhatsAppCall(ConnectWhatsAppCallRequest) returns (ConnectWhatsAppCallResponse);
+
+  rpc RejectWhatsAppCall(RejectWhatsAppCallRequest) returns (RejectWhatsAppCallResponse);
+  
+  rpc ListWhatsAppCalls(ListWhatsAppCallsRequest) returns (ListWhatsAppCallsResponse);
+}
+
+message CreateWhatsAppParticipantRequest {
+  string from_phone_number_id = 1;
```
biglittlebigben: What is a phone number id? Is this either phone number or id? Or is this a WhatsApp contract?

protobufs/livekit_connectors.proto
```diff
@@ -0,0 +1,88 @@
...
+  optional string biz_opaque_callback_data = 3; // for logging purposes
+}
+
+message CreateWhatsAppParticipantResponse {
+}
+
+message DeleteWhatsAppParticipantRequest {
+  optional string whatsapp_call_id = 1;
+  optional livekit.connectors.WhatsAppCallWebhook call_terminate_webhook = 2;
+  optional bool is_call_rejected = 3;
```
biglittlebigben: Should this be a status enum?

protobufs/livekit_connectors.proto
anunaym14: It is a whatsapp contract, from what I've seen, it seems to be the phone number with country code but without the '+' prefix. But there might be other cases, better to use the same term as whatsapp.

protobufs/livekit_connectors.proto
anunaym14: Updated this, please have another look.

protobufs/livekit_connectors.proto
```diff
@@ -0,0 +1,170 @@
...
+  // Call ID sent by Meta
+  string whatsapp_call_id = 1;
+}
+
+message DeleteWhatsAppParticipantForTerminatedCall {
+  // Call ID sent by Meta
+  string whatsapp_call_id = 1;
+}
+
+message DeleteWhatsAppParticipantResponse {
```
biglittlebigben: should this be an `emptypb` until we have a field to return in this message? `emptypb` would make it backward and forward compatible

protobufs/livekit_connectors.proto
```diff
@@ -0,0 +1,138 @@
...
+  string whatsapp_call_id = 1;
+}
+
+message DisconnectWhatsAppCallRequest {
+  // Call ID sent by Meta
+  string whatsapp_call_id = 1;
+  string whatsapp_api_key = 2;
+}
+
+message DisconnectWhatsAppCallResponse {
```
biglittlebigben, line 71: Why not an emptypb here?

anunaym14: No particular reason, I saw both being used in the repo, empty responses as well as emptypb. No strong opinions either, both are fine.

PR #1206, opened 2025-09-11
Adding utilities for phonenumber functionality

nishadmusthafa:
1. DB prefixes
2. Helper utilities for area code and number type


sip/phonenumber.go
```diff
@@ -0,0 +1,54 @@
...
+	// Get the national number and extract first 3 digits (area code for US)
+	nationalNumber := phonenumbers.GetNationalSignificantNumber(num)
+	if len(nationalNumber) < 3 {
+		return ""
+	}
+	return nationalNumber[:3]
+}
+
+// DetermineNumberType determines the phone number type using the phonenumbers library
+func DetermineNumberType(phoneNumber string) string {
```
dennwc: This should probably return enum of some kind. Do we have it defined in protobufs somewhere?

PR #1208, opened 2025-09-12
Add push flag to data stream header

ladvoc:
This PR adds a `push` flag to the data stream header to facilitate the new behavior for `sendText` and `sendBytes` in which streams are automatically read and dispatched by the receiver:

WIP implementation in Rust, currently using an internal attribute in place of this flag:
- https://github.com/livekit/rust-sdks/pull/691

### DIFF
--- a/.changeset/hip-toes-matter.md
+++ b/.changeset/hip-toes-matter.md
@@ -1,0 +1,5 @@
+---
+"github.com/livekit/protocol": patch
+---
+
+Add push flag to data stream header


protobufs/livekit_models.proto
```diff
@@ -731,6 +731,7 @@ message DataStream {
     string topic = 3;
     string mime_type = 4;
     optional uint64 total_length = 5;  // only populated for finite streams, if it's a stream of unknown size this stays empty
+    bool push = 6;  // stream should be pushed to destination participants without requiring them to explicitly accept it.
```
lukasIO, line 734: it does more than "just" push though, right? It also internally collects the stream and only forwards the "complete" value. 

we could think about making `total_length` required for push cases as well.

ladvoc: Yes, that's correct; maybe it should be called something other than push? I agree, `total_length` should be required when "push" is enabled. I can also see possibly adding an option on the receiving end to reject pushed streams over a specified length.

ladvoc:
No longer being considered

PR #1216, opened 2025-09-17
Alex/adding display name

alexlivekit:


nishadmusthafa:
> I'm guessing `mage proto` just regenerates everything, including unmodified files.
> 
> I have had more recent versions of both protoc and protoc-gen-go-grpc and that resulted in some major changes in a lot of files, so this is the result of the minimal set of changes I managed to produce thus far.
> 
> Please let me know if there's a better way.

Usually I only use mage proto to make sure the generation is running fine and only commit the .proto files. There is a github action that will generate everything and create a commit for the autogenerated files.

protobufs/rpc/sip.proto
```diff
@@ -105,7 +105,15 @@ message InternalCreateSIPParticipantRequest {
...
-  // NEXT ID: 31
+  // Optional display name for the 'From' SIP header.
+  //
+  // Cases:
+  // 1) Unspecified: Use legacy behavior - display name will be set to be the caller's number. 
+  // 2) Empty string: Do not send a display name, which will result in a CNAM lookup downstream.
+  // 3) Non-empty: Use the specified value as the display name.
+  optional string display_name = 31;
+
+  // NEXT ID: 32
```
nishadmusthafa, line 116: @dennwc perhaps for another change, but couldn't we reuse CreateSIPParticipantRequest here ?

dennwc: Not directly, but yes, this is on my TODO list as well.

Sharing one structure likely won't work well, because internal has a different representation of a few things (e.g. feature flags).

I was thinking that we should slice this structure into a few smaller ones. Everything related to media setup can be one structure, shared for internal Create, public Create and also for inbound trunks. Similarly, there are fields related to how SIP should join LK room, this could be another structure shared for Create and EvaluateDispatch. The plan is also to allow storing these objects directly into the DB, so that we don't have to flatten/unflatten from SQL and add columns each time we add another flag that is never used as a filter for DB queries.

protobufs/rpc/sip.proto
```diff
@@ -105,7 +105,15 @@ message InternalCreateSIPParticipantRequest {
...
   bool wait_until_answered = 29;
 
-  // NEXT ID: 31
+  // Optional display name for the 'From' SIP header.
+  //
+  // Cases:
+  // 1) Unspecified: Use legacy behavior - display name will be set to be the caller's number. 
+  // 2) Empty string: Do not send a display name, which will result in a CNAM lookup downstream.
+  // 3) Non-empty: Use the specified value as the display name.
+  optional string display_name = 31;
```
dennwc, line 114: In theory, we could avoid `optional` here, since we can populate `display_name` to a number in `NewCreateSIPParticipantRequest`. The only concern would be that the backend still cannot know if it should follow the old logic or a new one.

dennwc (approved):

PR #1217, opened 2025-09-18
Extending models to support MP3 audio encoding in egress

milos-lk:


protobufs/livekit_models.proto
```diff
@@ -68,6 +68,7 @@ enum AudioCodec {
   DEFAULT_AC = 0;
   OPUS = 1;
   AAC = 2;
+  CODEC_MP3 = 3;
```
milos-lk: Couldn't leave it as MP3 (clashes with the encoded file type enum value - and protobuf doesn't allow that because of c++). If you have any better suggestions please let me know.

biglittlebigben: Yeah, we messed up on this one. We now usually prefix all our constants with the type name. If we go the route of the "MP3" filetype, maybe we should use AC_MP3.

protobufs/livekit_egress.proto
```diff
@@ -157,6 +157,7 @@ enum EncodedFileType {
   DEFAULT_FILETYPE = 0; // file type chosen based on codecs
   MP4  = 1;
   OGG = 2;
+  MP3 = 3;
```
biglittlebigben, line 160: The exact/pedantic file format for what folks usually refer as "mp3" as RAW, as there is no muxer (it's just a succession of audio frames, like .AAC). This is most likely too confusing for non experts so'MP3' may be the better option indeed. However:
- It creates overlap the issue with the codec below.
- We should need to create an entry for every RAW format we may eventually support (AAC/ADTS, H264 NALUs, etc)

milos-lk: I am slightly leaning towards having these explicit file types here. It's a bit more verbose but I also think it makes it less confusing. will update the constant name below.

biglittlebigben (approved):

milos-lk:
ugh - there are some conflicts with the autogenerated files which I tried to resolved but somehow messed up - will close this PR and create a new PR with this 2 liner.

PR #1223, opened 2025-09-22
fix(data_channel_rpc): avoid using DeadlineExceeded for PSRPC errors

anunaym14:


utils/data_channel_rpc.go
```diff
@@ -64,9 +64,9 @@ func (e *DataChannelRpcError) PsrpcError() psrpc.Error {
 	case DataChannelRpcApplicationError:
 		return psrpc.NewErrorf(psrpc.Internal, e.Message, "data", e.Data)
 	case DataChannelRpcConnectionTimeout:
-		return psrpc.NewErrorf(psrpc.DeadlineExceeded, e.Message, "data", e.Data)
+		return psrpc.NewErrorf(psrpc.Canceled, e.Message, "data", e.Data)
 	case DataChannelRpcResponseTimeout:
-		return psrpc.NewErrorf(psrpc.DeadlineExceeded, e.Message, "data", e.Data)
```
boks1971, line 69: how does that translate to API error?  Feels like `DeadlineExceeded` is the right one and that will translate to `Timeout`

boks1971 (approved):

utils/data_channel_rpc.go
anunaym14: `"error": "twirp error cancelled: twirp error unknown: Response timeout%!`

utils/data_channel_rpc.go
```diff
@@ -64,9 +64,9 @@ func (e *DataChannelRpcError) PsrpcError() psrpc.Error {
 	case DataChannelRpcApplicationError:
 		return psrpc.NewErrorf(psrpc.Internal, e.Message, "data", e.Data)
 	case DataChannelRpcConnectionTimeout:
-		return psrpc.NewErrorf(psrpc.DeadlineExceeded, e.Message, "data", e.Data)
+		return psrpc.NewErrorf(psrpc.Canceled, e.Message, "data", e.Data)
```
paulwe: these args don't make sense for errorf (args 1...n are sprintf template and args for the error message)

PR #1229, opened 2025-09-25
Data track signaling

ladvoc:


protobufs/livekit_models.proto
```diff
@@ -245,6 +245,30 @@ message TrackInfo {
   BackupCodecPolicy backup_codec_policy = 20;
 }
 
+message DataTrackInfo {
+  string sid = 1;
+
+  // 16-bit, SFU-assigned identifier used to associate packets with the track, unique per room.
+  uint32 handle = 2;
```
boks1971: Typed out a long message and my laptop power went out :-( 

Trying again
I think it is better for client (i. e. publisher of data track) to generate this. Reason being generating this for distributed room would be complicated. Something that would be good to avoid.

SFU can generate a different ID for subscriber side. And communicate that to subscribing client using one of two options
1. A signal message to indicate mapping of handle ID -> track SID when a subscription happens. One issue is some data track packets arriving before signal messages arrives. Given that late joining subscriber will anyway not get all packets, this can be thought of a slightly delayed start.
2. Send as a header extension in the first 10 - 20 packets. The risk is all these packets getting lost. For media, this is usually handled by sending end adding that extension till the receiver sends an RTCP report. We can do this as this will introduce two important things for data tracks (extensions and feedback report). Feedback report can also be used for NACK + retransmission down the line. For this case, the feedback report could be something like RTCP Receiver Report.

If we do it, `handle` can be removed from `DataTrackInfo` as it does not have any use.

protobufs/livekit_models.proto
```diff
@@ -246,6 +246,27 @@ message TrackInfo {
   BackupCodecPolicy backup_codec_policy = 20;
 }
 
+message DataTrackInfo {
+  string sid = 1;
```
boks1971: let's add `handleID` to this model. I think it is useful to have it in the model.

protobufs/livekit_rtc.proto
```diff
@@ -479,6 +539,11 @@ message RequestResponse {
     QUEUED = 4;
     UNSUPPORTED_TYPE= 5;
     UNCLASSIFIED_ERROR = 6;
+    DATA_TRACK_INVALID_HANDLE = 7;
+    DATA_TRACK_INVALID_NAME = 8;
+    DATA_TRACK_INVALID_MIME_TYPE = 9;
+    DATA_TRACK_NAME_TAKEN = 10;
+    reserved 11 to 15; // Reserved for future data track errors.
```
boks1971: don't need any reserved stuff. This can just grow as needed.

boks1971: And all the data track specific name can be made generic by dropping the `DATA_TRACK_` prefix. We can re-use if it fits a different scenario.

boks1971: For example, an `UnpublishDataTrackRequest` could get a reply with the existing `NOT_FOUND` reason.

protobufs/livekit_rtc.proto
```diff
@@ -160,6 +172,43 @@ message AddTrackRequest {
...
+  // This must be a valid MIME type as defined by RFC 2046.
+  string mime_type = 3;
+
+  // Method used for end-to-end encryption (E2EE) on frame payloads.
+  Encryption.Type encryption = 4;
+}
+
+message PublishDataTrackResponse {
+  // Handle of the track that was published.
+  uint32 handle = 1;
```
boks1971: If `handle` in included in the model, this field can be dropped from this message.

protobufs/livekit_rtc.proto
```diff
@@ -160,6 +172,43 @@ message AddTrackRequest {
...
+}
+
+message UnpublishDataTrackRequest {
+  // Handle of the track to unpublish.
+  uint32 handle = 1;
+}
+
+message UnpublishDataTrackResponse {
+  // Handle of the track that was unpublished.
+  uint32 handle = 1;
```
boks1971: Can change this to return `DataTrackInfo` so that you have both `handle` (assuming it is added to `DataTrackInfo` model) and `Sid` also.

boks1971: I am going to make a branch off your branch locally as I am starting to work on some initial signalling handling around this. Please let me know if you have objections to the changes I have suggested. We can discuss and pick out what works.

protobufs/livekit_rtc.proto
```diff
@@ -160,6 +172,43 @@ message AddTrackRequest {
...
+}
+
+message DataTrackSubscriberHandles {
+  // Mapping from data track SIDs to handles subscribers will see on incoming packets.
+  map<string, uint32> handles = 1;
```
boks1971: also proposing changing this to `map<uint32, string>`.

The idea being
- receives packets for unknown handle
- look up map indexed by handle
- if a mapping is found, ensure `Sid` can be found in remote participants.
- create a new subscribed data track for thatn handle

protobufs/livekit_models.proto
ladvoc: Added. I called it `pub_handle` to disambiguate from the handle seen by subscribers.

protobufs/livekit_rtc.proto
ladvoc: Fixed. I got this backwards; mapping from handle to SID is what is needed on the client side.

protobufs/livekit_rtc.proto
```diff
@@ -479,6 +538,10 @@ message RequestResponse {
     QUEUED = 4;
     UNSUPPORTED_TYPE= 5;
     UNCLASSIFIED_ERROR = 6;
+    INVALID_HANDLE = 7;
+    INVALID_NAME = 8;
+    INVALID_MIME_TYPE = 9;
```
boks1971: @ladvoc Thinking about the checks. What conditions would trigger these two? `INVALID_NAME` and `INVALID_MIME_TYPE`?
- Is name more like length check?
- What mime types are supported? I am guessing we are not going to do any processing using this unlike media where it is used for things like parsing some bits of packet to figure out if it is a key frame etc.

ladvoc: - A valid name has to have a length above zero and below a maximum (TBD). Having the maximum length will allow this field to be stack allocated on embedded.
- A valid MIME type must follow the basic formatting rules of RFC 2046 (top-level type, separator, and contain valid characters). This will just be for informational purposes now, but it should be validated for future use.

protobufs/livekit_rtc.proto
```diff
@@ -479,6 +534,9 @@ message RequestResponse {
     QUEUED = 4;
     UNSUPPORTED_TYPE= 5;
     UNCLASSIFIED_ERROR = 6;
+    INVALID_HANDLE = 7;
+    INVALID_NAME = 8;
+    NAME_TAKEN = 9;
```
boks1971: Sorry @ladvoc . Upon further thought, can we just collapse this to one reason `INVALID_PARAM`. Thinking we can keep it generic and provide a more detailed reason in `message` field below.

boks1971: One more sorry. In addition to the above, can you please add another generic one called `DUPLICATE`. Would like to use that when a publish request with a currently published handle is re-requested. 

Will use the same when `name` matches an existing published track's name too.

protobufs/livekit_rtc.proto
```diff
@@ -160,6 +172,38 @@ message AddTrackRequest {
   repeated AudioTrackFeature audio_features = 17;
 }
 
+message PublishDataTrackRequest {
+  // Client-assigned, 16-bit identifier that will be attached to packets sent by the publisher.
+  // This must be non-zero and unique for each data track published by the publisher.
+  uint32 pub_handle = 1;
+
+  // Human-readable identifier (e.g., `geoLocation`, `servoPosition.x`, etc.), unique per publisher.
+  string name = 2;
```
boks1971, line 184: is this mandatory? I am adding a check on server side for non-zero length and length <=256. Wondering if zero length should be checked.

protobufs/livekit_rtc.proto
boks1971: Hmm, wondering if we should just use `NOT_ALLOWED` with a proper reason message for the `DUPLICATE` case. I will just use that. No need for new one.

protobufs/livekit_rtc.proto
```diff
@@ -231,6 +275,17 @@ message UpdateSubscription {
   repeated ParticipantTracks participant_tracks = 3;
 }
 
+message UpdateDataSubscription {
+  message Update {
+    string sid = 1;
```
boks1971: ~~can you please add `participant_identity`? Otherwise, have to search all participants in the room for that track SID.

For media, we are using `participant_sid`. But, we have been moving away from SID and using identity. I think for this one, we can use the identity.~~

Strike that, server uses a track manager per room and can look it up.

boks1971: Not sure why GH markdown strike through does not work.

protobufs/livekit_rtc.proto
ladvoc: It’s not strictly mandatory, but I’d recommend disallowing zero-length names. An empty string doesn’t really serve as a meaningful track name and could cause confusion or edge cases down the line. A minimum length of 1 character along with your existing max length check seems reasonable to me.

protobufs/livekit_rtc.proto
ladvoc: For the client side, it is preferable if the failure reason can be mapped to a discrete enum case so we can have strongly typed errors. Please let me know your thoughts on this.

protobufs/livekit_rtc.proto
boks1971: okay, then it kind of becomes mandatory right?

protobufs/livekit_rtc.proto
boks1971: Ok, in that case, can you please an entry for `DUPLICATE` as I was planning to use it for handle/name collision of publish request with an existing track.

protobufs/livekit_rtc.proto
ladvoc: Sorry, I misread. Yes, the field is mandatory.

boks1971: Cool. can you please add a comment to that field in the request definition?

protobufs/livekit_rtc.proto
boks1971: Maybe, two here also
- `DUPLICATE_HANDLE`
- `DUPLICATE_NAME`
if you are looking for possible causes of duplicate.

ladvoc: Yeah, that's probably ideal since duplicate handle is an internal error vs. duplicate name is a client error. I will change it.

protobufs/livekit_rtc.proto
ladvoc: So we should just include the track SID?

boks1971: Actually, leave this as is for now @ladvoc . Yeah, I think track SID might be enough, but let me do some work around it and come back if we can leave in identity or remove it. Apologies for the back and forth.

ladvoc: No problem, back and forth is the way to get it right the first time 🙂.

PR #1230, opened 2025-09-26
update api

frostbyte73:


protobufs/cloud_replay.proto
```diff
@@ -49,7 +50,7 @@ message PlaybackResponse {
 
 message SeekRequest {
   string playback_id = 1;
-  int64 start_time = 2; // initial cursor
+  int64 ts = 2; // timestamp (ms)
```
biglittlebigben: `ts` may be a bit too succint for folks who haven't interacted with media pipelines. It's also not obvious what the "0" reference is (the room start, I suppose?). Maybe "room_time_offset", or "seek_time", or "seek_offset", or something similar?

protobufs/cloud_replay.proto
```diff
@@ -19,7 +19,8 @@ service Replay {
 }
 
 message ListReplaysRequest {
-  string room_name = 1; // optional
+  string room_name = 1;     // optional
+  int64 started_before = 2; // optional (for pagination), unix ms
```
biglittlebigben: We already have at least 2 schemes for pagination (Pagination and TokenPagination). Do we have a strong reason to add a 3rd one?

frostbyte73: no, we should stick to one of those. where are those being used?

biglittlebigben: Both are used in the SIP APIs. I vaguely remember @dennwc mentioning that TokenPagination was preferred these days, at least when the pagination iterator semantics can be hidden from the caller. We should check with him though.

boks1971 (approved):

milos-lk (approved):

PR #1232, opened 2025-09-29
Making the sip refer to compliant to the RFC 

nishadmusthafa:
Using the <> braces that enables more complex strings to be used as refer-to. Adding validation/tests for the same

### DIFF
--- a/livekit/sip.go
+++ b/livekit/sip.go
@@ -698,6 +698,28 @@ func (p *TransferSIPParticipantRequest) Validate() error {
 	if p.TransferTo == "" {
 		return errors.New("missing transfer to")
 	}
+
+	// Validate TransferTo URI format and ensure RFC compliance
+	validURIRegex := regexp.MustCompile(`^(sip:|tel:)`)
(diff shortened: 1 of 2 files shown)


livekit/sip.go
```diff
@@ -698,6 +698,28 @@ func (p *TransferSIPParticipantRequest) Validate() error {
 	if p.TransferTo == "" {
 		return errors.New("missing transfer to")
 	}
+
+	// Validate TransferTo URI format and ensure RFC compliance
+	validURIRegex := regexp.MustCompile(`^(sip:|tel:)`)
```
dennwc: Compilation takes time, so it must be defined on the package level.

livekit/sip.go
```diff
@@ -698,6 +698,28 @@ func (p *TransferSIPParticipantRequest) Validate() error {
...
+	angleBracketRegex := regexp.MustCompile(`^<.*>$`)
```
dennwc: Maybe just check with `strings.HasPrefix` and `HasSuffix` instead of both regexps. Looks like an overkill.

PR #1235, opened 2025-09-30
add server region to agent info and create for log handling

real-danm:


protobufs/livekit_cloud_agent.proto
```diff
@@ -51,6 +51,7 @@ message CreateAgentResponse {
     string version = 4;
     string presigned_url = 5;
     string tag = 6;
+    repeated string server_regions = 7;
```
yoonsio, line 54: nit: any comments here would be super helpful. Is there any difference between regions vs server_regions?

PR #1238, opened 2025-10-04
Added more scrutinizing SIP header validation warnings

alexlivekit:
I still need to test this, but this is the rough spirit here.

The few qualms that I still have with this:
- This introduces logging into `protocol/livekit`. I'm not sure I like this.
- The validation on some of these feels a little like overkill. I've tired it with a few other methods (map, regex, static & dynamic arrays). A dynamic array is almost as fast as this, and while a static array is faster, it isn't thread/goroutine friendly.
- I'm pretty sure this is a decent approach, but let me know if there are better ways out there.

### DIFF
--- a/livekit/sip_validation.go
+++ b/livekit/sip_validation.go
@@ -23,49 +23,49 @@ import (
 
 // RFC 3261 compliant validation functions for SIP headers and messages
 
-type AllowedCharacters struct {
+type allowedCharacters struct {
 	ascii [127]bool
 	utf8  bool
 }
 
-func NewAllowedCharacters() *AllowedCharacters {
-	return &AllowedCharacters{}
+func NewAllowedCharacters() *allowedCharacters {
+	return &allowedCharacters{}
 }
 
-func (a *AllowedCharacters) AddUTF8() error {
+func (a *allowedCharacters) AddUTF8() error {
 	a.utf8 = true
 	return nil
 }
 
-func (a *AllowedCharacters) AddNumbers() error {
+func (a *allowedCharacters) AddNumbers() error {
 	for r := '0'; r <= '9'; r++ {
 		a.ascii[r] = true
 	}
 	return nil
 }
 
-func (a *AllowedCharacters) AddLowercaseASCII() error {
+func (a *allowedCharacters) AddLowercaseASCII() error {
 	for r := 'a'; r <= 'z'; r++ {
 		a.ascii[r] = true
 	}
 	return nil
 }
 
-func (a *AllowedCharacters) AddUppercaseASCII() error {
+func (a *allowedCharacters) AddUppercaseASCII() error {
 	for r := 'A'; r <= 'Z'; r++ {
 		a.ascii[r] = true
 	}
 	return nil
 }
 
-func (a *AllowedCharacters) AddPrintableLienarASCII() {
+func (a *allowedCharacters) AddPrintableLienarASCII() {
 	// Anything between 0x20 and 0x7E
 	for i := 0x20; i <= 0x7E; i++ {
 		a.ascii[i] = true
 	}
 }
 
-func (a *AllowedCharacters) Add(chars string) error {
+func (a *allowedCharacters) Add(chars string) error {
 	for _, char := range chars {
 		if int(char) >= len(a.ascii) {
 			return fmt.Errorf("char %d out of range, consider explicilty adding utf8 characters", char)
@@ -75,7 +75,7 @@ func (a *AllowedCharacters) Add(chars string) error {
 	return nil
 }
 
-func (a *AllowedCharacters) Remove(chars string) error {
+func (a *allowedCharacters) Remove(chars string) error {
 	for _, char := range chars {
 		if int(char) >= len(a.ascii) {
 			return fmt.Errorf("char %d out of range, consider explicilty adding utf8 characters", char)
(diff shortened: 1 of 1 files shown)


livekit/sip.go
```diff
@@ -367,6 +391,15 @@ func (p *SIPInboundTrunkInfo) Validate() error {
 	if err := validateHeaderValues(p.AttributesToHeaders); err != nil {
 		return err
 	}
+	if err := validateHeaders(p.Headers); err != nil {
+		logger.Warnw("Header validation failed for Headers field", err)
+		// No error, just a warning for SIP RFC validation for now
+	}
+	// Don't bother with HeadersToAttributes. If they're invalid, we just won't match
```
dennwc, line 399: That's also an issue from the user's point of view. For example, adding an extra space may ignore the header mapping silently.

livekit/sip.go
```diff
@@ -367,6 +391,15 @@ func (p *SIPInboundTrunkInfo) Validate() error {
 	if err := validateHeaderValues(p.AttributesToHeaders); err != nil {
 		return err
 	}
+	if err := validateHeaders(p.Headers); err != nil {
+		logger.Warnw("Header validation failed for Headers field", err)
```
dennwc, line 396: We should pass the logger here then. This way we could print it in the context of the request, not globally.

livekit/sip_validation.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+// validateDisplayName validates a display name in name-addr format
+func validateDisplayName(displayName string) error {
+	if displayName == "" {
+		return nil
+	}
+
+	// Check if display name is quoted
+	if strings.HasPrefix(displayName, "\"") && strings.HasSuffix(displayName, "\"") {
```
dennwc:
```suggestion
	if strings.HasPrefix(displayName, `"`) && strings.HasSuffix(displayName, `"`) {
```

livekit/sip_validation.go
```diff
@@ -0,0 +1,395 @@
...
+		// Quoted display name - basic validation
+		quoted := displayName[1 : len(displayName)-1]
+		// Check for proper escaping
```
dennwc: We could probably do `strconv.Unquote`, they probably share some assumptions, at least.

livekit/sip_validation.go
```diff
@@ -0,0 +1,395 @@
...
+		if r == '<' || r == '>' || r == '"' || r == '\\' || r == '&' {
+			return errors.New("unquoted display name contains special characters")
+		}
+	}
+
+	return nil
+}
+
+// validateURI validates URIs that can appear in name-addr format
+func validateURI(uri string) error {
```
dennwc, line 344: Maybe we could reuse the parser from sipgo for validation?

alexlivekit: Tempting!

The reason I didn't is I was under the impression we don't want more dependencies in `protocol`.
Additionally, do we want `livekit/sipgo`, `/emiago/sipgo@v0.24.1` (the same verion we have our own depend on), or `emiago/sipgo@latest`. That last one feels like a problem waiting to happen. Of the first two, I'd probably go with `livekit/sipgo` just to make sure we don't have to coordinate versions between those.
On the other hand, the only `livekit/` module `protocol` depends on is mageutil, and this feels like a big change.

livekit/sip.go
alexlivekit: How much validation do we want here in protocol? I could add this, but if you have a sip headers named `X-my-custom-header` and you specify `HeadersToAttributes["X-my-custom -header"] = "my.attribute"`, I don't know that we need to protect users from themselves on this, no?

livekit/sip_validation.go
dennwc: Yeah, you are right, adding this dependency to the protocol is a bit too much. Maybe we could do basic validation here and push the rest to the actual SIP server? It can validate these properly with a parser.

alexlivekit: that we can do no problem.

livekit/sip.go
dennwc: I'd say we should just guard against silly mistakes like adding spaces, punctuation, etc. This validation is reused for the CLI as well, so it's convenient when it detects these issues early without sending anything to the server. 

Checking against protected headers is a good idea too. Although if we want to change the list, it will require a pretty long update process if it's defined in the protocol.

livekit/sip.go
alexlivekit: That's the thing, this is protocol/livekit. I don't even know how to pass loggers in here.

The only thing I see here is logger.SetLogger, which we would need to explicitly call to make these messages available.

Plus, the point of these errors is to see what and how many failures we get, not necessarily attribute it to a specific client or project.

Got any ideas on how to pass a logger here?

alexlivekit: Rephrasing Denys: Since the validation is something we added on top of PB stuff, and that's getting called explicitly (either via an interface or directly), and the idea was to pass a logger to that.

However, given the scope of that change (either break existing api or add another `ValidateWithLogger()`) and start using that, it seemed like a saner choice to just global-log and later turn into an error.

livekit/sip_validation.go
```diff
@@ -0,0 +1,359 @@
...
+import (
+	"errors"
+	fmt "fmt"
+	"strconv"
+	"strings"
+)
+
+// RFC 3261 compliant validation functions for SIP headers and messages
+
+type AllowedCharacters struct {
```
dennwc: This should be private.
```suggestion
type allowedCharacters struct {
```

dennwc (approved):
LGTM after the last nit is addressed. Thanks!

Also, you'll probably need to downgrade your proto generator to not cause these large diffs.

livekit/sip_validation.go
alexlivekit: We're sure we would not want these in sip to validate better?

dennwc: We could always re-export it later. But SIP can use a real parser, so probably not.

PR #1243, opened 2025-10-07
[WIP] filter webhook

barrust:


protobufs/livekit_models.proto
```diff
@@ -757,13 +757,18 @@ message DataStream {
   }
 }
 
+message FilterParams {
+  repeated string include_events = 1;
+  repeated string exclude_events = 2;
```
kuba--, line 762: I wonder if it makes sense to have `oneof { include, exclude }`

kuba--: IMO, setup include & exclude does not make sense.

barrust: I thought that it could be simplified, but the logic was there before: 

https://github.com/livekit/protocol/blob/1f82faed9f78220b0c2f55e0bc308143940aae64/webhook/notifier.go#L187-L190

boks1971: Please feel free to change what makes sense. I did this as part of a different effort for webhooks processing. As I did not put it in `protobuf`, did not have the `OneOf` option.

BTW, I am getting an error in `cloud` because of this. Will try to fix it.

PR #1250, opened 2025-10-11
Adding a trunk info field that will give us more details

nishadmusthafa:
on what kind of trunk is used so decisions can be made on the basis of that during the call flow


protobufs/livekit_sip.proto
```diff
@@ -191,6 +191,18 @@ enum SIPMediaEncryption {
...
+  ProviderType type = 3;
+}
+
+enum ProviderType {
+  PROVIDER_TYPE_UNKNOWN = 0;
+  PROVIDER_TYPE_INTERNAL = 1;  // Internally implemented
+  PROVIDER_TYPE_EXTERNAL = 2;  // Vendor provided
+}
+
 message SIPTrunkInfo {
```
dennwc, line 206: This is a deprecated message. You'll have to update SIPInboundTrunk and SIPOutboundTrunk instead.

protobufs/rpc/io.proto
```diff
@@ -105,6 +105,7 @@ message GetSIPTrunkAuthenticationResponse {
   string sip_trunk_id = 4;
   // Used in Cloud only
   string project_id = 5;
+  livekit.SIPTrunkInfo trunk_info = 6;
```
dennwc: Same here, we can't use a deprecated message going forward.

protobufs/livekit_sip.proto
nishadmusthafa: Ahh okay. However, I actually want something that isn't specifically inbound or outbound. The goal is pass this in SipCallInfo and other messaging during and after call flow. How about something new call TrunkDetails ??

PR #1254, opened 2025-10-15
Add twilio connector

cnderrauber:


protobufs/livekit_connector_twilio.proto
```diff
@@ -0,0 +1,51 @@
...
+option go_package = "github.com/livekit/protocol/livekit";
+option csharp_namespace = "LiveKit.Proto";
+option ruby_package = "LiveKit::Proto";
+
+import "livekit_agent_dispatch.proto";
+
+message ConnectTwilioCallRequest {
+  // The Media Stream Direction for the call
+  TwilioStreamDirection stream_direction = 1;
+  // What LiveKit room should this call be connected too
```
boks1971: :nit: `connected to`

protobufs/livekit_connector_twilio.proto
```diff
@@ -0,0 +1,51 @@
...
+  string room_name = 2;
+  // Optional agents to dispatch the call to
+  repeated RoomAgentDispatch agents = 3;
+  // Optional identity of the participant in LiveKit room
+  string participant_identity= 4;
```
boks1971, line 38: can this be marked `optional` so that it is clear from protocol definition itself? Same for the below optional fields.

protobufs/livekit_connector_twilio.proto
```diff
@@ -0,0 +1,51 @@
...
+  // Country where the call terminates as ISO 3166-1 alpha-2 (https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2). This will be used by the livekit infrastructure to route calls.
+  string destination_country = 8;
+}
+
+message ConnectTwilioCallResponse {
+  // The websocket URL which twilio media stream will connect to
+  string connect_url = 1;
+}
+
+enum TwilioStreamDirection {
```
boks1971: can we embed this in `ConnectTwilioCallRequest` so that it is in that context as it is used only there?

boks1971 (approved):

protobufs/livekit_connector_twilio.proto
anunaym14: We decided not to mark things optional because for the generated Go protobufs, it becomes *string, which is ugly. For whatsapp connector (as well as SIP), things are treated as optional by passing an empty string

anunaym14: @cnderrauber flagging this since you've already updated it

protobufs/livekit_connector_twilio.proto
```diff
@@ -0,0 +1,52 @@
...
+
+message ConnectTwilioCallRequest {
+
+  enum TwilioStreamDirection {
+    TWILIO_STREAM_UNIDIRECTIONAL = 0;
+    TWILIO_STREAM_BIDIRECTIONAL = 1;
+  }
+
+  // The Media Stream Direction for the call
+  TwilioStreamDirection stream_direction = 1;
```
anunaym14: We should rename this to `twilio_stream_direction`, we kept the pattern for whatsapp too to mark anything non-livekit explicitly

protobufs/livekit_connector_twilio.proto
cnderrauber: revert to be consistent with other connector(whatsapp).

cnderrauber:
> LGTM! I'm just not certain whether we want to expose the capability to open unidirectional streams, to keep things simple. Bidirectional streams are the use case we've heard about so far.
> 
> cc @davidzhao for review as well.

Remove the stream direction field as we will only support bidirectional stream now, we can add it back if unidirectional is required in future. Add call direction field to distinguish inbound/outbound calls.

protobufs/livekit_connector_twilio.proto
```diff
@@ -0,0 +1,52 @@
...
+package livekit;
+option go_package = "github.com/livekit/protocol/livekit";
+option csharp_namespace = "LiveKit.Proto";
+option ruby_package = "LiveKit::Proto";
+
+import "livekit_agent_dispatch.proto";
+
+message ConnectTwilioCallRequest {
+
+  enum TwilioCallDirection {
```
biglittlebigben, line 26: Inbound vs outbound can be a confusing thing for telephony APIs as they depend on the view point. We should add some doc to define precisely what either one means. (Twilio to livekit or livekit to twilio)

PR #1268, opened 2025-10-22
Destination region in create sip participant

nishadmusthafa:


livekit/sip.go
```diff
@@ -685,6 +686,29 @@ func (p *CreateSIPParticipantRequest) Validate() error {
 		// TODO: Validate display name doesn't contain invalid characters
 	}
 
+	// Validate destination if provided
+	if p.Destination != nil {
```
dennwc: This could be pushed down to the `Destination` method:
```go
func (d *Destination) Validate() error {
   if d == nil {
      return nil
   }
```
(this is safe in Go)

livekit/sip.go
```diff
@@ -11,6 +11,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	"github.com/flaticols/countrycodes"
```
dennwc: Will [golang.org/x/text](https://pkg.go.dev/golang.org/x/text@v0.30.0/language) help validate it instead? Would be not to introduce a new dependency.

livekit/sip.go
```diff
@@ -685,6 +686,29 @@ func (p *CreateSIPParticipantRequest) Validate() error {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (d *Destination) Validate() error {
+	// Rule 1: If city is specified, country must be specified
+	if d.City != "" && d.Country == "" {
```
dennwc:
```suggestion
	if d.City != "" && d.Country == "" && d.Region == "" {
```
I guess it's okay if either region or country is set.

livekit/sip.go
nishadmusthafa: I tried it out. It doesn't work great. It is a bit permissive and for 2 letter country codes, it parses it without errors. I think this may be because of countries that don't exist yet ?

livekit/sip.go
```diff
@@ -685,6 +686,49 @@ func (p *CreateSIPParticipantRequest) Validate() error {
 		// TODO: Validate display name doesn't contain invalid characters
 	}
 
+	// Validate destination if provided
+	p.Destination.Validate()
```
dennwc: Should return an error.

PR #1272, opened 2025-10-24
Adding destination to InternalCreateSIPParticipantRequest

nishadmusthafa:


protobufs/rpc/sip.proto
```diff
@@ -113,7 +113,10 @@ message InternalCreateSIPParticipantRequest {
   // 3) Non-empty: Use the specified value as the display name.
   optional string display_name = 31;
 
-  // NEXT ID: 32
+  // Destination information for routing decisions
+  optional livekit.Destination destination = 32;
```
dennwc:
```suggestion
  livekit.Destination destination = 32;
```
It;s okay to omit optional here. Field above is a special case where we really need that annotation.

PR #1277, opened 2025-10-29
fix(connector): update protos for api version, errors, cleanup

anunaym14:


protobufs/livekit_connector_whatsapp.proto
```diff
@@ -62,15 +64,11 @@ message DisconnectWhatsAppCallResponse {
...
 message ConnectWhatsAppCallRequest {
-  // The number of the business that is conencting the call
-  string whatsapp_phone_number_id = 1;
-  // The API key of the business that is connecting the call
-  string whatsapp_api_key = 2;
   // Call ID sent by Meta
-  string whatsapp_call_id = 3;
+  string whatsapp_call_id = 1;
   // The call connect webhook comes with SDP from Meta
   // It is the answer SDP for a business initiated call
```
anunaym14, line 74: Oversight on my end to leave the extra fields, shall I keep the fieldIDs for rest of them the same as before or is it fine to update them like this? no server sdks have connector support right now.

paulwe (approved):

protobufs/livekit_connector_whatsapp.proto
```diff
@@ -45,6 +45,8 @@ message DialWhatsAppCallRequest {
   map<string, string> participant_attributes = 10;
   // Country where the call terminates as ISO 3166-1 alpha-2 (https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2). This will be used by the livekit infrastructure to route calls.
   string destination_country = 11;
+  // Meta cloud API version 
+  float whatsapp_cloud_api_version = 12;
```
anunaym14: API versions are like v23.0/v24.0

I don't think it's required for the user to add 'v' since that's the convention, for the rest, is string better or float?

biglittlebigben: String. We should assume this is/may become a sem ver. We should document in more details what to set there (ie 24.0 as of today). Is this a required field? If so, we should say so, and probably move the field higher in message (after whatsapp_api_key?). If not, is there a sensible default?

protobufs/livekit_connector_whatsapp.proto
biglittlebigben: It is still fine to change the ids If no SDKs with this version of the protocol has been released.

protobufs/livekit_connector_whatsapp.proto
```diff
@@ -102,7 +100,8 @@ message AcceptWhatsAppCallRequest {
   map<string, string> participant_attributes = 11;
   // Country where the call terminates as ISO 3166-1 alpha-2 (https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2). This will be used by the livekit infrastructure to route calls.
   string destination_country = 12;
-  
+  // Meta cloud API version
+  float whatsapp_cloud_api_version = 13;
```
biglittlebigben: Same remark about documentation.

protobufs/livekit_connector_whatsapp.proto
anunaym14: There's no sensible default, it has to be same as the version set in the Meta/Facebook Developer Dashboard

which makes it a required field

protobufs/livekit_connector_whatsapp.proto
```diff
@@ -29,6 +29,8 @@ message DialWhatsAppCallRequest {
   string whatsapp_to_phone_number = 2;
   // The API key of the business that is initiating the call
   string whatsapp_api_key = 3;
+  // WhatsApp Cloud API version, 23.0 and 24.0 are supported as of today
```
biglittlebigben: My feedback was confusing, sorry. I don't think we should have "as of today" in the comment as it doesn't convey any information (when is "today"). We should also explicitly state that the field is required. (We should probably also state that whatsapp_phone_number_id, whatsapp_to_phone_number and whatsapp_api_key are required, but I think that it's more intuitive that these are required than the whatsapp_api_key where folks may believe we have a default).

anunaym14: Sorry for the confusion, updated it, please check again

PR #1278, opened 2025-10-30
Allow webhook notifiers to skip TLS certificate verification

anandharaj-dotworld:


biglittlebigben:
Can you provide context as to why this is needed?

anandharaj-dotworld:
@biglittlebigben We’re using LiveKit webhooks with a locally hosted HTTPS endpoint that uses a self-signed certificate. To ensure webhook deliveries keep working in this setup, we need the notifier to either:

Allow skipping TLS verification, or

Provide a way to attach and trust a custom root CA.

This lets local or development environments using self-signed certificates receive webhook events without failing TLS validation, while still maintaining secure defaults for production environments.

davidzhao:
You should be able to attach custom root certs via env var: `SSL_CERT_DIR`. that'd be the recommended approach instead of disabling verification

anandharaj-dotworld:
@davidzhao How can i use this with livekit docker image.

PR #1279, opened 2025-10-30
Cement SIP header validation

alexlivekit:
Now that the new validation has been out for a bit, but only as a warning, this is a follow-up.

**Behavior changes:**
1. Fail on new validation failing, no longer just a warning.
2. Dropped global logs.
3. Validate proper quoting for display name.

**Validation changes:**
1. Added the '\t' character previously omitted.
5. Allow UTF-8 characters in header values (Spotted one customer with U-8217, or 舗 which apparently means shop).
6. Allow empty headers. These do no harm imo.
7. Fixed validation of 0x7F.

### DIFF
--- a/.changeset/swift-vans-sneeze.md
+++ b/.changeset/swift-vans-sneeze.md
@@ -1,0 +1,5 @@
+---
+"github.com/livekit/protocol": patch
+---
+
+Switch to new header validation for SIP commands


livekit/sip_validation.go
```diff
@@ -222,9 +226,11 @@ func ValidateHeaderName(name string) error {
 	}
 
 	// Convert to lowercase for case-insensitive comparison
-	lowerName := strings.ToLower(name)
-	if forbidden, exists := FrobiddenSipHeaderNames[lowerName]; exists && forbidden {
-		return fmt.Errorf("header name %s not supported", name)
+	if restrictNames {
```
alexlivekit, line 229: Gated behind a boolean check - we want to be able to specify "forbidden" headers in `HeadersToAttributes` and `AttributesToHeaders`, but not in `Headers`.

dennwc (approved):

PR #1286, opened 2025-11-06
Add redacted field options to egress fields

biglittlebigben:
Makes sure RPC systems that use the livekit logger wrappers do not log these fields.

This also makes indenting and spaces more consistent by using passing the file through clang-format. Happy to revert this if we want to avoid large cosmetic PRs

These other following changes are suggestions and open for debate:
- changes the go_package location for logger/options.proto to github.com/livekit/protocol/livekit/logger" since github.com/livekit/protocol/logger already exists and has some non generated go code inside. 
- change the redaction code to use <field_name> as redacted value to match the manual redaction code in egress/redact.go. 
- add a unit test for the egress redaction. The unit test lives in github.com/livekit/protocol/livekit because protocol/livekit/sip.go imports github.com/livekit/protocol/logger, so putting the test in github.com/livekit/protocol/logger would create an import loop.


logger/proto.go
```diff
@@ -49,7 +49,7 @@ func (p protoMarshaller) MarshalLogObject(e zapcore.ObjectEncoder) error {
 		v := p.m.Get(f)
 
 		if proto.HasExtension(f.Options(), logger.E_Redact) {
-			e.AddString(k, "<redacted>")
+			e.AddString(k, fmt.Sprintf("<%s>", f.TextName()))
```
paulwe: the name is already the field key in the log and replacing the value with the text "redacted" makes it clear why it's been omitted to readers unfamiliar with the message schema.

protobufs/logger/options.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
```
paulwe, line 1: this file is missing the license header

logger/proto.go
biglittlebigben: The trade off is consistency with the existing logic. I have no strong opinion either way. @frostbyte73 what is your opinion here since I believe you wrote the existing redaction code?

PR #1293, opened 2025-11-07
Egress stream retry info

milos-lk:
When egress uses RTMP streams as outputs - it could happen that a remote RTMP server (or something in between) closes connection. Egress is auto retrying already started recordings but frequent retries could cause multiple video files being recorded on the remote (vendor side - we had a specific case of too many small youtube videos being created)

The suggested protocol update would enable us to communicate when this happens so that publishing side could take some action if needed.


milos-lk:
> How is this communicated to the client? Through a webhook?

Yes - that's correct - through web hook

protobufs/livekit_egress.proto
```diff
@@ -411,6 +411,8 @@ message StreamInfo {
   int64 duration = 4;
   Status status = 5;
   string error = 6;
+  int64 last_retry_at = 7;
+  int64 retries = 8;
```
anunaym14: maybe timestamp for last_retry_at and uint for retries?

milos-lk: thought about it - chose int64 for retry_at to be consistent with started_at and ended_at.
retries could be uint though

PR #1299, opened 2025-11-11
kind details for connector

cnderrauber:


auth/grants.go
```diff
@@ -653,3 +665,24 @@ func kindToProto(sourceStr string) livekit.ParticipantInfo_Kind {
...
+
+func kindDetailsFromProto(details []livekit.ParticipantInfo_KindDetail) []string {
+	result := make([]string, 0, len(details))
+	for _, d := range details {
+		result = append(result, strings.ToLower(d.String()))
+	}
+	return result
+}
+
+func kindDetailsToProto(details []string) []livekit.ParticipantInfo_KindDetail {
```
boks1971, line 677: not including the `connectors` in this?

cnderrauber: good catch! fixed here https://github.com/livekit/protocol/pull/1300

PR #1302, opened 2025-11-12
support deleting existing secrets

yoonsio:
ref HA-315


protobufs/livekit_cloud_agent.proto
```diff
@@ -160,6 +160,7 @@ message UpdateAgentSecretsRequest {
     string agent_name = 2;
     bool overwrite = 3;
     repeated AgentSecret secrets = 4;
+    repeated string delete = 5;
```
rektdeckard: We should call it `remove`, we've had issues in the past using common reserved keywords like `del`/`delete`.

yoonsio: fixed

PR #1305, opened 2025-11-13
Sip to grpc error code mapping update

alexlivekit:
The only things changed here, Ignoring automatic protoc version bump by GHA, are:
- protobufs/livekit_sip.proto : Added more SIP codes to enum.
- livekit/sip.go : Changed to explicit `SIPStatusCode`-to-`codes.Code` map.
- livekit/sip_test.go : Modified test to expect the updated status.

### DIFF
--- a/livekit/sip.go
+++ b/livekit/sip.go
@@ -136,7 +136,7 @@ func (p *SIPStatus) GRPCStatus() *status.Status {
 		} else if code < 600 {
 			code = codes.FailedPrecondition // 5xx from remote server, per guideline (c) in gRPC docs
 		} else if code < 700 {
-			code = codes.InvalidArgument // Same as 4xx ,but authoratative
+			code = codes.InvalidArgument // Same as 4xx ,but authoritative
 		}
 	}
 	msg := p.Status


livekit/sip.go
```diff
@@ -46,39 +46,103 @@ func (p *SIPStatus) Error() string {
...
+		if p.Code < 200 {
+			code = codes.Unknown // 1xx are not final responses, something is wrong.
+		} else if p.Code < 300 {
+			return status.New(codes.OK, "OK") // Preserving previous behavior
+		} else if code < 500 {
+			code = codes.InvalidArgument
+		} else if code < 600 {
+			code = codes.FailedPrecondition // 5xx from remote server, per guideline (c) in gRPC docs
+		} else if code < 700 {
+			code = codes.InvalidArgument // Same as 4xx ,but authoratative
```
dennwc:
```suggestion
			code = codes.InvalidArgument // Same as 4xx ,but authoritative
```

livekit/sip.go
```diff
@@ -46,39 +46,103 @@ func (p *SIPStatus) Error() string {
...
+	SIPStatusCode_SIP_STATUS_PAYMENT_REQUIRED:                 codes.PermissionDenied,
+	SIPStatusCode_SIP_STATUS_FORBIDDEN:                        codes.PermissionDenied,
+	SIPStatusCode_SIP_STATUS_NOTFOUND:                         codes.NotFound,
+	SIPStatusCode_SIP_STATUS_METHOD_NOT_ALLOWED:               codes.InvalidArgument,
+	SIPStatusCode_SIP_STATUS_NOT_ACCEPTABLE:                   codes.InvalidArgument,
+	SIPStatusCode_SIP_STATUS_PROXY_AUTH_REQUIRED:              codes.Unauthenticated,
+	SIPStatusCode_SIP_STATUS_REQUEST_TIMEOUT:                  codes.DeadlineExceeded,
+	SIPStatusCode_SIP_STATUS_CONFLICT:                         codes.FailedPrecondition,
+	SIPStatusCode_SIP_STATUS_GONE:                             codes.NotFound,
+	SIPStatusCode_SIP_STATUS_LENGTH_REQUIRED:                  codes.InvalidArgument,
```
dennwc, line 70: It's up to you, but we could potentially remove most of the 4xx mappings in favour of a default that we already have (it's `InvalidArgument` by default). Could probably keep the most frequent ones in the map and let everything else fall back to a default. But I don't mind if we keep the map as-is.

alexlivekit: I like there to be one source of "What does X turn to", and the fallback being a purely unexpected/unimplemented safety net. Easier to debug imo.

PR #1309, opened 2025-11-16
Redact metadata, attribute, participant name from logs.

boks1971:
Add size option to those.


paulwe (approved):

protobufs/livekit_internal.proto
```diff
@@ -81,7 +81,7 @@ message NodeStats {
   uint64 nack_total = 11;
   float bytes_in_per_sec = 12 [deprecated=true];
   float bytes_out_per_sec = 13 [deprecated=true];
-  float packets_in_per_sec = 14 [deprecated=true];
+  float packets_in_per_sec = 14;
```
boks1971: removing `deprecated` on this as it is used in capacity manager to check packet rate before using forwarding latency for load indication.

anunaym14:
Shall we also redact these from connector protos?

anunaym14:
> yeah, if we have that, will take a look

thank you!

dennwc (approved):

PR #1315, opened 2025-11-20
Utility to clone a proto with redaction.

boks1971:
Would be nice to set size of redacted field back into the message, but that does not work for all types.

Still not useful where size of the redacted field would be good to have.


boks1971:
> this will only work if clone is called directly on the proto with the redacted fields. if the redacted field is in a nested proto it won't be scrubbed

oh, it has to walk the tree? so not useful, will close the PR. Was trying to see if there is something we can do to avoid manual redaction in twirp services

boks1971:
> i didn't mean we should close it, it's a good idea. just need to add more reflection bits to traverse nested data structures

Got it. Will take a look.

PR #1316, opened 2025-11-20
Utility to clone a proto with redaction. 

boks1971:
Would be nice to set size of redacted field back into the message, but
that does not work for all types.

Still not useful where size of the redacted field would be good to have.


protobufs/livekit_models.proto
```diff
@@ -174,15 +174,21 @@ message ParticipantInfo {
   int64 joined_at = 6;
   // timestamp when participant joined room, in milliseconds
   int64 joined_at_ms = 17;
-  string name = 9;
+  string name = 9 [
```
boks1971, line 177: missed these in the previous round.

protobufs/livekit_models.proto
```diff
@@ -227,7 +233,10 @@ message SimulcastCodecInfo {
 message TrackInfo {
   string sid = 1;
   TrackType type = 2;
-  string name = 3;
+  string name = 3 [
```
paulwe, line 236: does the size of these matter?

boks1971: actually no, just using the same construct everywhere. Will remove `redact_format` here.

PR #1324, opened 2025-12-02
Add ai coustics feature usage

lukasIO:


protobufs/livekit_analytics.proto
```diff
@@ -230,6 +230,7 @@ message FeatureUsageInfo {
   enum Feature {
     KRISP_NOISE_CANCELLATION = 0;
     KRISP_BACKGROUND_VOICE_CANCELLATION = 1;
+    AIC_AUDIO_ENHANCEMENT = 2;
```
boks1971, line 233: Is it generic like this or specific like `NOISE_CANCELLATION`, `BVC`?

lukasIO: `audio enhancement` is what they use everywhere where I would expect krisp to use `noise cancellation` . 

We could be more specific by specifying the model name? 
`AIC_AUDIO_ENHANCEMENT_QUAIL` ? 

for reference https://github.com/ai-coustics/aic-sdk-py/blob/main/README.md

boks1971: Okay. Not sure about model. I think if they have only one line of products, I think it is okay to call it that.

boks1971: We can change name later too.

lukasIO: they also are going to have `VAD` which I think we could still call `AIC_VAD`

PR #1331, opened 2025-12-08
Add participant kind & detail for rtsp bridge

cnderrauber:


protobufs/livekit_models.proto
```diff
@@ -154,13 +154,18 @@ message ParticipantInfo {
...
+    BRIDGE = 8;
+    // NEXT_ID: 9
   }
   enum KindDetail {
     CLOUD_AGENT = 0;
     FORWARDED = 1;
     CONNECTOR_WHATSAPP = 2;
     CONNECTOR_TWILIO = 3;
+    // reserved 4-9 for other connector types
+    BRIDGE_RTSP = 10;
```
boks1971: Should we reserve? I think it is going to break anyway if there are more than that many connectors. I think it is fine to keep it continuous.

PR #1332, opened 2025-12-08
Adjust memory usage for inactive file.

boks1971:
Inactive file can be reclaimed if necessary. But, croup memory.usage_in_bytes shows a high value if the memory has grown and does not need to be reclaimed. So, using that inflates the actual memory usage.

Adjust usage by discounting reclaimable inactive file memory.

### DIFF
--- a/utils/hwstats/memory_linux.go
+++ b/utils/hwstats/memory_linux.go
@@ -27,15 +27,15 @@ import (
 )
 
 const (
-	memUsagePathV1          = "/sys/fs/cgroup/memory/memory.usage_in_bytes"
-	memLimitPathV1          = "/sys/fs/cgroup/memory/memory.limit_in_bytes"
-	memStatPathV1           = "/sys/fs/cgroup/memory/memory.stat"
-	totalInactiveFilePathV1 = "total_inactive_file"
+	memUsagePathV1      = "/sys/fs/cgroup/memory/memory.usage_in_bytes"
+	memLimitPathV1      = "/sys/fs/cgroup/memory/memory.limit_in_bytes"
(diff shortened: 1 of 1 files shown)


utils/hwstats/memory_linux.go
```diff
@@ -95,6 +102,10 @@ func (cg *memInfoGetterV1) getMemory() (uint64, uint64, error) {
 	if err != nil {
 		return 0, 0, err
 	}
+	totalInactiveFile, err := readValueOfKeyFromFile(memStatPathV1, totalInactiveFileV1)
```
milos-lk, line 105: I think the error below is genuine -  did you mean `totalInactiveFilePathV1` ?

utils/hwstats/memory_linux.go
```diff
@@ -155,3 +170,35 @@ func readValueFromFile(file string) (uint64, error) {
...
+	fd, err := os.Open(file)
+	if err != nil {
+		return 0, err
+	}
+	defer fd.Close()
+
+	scanner := bufio.NewScanner(fd)
+	for scanner.Scan() {
+		line := scanner.Text()
+		i := strings.IndexRune(line, ':')
```
milos-lk: just double checking -  these aren't space separated?

utils/hwstats/memory_linux.go
boks1971: yep, forgot to try `mage buildLinux` to confirm this. Fixing. Thank you!

utils/hwstats/memory_linux.go
boks1971: You are right, I took code from go-osstat which does `/proc/meminfo`. Will fix.

PR #1341, opened 2025-12-17
feat(whatsapp): add disconnect reason field for disconnect request

anunaym14:


protobufs/livekit_connector_whatsapp.proto
```diff
@@ -68,10 +68,24 @@ message DialWhatsAppCallResponse {
...
+  }
+
   // Required - Call ID sent by Meta
   string whatsapp_call_id = 1;
-  // Required - The API key of the business that is disconnecting the call
+  // The API key of the business that is disconnecting the call
+  // Optional - only if the user disconnected the call
   string whatsapp_api_key = 2 [(logger.redact) = true];
+  // The reason for disconnecting the call
+  DisconnectReason disconnect_reason = 3;
```
biglittlebigben, line 89: We already imports livekit_models.proto indirectly through livekit_rtc.proto, so I guess this does not conflict, but how does this interact with the `DisconnectReason` type there?

protobufs/livekit_connector_whatsapp.proto
```diff
@@ -68,10 +68,24 @@ message DialWhatsAppCallResponse {
...
+    // Note that this webhook will also be sent when the call is disconnected by the business.
+    // Calling the API twice in such cases will result in an error.
+    USER_INITIATED = 1;
+  }
+
   // Required - Call ID sent by Meta
   string whatsapp_call_id = 1;
-  // Required - The API key of the business that is disconnecting the call
+  // The API key of the business that is disconnecting the call
+  // Optional - only if the user disconnected the call
```
biglittlebigben: I thought I understood previously that this field was only needed if the business ended the call, as we don't need to call into WhatsApp if the user disconnects?

protobufs/livekit_connector_whatsapp.proto
anunaym14: https://github.com/livekit/protocol/pull/1341/files#diff-b03fd2d20c68ec5c91c1c1f2639b328dea367fa8b8e0006843bf05167d9937c7R85-R96

The final name for the embedded enum is different, so it does not conflict. I have verified that the generated code uses the embedded type.

protobufs/livekit_connector_whatsapp.proto
anunaym14: That is correct

biglittlebigben: Maybe I understand the comment. I think we need to be clear about when it's required.

anunaym14: How about now?

protobufs/livekit_connector_whatsapp.proto
```diff
@@ -68,10 +68,25 @@ message DialWhatsAppCallResponse {
...
+    // Calling the API twice in such cases will result in an error.
+    USER_INITIATED = 1;
+  }
+
   // Required - Call ID sent by Meta
   string whatsapp_call_id = 1;
-  // Required - The API key of the business that is disconnecting the call
+  // The API key of the whatsapp business.
+  // It is only required if the DisconnectReason is BUSINESS_INITIATED, as we need to make an API call.
+  // For USER_INITIATED, this is not required.
```
biglittlebigben:
```suggestion
  // Required if the DisconnectReason is BUSINESS_INITIATED.
  // Optional for  USER_INITIATED as no API call to WhatsApp is needed.
```

PR #1343, opened 2025-12-19
Updating dispatch rule error message

alexlivekit:


sip/sip.go
```diff
@@ -131,6 +131,13 @@ func printID(s string) string {
 	return s
 }
 
+func printName(s string) string {
```
alexlivekit, line 134: In the same spirit as `printID` right above it.

sip/sip.go
```diff
@@ -208,8 +215,9 @@ func (v *DispatchRuleValidator) Validate(r *livekit.SIPDispatchRuleInfo) error {
 				if v.opt.AllowConflicts {
 					continue
 				}
-				return twirp.NewErrorf(twirp.InvalidArgument, "Conflicting SIP Dispatch Rules: same Trunk+Number+PIN combination for for %q and %q",
-					printID(r.SipDispatchRuleId), printID(r2.SipDispatchRuleId))
```
alexlivekit, line 212: This is apparently confusing to users, so I have changed the wording a little bit.
cc @iKristy

sip/sip.go
```diff
@@ -208,8 +215,9 @@ func (v *DispatchRuleValidator) Validate(r *livekit.SIPDispatchRuleInfo) error {
...
+				return twirp.NewErrorf(twirp.InvalidArgument,
+					"Dispatch rule for the same trunk, number, and pin combination already exists in dispatch rule %q %q",
```
iKristy: Could we capitalize pin as `PIN`?

```suggestion
					"Dispatch rule for the same trunk, number, and PIN combination already exists in dispatch rule %q %q",
```

iKristy (approved):
🚢

PR #1349, opened 2025-12-26
add text message request and response

longcw:


protobufs/livekit_agent.proto
```diff
@@ -181,3 +185,19 @@ message JobAssignment {
 message JobTermination {
   string job_id = 1;
 }
+
+message TextMessageRequest {
```
theomonnom, line 190: Should we have the session_id too?

theomonnom: I see it's on the Notion, just not inside this PR

longcw: yeah this pr hasn't been updated yet, will create a new version based on the discussion on Notion

protobufs/livekit_agent.proto
```diff
@@ -181,3 +185,19 @@ message JobAssignment {
...
+  string session_id = 2;
+  string agent_name = 3;
+  string metadata = 4;
+  bytes session_data = 5;
+  string text = 6;
+}
+
+message TextMessageResponse {
+  string message_id = 1;
+  string response = 2;
```
theomonnom:
```suggestion
  string response = 2;
```
What is the response str used for?

PR #1419, opened 2026-02-19
feat(sip): Add from_host field to SIP outbound trunk

### LINKED ISSUE #578: [Feature Request] Allow customizing 'From' header domain (fromHost) in SIP Outbound Trunk
### Is your feature request related to a problem? Please describe.

Currently, when creating a SIP Outbound Trunk via `CreateSIPOutboundTrunk` (or `CreateSIPParticipant`), there is no option to explicitly set the domain part of the SIP `From` header (the `fromHost`).

The SIP service seems to automatically construct the `From` header using the authenticated username and the service's configured `sip.local_ip` or global domain.

**The Problem:**

Some strict SIP providers require the `From` header domain to match the registration domain exactly (or a specific verified domain), rejecting calls where the `From` domain is an IP address or the internal hostname of the LiveKit SIP service.

### Describe the solution you'd like

I would like to request a new field in the `SIPOutboundTrunkInfo` (and potentially `CreateSIPParticipantRequest`) protobuf definition, such as `from_domain` or `from_host`.

If set, the SIP service should use this value to construct the domain part of the `From` header in outgoing INVITE requests, overriding the default behavior (which currently seems to fall back to the local IP or global config).

**Example desired behavior:**

```go
// Proposed field
trunk := &livekit.SIPOutboundTrunkInfo{
    // ... existing fields
    FromDomain: "sip.myprovider.com", 
}
```

Resulting header: `From: <sip:username@sip.myprovider.com>`

### Describe alternatives you've considered

1. **Global Config:** Changing the global `domain` in `config.yaml`. This is not ideal because a single SIP service instance might need to connect to multiple providers with different domain requirements.
2. **Custom Headers:** Trying to pass `From` in the custom `headers` map. This usually doesn't work as the SIP stack overwrites standard headers like `From` to ensure protocol compliance.

### Additional context

- **SDK/Version:** LiveKit SIP Service
- **Use Case:** Connecting to a legacy SIP provider that enforces strict domain validation on the `From` header.

civilcoder55:
* Add from_host field to SIPOutboundTrunkInfo and SIPOutboundConfig
* Add validation for from_host to ensure it's a domain/IP, not a SIP URI
* Update SIPOutboundTrunkUpdate to support from_host changes
* Use from_host as hostname in CreateSIPParticipantRequest, fallback to own hostname if not set
---
 
This option lets users define a **custom From-header hostname** when setting up an outbound trunk, allowing SIP messages to have a hostname in the From header that differs from the server’s default.

This resolves https://github.com/livekit/sip/issues/578 

**Before** 
    
```json
{
  "trunk": {
    "name": "Normal outbound",
    "address": "normal.pstn.twilio.com",
    "numbers": ["+15105550100"],
    "authUsername": "normal",
    "authPassword": "normal"
  }
}
```
image

**After**

```json
{
  "trunk": {
     "name": "Overridden outbound",
    "address": "normal.pstn.twilio.com",
    "numbers": ["+15105550101"],
    "authUsername": "normal",
    "authPassword": "normal",
    "fromHost": "overridden.com"
  }
}
```

image


livekit/sip_test.go
```diff
@@ -200,6 +200,29 @@ func TestSIPValidate(t *testing.T) {
...
 
+func TestValidateHost(t *testing.T) {
+	cases := []struct {
+		name string
+		host string
+		exp  bool
+	}{
+		{name: "empty", host: "", exp: true},
+		{name: "domain", host: "from.example.com", exp: true},
+		{name: "ipv4", host: "1.2.3.4", exp: true},
```
dennwc, line 214: Please add IPv6 as well. It might be tempting to add `:` to the list of disallowed characters, but it's valid in v6, so having a test for it might be a good idea.

rpc/sip.go
```diff
@@ -163,13 +166,17 @@ func NewCreateSIPParticipantRequest(
...
+	}
+
 	return &InternalCreateSIPParticipantRequest{
 		ProjectId:             projectID,
 		SipCallId:             callID,
 		SipTrunkId:            trunkID,
 		DestinationCountry:    destinationCountry,
 		Address:               hostname,
-		Hostname:              ownHostname,
+		Hostname:              fromHost,
```
dennwc: Note that this change will affect LiveKit Cloud. For now, we do not want users to be able to override the domain in the `From` header.

It totally makes sense for OSS, though. So I think we'll need to pass an extra parameter like `allowCustomHostname` to this function. If it is false, `fromHost` must have no effect.

dennwc (changes requested):
Thank you for implementing it! The change looks good, but I'm blocking it for now. Please see the reason below.

livekit/sip_test.go
civilcoder55: @dennwc can you check it now?

rpc/sip.go
civilcoder55: updated
now, the allowCustomHostname  option has been added to the SIP server configuration
https://github.com/livekit/sip/pull/594

rpc/sip.go
```diff
@@ -170,6 +173,7 @@ func NewCreateSIPParticipantRequest(
 		DestinationCountry:    destinationCountry,
 		Address:               hostname,
 		Hostname:              ownHostname,
+		FromHostname:          fromHost,
```
dennwc, line 176: I think we should not introduce a new field here. Lets just keep the `Hostname` here and pass `allowCustomFrom` (or whatever we want to call it) to the `NewCreateSIPParticipantRequest` function. This will intentionally break compatibility and we'll have to pass something to it in Cloud. This will ensure it cannot silently change the behavior.

civilcoder55: @dennwc 
The challenge is that I can’t read the SIP server configuration when executing `NewCreateSIPParticipantRequest` because it is being called from LiveKit, not from SIP. 
What do you suggest?

dennwc: That's okay, OSS can always pass `true` here.

protobufs/rpc/sip.proto
```diff
@@ -50,6 +50,8 @@ message InternalCreateSIPParticipantRequest {
   string address = 2;
   // Hostname for the 'From' SIP address in INVITE
   string hostname = 20;
+  // custom Hostname for the 'From' SIP address in INVITE
+  string from_hostname = 34;
```
dennwc, line 54: I might be missing something here, but how it's different from the `hostname` above?

civilcoder55: image

@dennwc 
The hostname value is always set to ownHostname, which is passed as an empty string from livekit-server and defaults later to the SIP server Host value

dennwc: Lets change it in one of those places then. We shouldn't introduce another field in the protocol for this to work. We already have a field, just need to use it differently.

dennwc:
Replaced with https://github.com/livekit/protocol/pull/1577 (allows setting `From` for OSS edition).

PR #1485, opened 2026-04-08
AGT-2669: Add turn detection and interruption protobufs

chenghao-mou:
Add protobuf messages for both client <-> Inference and Inference <-> GPU host.


protobufs/agent/livekit_agent_turn_detector.proto
```diff
@@ -0,0 +1,142 @@
...
+    TdInferenceStarted inference_started = 2;
+    TdInferenceStopped inference_stopped = 3;
+    TdEouPrediction eou_prediction = 4;
+    TdSessionClosed session_closed = 5;
+    TdError error = 6;
+  }
+  optional string request_id = 7;
+  google.protobuf.Timestamp server_created_at = 8;
+  optional google.protobuf.Timestamp client_created_at = 9;
+}
```
theomonnom: Should we split this into a Request/Response message?

theomonnom: For the RemoteSession, I have Request/Response and events 
https://github.com/livekit/protocol/blob/main/protobufs/agent/livekit_agent_session.proto

theomonnom: It's nit tho

chenghao-mou: I guess it depends on how you look at them: e.g. `TdInferenceStart` and `TdInferenceStarted` is a Request-and-Response pair, we just wrap all the requests and responses in `TdClientMessage` and `TdServerMessage`.

davidzhao (commented):
is this only for the turn detector? or more designed to be the generic websocket protocol?

chenghao-mou:
> is this only for the turn detector? or more designed to be the generic websocket protocol?

They are only for the turn detector. Barge-in or STT have somewhat different control messages or different formats:

Bargein (ws binary+json): in: audio frame (pcm), inference start/end, out: prediction
STT (json): in: audio frame (pcm), out: transcripts
Turn detection (protobuf binary): audio frame (opus), chat messages, inference start/end, out: prediction

Do you think we should try to unify them?

protobufs/agent/livekit_agent_inference.proto
```diff
@@ -0,0 +1,219 @@
...
+  google.protobuf.Duration prediction_duration = 5;
+}
+
+message Prediction {
+  oneof prediction {
+    EotPrediction eot_prediction = 1;
+    InterruptionPrediction interruption_prediction = 2;
+  }
+}
+
```
theomonnom, line 205: I think it's unnecessary nesting here, let's directly use `EotPrediction` and `InterruptionPrediction` inside the ServerMessage.

protobufs/agent/livekit_agent_inference.proto
```diff
@@ -0,0 +1,219 @@
...
+  google.protobuf.Duration inference_duration = 3;
+}
+
+message ProcessingStats {
+  google.protobuf.Timestamp earliest_client_created_at = 1;
+  google.protobuf.Timestamp latest_client_created_at = 2;
+  // client-side e2e latency (client send to client receive)
+  google.protobuf.Duration e2e_latency = 3;
+  InferenceStats inference_stats = 4;
+}
```
theomonnom, line 154: maybe?
```suggestion
message ProcessingStats {
  message InferenceStats {
    // server-side e2e latency (server input to server output)
    google.protobuf.Duration e2e_latency = 1;
    google.protobuf.Duration preprocessing_duration = 2;
    google.protobuf.Duration inference_duration = 3;
  }

  google.protobuf.Timestamp earliest_client_created_at = 1;
  google.protobuf.Timestamp latest_client_created_at = 2;
  // client-side e2e latency (client send to client receive)
  google.protobuf.Duration e2e_latency = 3;
  InferenceStats inference_stats = 4;
}
```

protobufs/agent/livekit_agent_inference.proto
```diff
@@ -0,0 +1,219 @@
...
+  }
+}
+
+message Error {
+  string message = 1;
+  // error code follows the HTTP status code convention
+  // 4xx for client errors
+  // 5xx for server errors
+  uint32 code = 2;
+}
```
theomonnom, line 50: This Error message is too generic. Can we have another name?

Fwiw, we only do a `string error` in other services

protobufs/agent/livekit_agent_inference.proto
chenghao-mou: The model-to-Inference response needs to include `InferenceStats` but not `ProcessingStats` (`EotInferenceResponse` and `InterruptionInferenceResponse`)

protobufs/agent/livekit_agent_inference.proto
chenghao-mou: The code here is the convention from Inference. We use that for the `status_code` in `APIStatusError`. 

Renamed it to `InferenceError`.

theomonnom (approved):

protobufs/agent/livekit_agent_inference.proto
theomonnom: I see, it's kind of confusing, could we maybe just have one message for stats? 

and two nested optional structs inside it?

chenghao-mou: Made a flat InferenceStats and Inference-to-model fields optional, so now we only have one message for both senders and receivers.

PR #1491, opened 2026-04-09
Reverting node suspended state.

boks1971:
Cannot use this to steer a kube deployment to scale up while keeping a pod alive. So, reverting.


lukasIO (approved):

milos-lk (approved):

boks1971:
Apologise for the spam here. The extra state is needed to ensure that new participants are not allocated to the node. So, not going to merge this revert.

PR #1501, opened 2026-04-16
Export analytics proto from JS package

lukasIO:


1egoman (approved):
Makes sense. Out of curiosity, what is this being exported in support of? Something agents specific?

PR #1502, opened 2026-04-17
Add TimedString proto

longcw:
Add `TimedString` message to `livekit_agent_session.proto` mirroring the Python [`TimedString`](https://github.com/livekit/agents/blob/main/livekit-agents/livekit/agents/types.py#L115) class, with fields for text, optional start/end time, confidence, start_time_offset, and speaker_id.


chenghao-mou (approved):

davidzhao (approved):
lg

PR #1505, opened 2026-04-17
Add SDK capabilities to ClientInfo

chenosaurus:


protobufs/livekit_models.proto
```diff
@@ -532,6 +532,15 @@ message ClientInfo {
     ESP32 = 14;
   }
 
+  // Optional capabilities advertised by the client at connect time. The SFU
+  // uses these flags to decide whether to enable features that require
+  // client-side support (e.g. passing RTP packet trailers through to the
+  // subscriber instead of stripping them).
+  enum Capability {
+    CAP_UNKNOWN = 0;
```
lukasIO, line 540: what's the intention behind having an `Unknown` value here? 
I don't think we would want/need an unknown value

protobufs/livekit_models.proto
```diff
@@ -532,6 +532,15 @@ message ClientInfo {
...
+    CAP_PACKET_TRAILER = 1;
```
boks1971, line 541: should this be more specific, something like `CAN_DECODE_PACKET_TRAILER`? I think most of these would be to indicate that it can handle something on receive, but I am thinking there may be cases where it indicates encoding/sending capabilities. So, making these explicit would be great.

We can change the name later too as long as the enum value stays the same, but just wanted to bring it up for more clarity/readability.

protobufs/livekit_models.proto
boks1971: I would probably rename to `CAP_UNUSED`. I think it is useful to mark the default value as unused. In this case, it is not really useful as `repeated` field will not include the default, but I think it is a good practice to just block off defaults so that it does not get used unintentionally anywhere.

boks1971 (approved):

protobufs/livekit_models.proto
chenosaurus: Exactly, just to block off the default value so we'd never have a situation where if a field is unset the value is ambiguous.  I will update to `CAP_UNUSED` instead to be more clear.

protobufs/livekit_models.proto
chenosaurus: I think for packet trailers, we wouldn't have separate capabilities for encode/decode.  But noted for future capabilities.

PR #1514, opened 2026-04-23
TEL-543: same trunk can match duplicate normalized SIP numbers

hechen-eng:


alexlivekit (approved):
nice. 

Do we want to run that check when updating the definition as well?

regardless, lgtm

hechen-eng:
> nice.
> 
> Do we want to run that check when updating the definition as well?
> 
> regardless, lgtm

yeah, I think we should do that 👍

Let me do it in a separate PR though.

PR #1520, opened 2026-04-28
Hosted agents: add agent name to proto

shawnyang-coder:


protobufs/livekit_cloud_agent.proto
```diff
@@ -171,7 +173,8 @@ message RestartAgentResponse {
 
 message DeployAgentRequest {
     string agent_id              = 1 [(logger.name) = "agentID"];
-    string agent_name            = 2 [deprecated=true];
+    reserved 2;
+    reserved "agent_name";
```
real-danm, line 177: why the different format here than the others?

real-danm (approved):

protobufs/livekit_cloud_agent.proto
shawnyang-coder: reserved is to make sure we don't use agent_name. it's stronger than deprecated. 
It's what we want for Deploy.

PR #1521, opened 2026-04-29
Fix the egress results serialization format

biglittlebigben:
Only serialize in the non deprecated format so no "ResultType" field is needed.

### DIFF
--- a/observability/egressobs/egress.go
+++ b/observability/egressobs/egress.go
@@ -25,10 +25,10 @@ const (
 )
 
 type EgressResults struct {
-	FileResults    []*livekit.FileInfo
-	StreamResults  []*livekit.StreamInfo
-	SegmentResults []*livekit.SegmentsInfo
-	ImageResults   []*livekit.ImagesInfo
+	FileResults    []*livekit.FileInfo     `json:"file_results,omitempty"`
+	StreamResults  []*livekit.StreamInfo   `json:"stream_results,omitempty"`
+	SegmentResults []*livekit.SegmentsInfo `json:"segment_results,omitempty"`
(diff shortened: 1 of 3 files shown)


cacheonly (approved):

paulwe (approved):

observability/egressobs/egress.go
```diff
@@ -25,10 +25,10 @@ const (
...
 
 type EgressResults struct {
-	FileResults    []*livekit.FileInfo
-	StreamResults  []*livekit.StreamInfo
-	SegmentResults []*livekit.SegmentsInfo
-	ImageResults   []*livekit.ImagesInfo
+	FileResults    []*livekit.FileInfo     `json:"file_results,omitempty"`
+	StreamResults  []*livekit.StreamInfo   `json:"stream_results,omitempty"`
+	SegmentResults []*livekit.SegmentsInfo `json:"segment_results,omitempty"`
+	ImageResults   []*livekit.ImagesInfo   `json:"image_results,omitempty"`
```
paulwe, line 31: is this currently snake case? what consumes this? does this break backward compatibility?

biglittlebigben: This was currently not working because:
1. cloud-api-server tries to parse the content of this field as if were in the same format as the analytics corresponding field
2. egress is serializing this in different formats depending on the type of result without actually indicating what the type is

See https://linear.app/livekit/issue/CS-1673/v3-egress-missing-data-points

I can change the schema to write to a new column if you want to somehow keep compatibility with the data that's currently in the "result" column. Do we need to considering the context above?

paulwe: backward compatibility with the existing data isn't important but the way api server is parsing this is sketchy...

PR #1528, opened 2026-05-02
add duration seconds reporting

paulwe:


observability/sessiontimer.go
```diff
@@ -1,25 +1,53 @@
...
 
-func (h *SessionTimer) Advance(now time.Time) (millis, mins int64) {
+func (h *SessionTimer) Advance(now time.Time) (millis, secs, mins int64) {
 	ts := now.UnixMilli()
 	if ts > h.lastMilli {
 		millis = ts - h.lastMilli
 		h.lastMilli = ts
 	}
+	if ts > h.lastSec {
+		n := max((ts-h.lastSec+999)/1000, h.minSecs)
```
boks1971: comments says so, but just wanted to confirm that this is intentional, i. e. however small the first advance is and however min big min seconds, it is applied.

To take an exaggerated example, if min seconds is 100 and the first advance is 1ms, it will count 100 the first time and then let us say there are 100 more updates every second, the total would be close 200 seconds at the end, i.e 100s wall block time elapsed, but 200 secs counted. Is that understanding correct?

My mental model was "100s minimum, anything less than 100s will round up to 100s, after that, it will be elapsed time with 1 second granularity"

boks1971 (approved):

observability/sessiontimer.go
paulwe: `h.lastSec` is advanced by `n * 1000` so on the first tick in your example we return 100 then for the next 100 seconds we return 0 then we begin returning rounded/elapsed seconds

boks1971: 🤦 missed that detail. Thank you Paul.

PR #1530, opened 2026-05-02
Allow setting a list of SIP codecs for SDP.

dennwc:


protobufs/rpc/sip.proto
```diff
@@ -121,7 +121,8 @@ message InternalCreateSIPParticipantRequest {
   // Max call duration.
   google.protobuf.Duration max_call_duration = 24;
 
-  livekit.SIPMediaEncryption media_encryption = 28;
+  livekit.SIPMediaEncryption media_encryption = 28 [deprecated = true];
```
nishadmusthafa, line 124: How widely used is this right now ?

protobufs/livekit_sip.proto
```diff
@@ -223,6 +223,19 @@ enum SIPMediaEncryption {
...
+  string name = 1;
+  uint32 rate = 2;
+}
+
+message SIPMediaConfig {
+  // if set, ignore the default codecs and use the list below.
+  bool only_listed_codecs = 1;
+  // List of allowed codecs. If only_listed_codecs is not set, this list is added to default codecs.
+  repeated SIPCodec codecs = 2;
+  SIPMediaEncryption encryption = 3;
```
nishadmusthafa: What else might we potentially add here in the future ?

nishadmusthafa (approved):

protobufs/rpc/sip.proto
dennwc: Just marking it as deprecated so it's easier to update in our codebase. The old parameter will still continue to work as this PR applies it for both forward and backward compatibility.

protobufs/livekit_sip.proto
dennwc: Krisp is an obvious one we should move here, although I'm not sure how it should look like. Krisp shouldn't be a flag as it is right now, it's better if we allow user to pick from multiple noise cancellation (or other) models. But that would be a bigger change.

livekit/sip.go
```diff
@@ -1024,3 +1030,53 @@ var (
...
+
+func (p *SIPCodec) Validate() error {
+	if p.Name == "" {
+		return errors.New("missing codec name")
+	}
+	if strings.ContainsAny(p.Name, " \t\n\r") {
+		return errors.New("invalid codec name")
+	}
+	if strings.Contains(p.Name, "/") {
+		return errors.New("codec name must not contain '/'")
```
hechen-eng, line 1042: is there a spec for the codec name?

hechen-eng (approved):

livekit/sip.go
dennwc: IT should match the one used in SDP. This just enforces that the name will not contain sample rate or channels, as it will be a separate field in protobuf.

PR #1534, opened 2026-05-05
Adding ability to specify per-trunk media timeouts

alexlivekit:


hechen-eng (approved):

dennwc:
@alexlivekit lets lend it after https://github.com/livekit/protocol/pull/1530 and move the media timeout into the new `SIPMediaConfig`. It will be easier to pass it everywhere that way.

livekit/sip.go
```diff
@@ -451,6 +471,10 @@ func (p *SIPInboundTrunkInfo) Validate() error {
 	if err := validateHeaderToAttributes(p.HeadersToAttributes); err != nil {
 		return err
 	}
+	timeout := MaxSIPMediaTimeout
```
dennwc: After https://github.com/livekit/protocol/pull/1530 you could use `new(MaxSIPMediaTimeout)` as it updates Go to 1.26.

protobufs/rpc/io.proto
```diff
@@ -240,7 +240,10 @@ message EvaluateSIPDispatchRulesResponse {
 
   map<string, string> feature_flags = 23;
 
-  // NEXT ID: 24
+  // Per-call RTP media timeout; if unset, SIP service defaults apply.
+  google.protobuf.Duration media_timeout = 24;
```
dennwc: Do we want to allow overriding the initial media timeout as well?

protobufs/livekit_sip.proto
```diff
@@ -355,7 +355,10 @@ message SIPInboundTrunkInfo {
   google.protobuf.Timestamp created_at = 17;
   google.protobuf.Timestamp updated_at = 18;
 
-  // NEXT ID: 19
+  // Max time a call can last without incoming RTP data received. If unset, defaults are used.
+  google.protobuf.Duration media_timeout = 19;
```
dennwc: Lets move it to the dispatch rule level, similar to https://github.com/livekit/protocol/pull/1530. If you move this field to the `SIPMediaConfig` it will already be on the right level.

In general we should avoid adding new settings to trunks, as they might be gone after we implement the webhooks.

protobufs/livekit_sip.proto
```diff
@@ -443,7 +447,10 @@ message SIPOutboundTrunkInfo {
   google.protobuf.Timestamp created_at = 16;
   google.protobuf.Timestamp updated_at = 17;
 
-  // NEXT ID: 18
+  // Max time a call can last without incoming RTP data received. If unset, defaults are used.
+  google.protobuf.Duration media_timeout = 18;
```
dennwc: Same here, we could just leave it on the `CreateSIPParticipant` level and avoid using trunks.

protobufs/rpc/io.proto
alexlivekit: that was the initial approach, but opted out of it for now - this is mainly for oddball hold behavior, which happens after setup.

protobufs/livekit_sip.proto
alexlivekit: That's the thing, I want it on the trunk, not on the dispatch rule.
So then we have a question - Do we add media config to the trunk? to be able to update per-trunk codecs and media timeouts? do we remove that from dispatch rules? Or are you suggesting moving this functionality to a dispatch rule (&CreateSipParticipant) to match codec specification?

Edit, just read the other comment - I get the suggestion to move this to a per-dispatch-rule instead. Ignore.

livekit/sip.go
```diff
@@ -702,7 +713,7 @@ func (p *SIPDispatchRuleUpdate) Apply(info *SIPDispatchRuleInfo) error {
 	}
 	applyListUpdate(&info.TrunkIds, p.TrunkIds)
 	applyUpdatePtr(&info.Rule, p.Rule)
-	applyUpdatePtr(&info.Media, p.Media)
+	applyUpdatePtr(&info.Media, p.Media) // TODO: Consider applying partial updates
```
alexlivekit, line 716: @dennwc , we might need to revisit this to avoid it being a pain long term

livekit/sip.go
```diff
@@ -1061,20 +1072,33 @@ func (p *CreateSIPParticipantRequest) Upgrade() {
 	p.Media = p.Media.UpgradeWith(p.MediaEncryption)
 }
 
+func (p *SIPMediaEncryption) Deref() SIPMediaEncryption {
+	if p == nil {
+		return SIPMediaEncryption_SIP_MEDIA_ENCRYPT_DISABLE
+	}
+	return *p
+}
```
alexlivekit, line 1080: Moved as-is.

PR #1540, opened 2026-05-06
fix SIP trunk-level MediaEncryption being silently dropped

hechen-eng:
fixing https://github.com/livekit/protocol/pull/1530


alexlivekit (approved):

rpc/sip.go
davidzhao, line 170: this needs a nil check, so we don't end up with a nil media, and creating an empty instance with the correct MediaEncryption value

we reference media.Encryption [below](https://github.com/livekit/protocol/pull/1540/changes#diff-e16b3dbd0a173707258bcdfdcf1fce9e22b4af1fa6b407db1a8df76ceade8869L199)

hechen-eng: `UpgradeWith` already handles nil case, so we should be fine 👍

dennwc (approved):

PR #1553, opened 2026-05-12
Data tracks schema metadata

ladvoc:
Closes BOT-318


protobufs/livekit_models.proto
```diff
@@ -282,6 +282,31 @@ message DataTrackInfo {
...
+  string name = 1;
+  DataTrackSchemaFormat format = 2;
+}
+
+message DataTrackSchemaDefinition {
+  DataTrackSchemaId id = 1;
+  bytes definition = 2;
+}
+
+enum DataTrackSchemaFormat {
```
biglittlebigben: I think some of these at least should have comments/links to specs

protobufs/livekit_rtc.proto
```diff
@@ -63,6 +63,8 @@ message SignalRequest {
     UnpublishDataTrackRequest unpublish_data_track_request = 20;
     // Update subscription state for one or more data tracks
     UpdateDataSubscription update_data_subscription = 21;
+    // Get the definition for a known data track schema
+    ResolveDataTrackSchemaRequest resolve_data_track_schema = 22;
```
biglittlebigben: Why "Resolve" and not "Get" or nothing here?

ladvoc: No good reason, changed in [1e760ab](https://github.com/livekit/protocol/pull/1553/commits/1e760ab61c183c44f4585b336030797a013b1584).

protobufs/livekit_models.proto
ladvoc: Added. Taking another look at the MCAP spec, I realized it isn't a good idea to conflate the concept of message (frame in data track terminology) and schema encoding, so I have separated them here accordingly.

biglittlebigben: It is not immediately clear to me what a schema encoding is (vs a frame encoding). Are both needed, and if so can we provide some info in the proto file?

protobufs/livekit_models.proto
```diff
@@ -282,6 +282,73 @@ message DataTrackInfo {
 
   // Method used for end-to-end encryption (E2EE) on packet payloads.
   Encryption.Type encryption = 4;
+
+  // Type information for this track. If unset, the track is untyped and frame payloads
+  // are treated as opaque bytes.
```
boks1971: Technically, there is no treatment in our system, right? Probably stop at `the track is untyped`?

boks1971: Can one of `frame_encoding`/`schema_encoding` be set and the other unset? Not sure if we want to elaborate the comment to say the typical encoding combinations and also maybe mention if some combination is prohibited. Maybe, a better place to add those might be `AddDataTrack` which is more user facing.

boks1971: oh you have
```
optional DataTrackSchemaId schema = 2;
```
Should that be optional or should it just be default unspecified?

protobufs/livekit_rtc.proto
```diff
@@ -561,6 +579,7 @@ message RequestResponse {
     INVALID_NAME = 8;
     DUPLICATE_HANDLE = 9;
     DUPLICATE_NAME = 10;
+    INVALID_SCHEMA = 11;
```
boks1971: maybe `UNKNOWN_SCHEMA`, there is no validation of schema in our system, so `INVALID` technically is not quite right. `UNKNOWN` feels better.

boks1971 (commented):
Probably hold this till we have some form of implementation?

Should `AddDataTrack` request fail if there is an existing schema id, but the definition bytes do not match? If so, need to add another error reason for this case.

Should schemas be per publisher? If that is the case, `GetSchema` should include participant.

If not per publisher, (i. e. per room), becomes trickier to coordinate in a distributed room, so would be good to avoid that.

protobufs/livekit_models.proto
biglittlebigben: Thinking that if these schema are MCAP specific, MCAP should be in the constant name.

protobufs/livekit_models.proto
ladvoc: Great question! In most cases, they are the same. For example, the schema for a Protobuf is defined using the Protobuf IDL and encoded in the Protobuf binary format. However, for frames encoded using CDR, the schema can be defined using _ROS2_MSG_, _ROS2_IDL_, or _OMG_IDL_ syntax (same wire format, different schema definition language). This is why MCAP distinguished between the two. Happy to add this explanation as a comment in the Protobuf if you think it belongs there.

protobufs/livekit_models.proto
```diff
@@ -282,6 +282,73 @@ message DataTrackInfo {
...
+  DATA_TRACK_SCHEMA_ENCODING_FLATBUFFER = 2;
+  DATA_TRACK_SCHEMA_ENCODING_ROS1_MSG = 3;
+  DATA_TRACK_SCHEMA_ENCODING_ROS2_MSG = 4;
+  DATA_TRACK_SCHEMA_ENCODING_ROS2_IDL = 5;
+  DATA_TRACK_SCHEMA_ENCODING_OMG_IDL = 6;
+  DATA_TRACK_SCHEMA_ENCODING_JSON_SCHEMA = 7;
+}
+
+// Type information for a data track.
+message DataTrackTypeInfo {
```
alan-george-lk: I don't really know the nuances of this design on the implementation side so I can't suggest an improvement confidently, but it's a tad confusing we have `DataTrackTypeInfo` and `DataTrackTypeDefinition` with nearly identical fields. Maybe a rename or consolidation of these types would reduce perceived redundancy

protobufs/livekit_models.proto
ladvoc: > Can one of frame_encoding/schema_encoding be set and the other unset? Not sure if we want to elaborate the comment to say the typical encoding combinations and also maybe mention if some combination is prohibited. Maybe, a better place to add those might be AddDataTrack which is more user facing.

Good point. Yes, one example is when your frame encoding is JSON. Since JSON is self-describing, a schema isn't required—but you can specify one using JSON Schema if desired. I will add a comment documenting explicitly which combinations are allowed/disallowed.

ladvoc: > Should that be optional or should it just be default unspecified?

I made that field optional since `DataTrackSchemaId` is a message (encapsulating both name and schema encoding), not a plain enum. Semantically, I find this cleaner since it communicates "you either have a schema ID or you don't." If the field isn't optional, you have to check the enum field within the message for the unspecified case to know if the track is untyped.

boks1971: Sorry missed that it was a message. Optional makes sense

ladvoc: > Should AddDataTrack request fail if there is an existing schema id, but the definition bytes do not match? If so, need to add another error reason for this case.

Also a good callout. The question is, is there a real use case where a schema would need to get redefined during the same session? @stephen-derosa, @alan-george-lk what do you think?

> If not per publisher, (i. e. per room), becomes trickier to coordinate in a distributed room, so would be good to avoid that.

I see, so, since this will live in an in-memory store and participants can be connected to different nodes, the schema will be stored on the node the publisher is connected to? Therefore, when a `GetSchema` request is made, the node the requesting participant is connected to may need to make an RPC to the other to get the schema definition. Just want to make sure I have a clear picture here.

stephen-derosa: since a DataTrack is unique per participant, i would think that we _should_ fail if the same schema id is used but the bytes dont match. This would likely be a an application side error since it means the same participant has two different message definitions for the same id

boks1971:
> Should AddDataTrack request fail if there is an existing schema id, but the definition bytes do not match? If so, need to add another error reason for this case.

> Also a good callout. The question is, is there a real use case where a schema would need to get redefined during the same session? @stephen-derosa, @alan-george-lk what do you think?

> since a DataTrack is unique per participant, i would think that we should fail if the same schema id is used but the bytes dont match. This would likely be a an application side error since it means the same participant has two different message definitions for the same id

sounds good. Should add an error reason for this.

boks1971:
> If not per publisher, (i. e. per room), becomes trickier to coordinate in a distributed room, so would be good to avoid that.

> I see, so, since this will live in an in-memory store and participants can be connected to different nodes, the schema will be stored on the node the publisher is connected to? Therefore, when a GetSchema request is made, the node the requesting participant is connected to may need to make an RPC to the other to get the schema definition. Just want to make sure I have a clear picture here.

Most probably, this will be propagated to other nodes a la `DataTrackInfo`. The challenge is more on `AddDataTrack` side. The schema has to be synced to all nodes so that when a publisher adds a schema it can be checked for collisions etc. That leads to races as cross-region sync could race with some other user publishing. So, per publisher makes that simpler as a participant is connected to one node. And `GetSchema` specifying publisher identity makes it cleaner to lookup the specific publisher.

Still have to think about migration cases, i. e. should clients add the schemas to `SyncState`?

protobufs/livekit_models.proto
ladvoc: We might be able to improve naming, but there is a subtle difference which makes both types necessary. In `DataTrackInfo` (which is sent down to clients in high frequency participant updates), only `DataTrackTypeInfo` is present to avoid sending down the whole definition every time. When initially publishing the data track or getting the definition, `DataTrackTypeDefinition` carries both the ID and definition itself. Each message provides the fields needed in the context in which it is used.

protobufs/livekit_rtc.proto
ladvoc: I removed `INVALID_SCHEMA`. Thinking through error cases a bit more, I think we will need some basic validation. Here is what I propose:
- `PublishDataTrackRequest` "returns" `INVALID_TYPE_DEFINITION` if any of the follow are true:
  -  The schema ID is invalid due to the name being empty or being above 256 characters
  -  The schema definition exceeds the limit (50KB) 
  -  The schema has previously been used but the definition changed
- `GetDataTrackSchemaRequest` "returns" `NOT_FOUND` (existing variant) if the requested schema does not exist

Open question: as decided in the above thread, schema definition isn't allowed to change, but should we allow it to be omitted entirely if it has already been sent when publishing a previous track?

As I noted before, only certain combinations of frame encoding and schema encoding can be used together. However, I think that validation is out of scope for the SFU to worry about (but I documented them on the enum variants in [5db1cd1](https://github.com/livekit/protocol/pull/1553/commits/5db1cd1e108d42b7fe70a6e30a48ac0813f44598)).

protobufs/livekit_models.proto
```diff
@@ -294,12 +294,19 @@ message DataTrackInfo {
 //
 enum DataTrackFrameEncoding {
   DATA_TRACK_FRAME_ENCODING_UNSPECIFIED = 0;
+  // ROS 1: must be described by `ROS1_MSG` schema encoding.
```
stephen-derosa: i dont think we will ever support a ROS1 wrapper, i think we should remove this functionality, maybe even leave a comment that its not supported

stephen-derosa: though on the other hand people could write their own ros1 wrapper so maybe its ok ?

ladvoc: > though on the other hand people could write their own ros1 wrapper so maybe its ok ?

Yeah, this is just for metadata passthrough, so I don't see any harm in leaving it in.

protobufs/livekit_models.proto
```diff
@@ -294,12 +294,19 @@ message DataTrackInfo {
 //
 enum DataTrackFrameEncoding {
   DATA_TRACK_FRAME_ENCODING_UNSPECIFIED = 0;
```
stephen-derosa: FMU why the redundant `DATA_TRACK_FRAME_ENCODING` thats in the `enum DataTrackFrameEncoding` ? I.e. why not `DATA_TRACK_FRAME_ENCODING_UNSPECIFIED` -> `UNSPECIFIED`

ladvoc: This follows the [Protobuf style guide](https://protobuf.dev/programming-guides/style/#enums) which says to do this to avoid name collisions. At least for some binding generators (prost for Rust is one such example), the generated enum drops the prefix automatically. Don't have a strong opinion on this though.

boks1971: Yeah, name collisions in the same package.

We have used a short form, in this case, you can use `DTFE_`, but full name is fine too. But, we are avoiding no prefix

protobufs/livekit_rtc.proto
boks1971: Regarding re-use of schema, I was thinking it would be tied to a track, but I think it makes sense at participant level. In that case, I would probably de-couple schema registry from `AddDataTrack` message as you can register once and re-use it for multiple tracks if you want.

A registry message is cleaner as piggybacking on `AddDataTrack` gives the indication that it is track specific.

ladvoc:
@boks1971, we thought though the edge cases of having schemas being scoped to specific participants and the conclusion is this will work with some minor caveats. Question: should this work based on participant ID or identity for the purposes of the protocol?

boks1971:
> @boks1971, we thought though the edge cases of having schemas being scoped to specific participants and the conclusion is this will work with some minor caveats. Question: should this work based on participant ID or identity for the purposes of the protocol?

Let's try and keep it at `identity` for consistency with others. 

Practically, it will be scoped to per participant session as in-memory would flush when a participant session closes. So, a reconnect means the client would have to send the schema again.

But, would be good to keep protocol at `identity` for consistency.

protobufs/livekit_models.proto
boks1971: I think this type should be removed and separate fields set.

To summarise
- `DataTrackSchemaDefinition` contains the full definition (includes `id` and `definition`)
- `DataTrackSchemaId` encapsulates the `id` cleanly so that it can be used in `Get...`
- I think the `DataTrackInfo` should have `DataTrackSchemaId` (for schema id reference) and `DataTrackFrameEncoding` as a separate field. Combining the two feels like the reference to schema id is tied to frame encoding which is a bit high cognitive load.

BTW, where does frame encoding come from? Will it be in `AddDataTrackRequest`?

protobufs/livekit_rtc.proto
```diff
@@ -297,6 +307,22 @@ message UpdateDataSubscription {
...
 
+message DefineDataTrackSchemaRequest {
+  // Schema to define.
+  DataTrackSchemaDefinition schema_definition = 1;
+}
+
+message GetDataTrackSchemaRequest {
+  string participant_identity = 1;
+  // ID of the schema to get the definition for.
+  DataTrackSchemaId schema_id = 2;
```
paulwe: `[(logger.name) = "schemaID"]`?

protobufs/livekit_models.proto
```diff
@@ -286,6 +286,89 @@ message DataTrackInfo {
...
+        // MessagePack: self-describing.
+        WELL_KNOWN_FRAME_ENCODING_MSGPACK = 6;
+        // JSON: self-describing or described by `JSON_SCHEMA` schema encoding.
+        WELL_KNOWN_FRAME_ENCODING_JSON = 7;
+    }
+
+    oneof value {
+        WellKnownFrameEncoding well_known = 1;
+        // Identifier of a custom encoding not covered by the well-known cases.
+        // This must be non-empty and no longer than 25 characters.
```
boks1971: can we make these 32, nothing special there except power of 2 :-)

protobufs/livekit_models.proto
```diff
@@ -286,6 +286,89 @@ message DataTrackInfo {
...
+      // OMG IDL: describes `CDR` frame encoding.
+      WELL_KNOWN_SCHEMA_ENCODING_OMG_IDL = 6;
+      // JSON Schema: describes `JSON` frame encoding.
+      WELL_KNOWN_SCHEMA_ENCODING_JSON_SCHEMA = 7;
+    }
+
+    oneof value {
+        WellKnownSchemaEncoding well_known = 1;
+        // Identifier of a custom encoding not covered by the well-known cases.
+        // This must be non-empty and no longer than 25 characters.
```
boks1971: same here.

protobufs/livekit_models.proto
```diff
@@ -286,6 +286,89 @@ message DataTrackInfo {
 
   // Method used for end-to-end encryption (E2EE) on packet payloads.
   Encryption.Type encryption = 4;
+
+  // Encoding for frame payloads on this track. If unspecified, the track is untyped.
+  DataTrackFrameEncoding frame_encoding = 5;
```
boks1971: should this be `optional` too?

boks1971 (approved):

protobufs/livekit_models.proto
ladvoc: Yes, addressed in [3462ce1](https://github.com/livekit/protocol/pull/1553/commits/3462ce14966c22cf83bd898c2e9e770b3dac220c)

PR #1556, opened 2026-05-13
Add TokenPagination field to ListEgress API and RPC

biglittlebigben:


protobufs/livekit_egress.proto
```diff
@@ -334,10 +334,14 @@ message ListEgressRequest {
   string room_name = 1;                               // (optional, filter by room name)
   string egress_id = 2 [(logger.name) = "egressID"];  // (optional, filter by egress ID)
   bool active = 3;                                    // (optional, list active egress only)
+  TokenPagination page_token = 4;
```
paulwe, line 337: we probably want `Pagination` unless we don't want to allow users to set `limit` lower than the default

milos-lk (approved):

protobufs/livekit_egress.proto
biglittlebigben: Ingress is using TokenPagination for the same purpose. I believe we've been leaning towards TokenPagination in most APIs recently to keep the token anonymous, which allows us to change its semantics more easily. The exception would be if there is a reason to allow the caller to define offsets or page sizes other than the default. I don't believe there is a strong incentive to do this here.

paulwe: if we set the page size too big we're overscanning for customers who don't care and if we set it too small they're just going to hammer the api until they get what they want and our api quotas are so generous as to be basically useless for preventing abuse...

biglittlebigben: That would be an argument to add a limit parameter to the request. We can add it at the top level while keeping token pagination which has the pro of making the token anonymous, and easier to change if needed. The 2 cursors particularly are confusing and users may think they need to provide them somehow. 

That being said, I am not convinced we should add a limit field proactively. From the customer's perspective, there is little incentive to request pages smaller than the maximum we are willing to provide so we should probably just return whatever makes sense to us (currently 300 items).

paulwe: *300???* 

this is a very blunt instrument... improving the api semantics would probably buy us a lot more than changing the internals

paulwe (approved):

PR #1565, opened 2026-05-15
TEL-336: fix sip error categorization

hechen-eng:


dennwc (approved):
Good catch! Please don't forget to add a changeset.

PR #1582, opened 2026-05-25
export missing proto messages

chenghao-mou:


davidzhao (approved):

theomonnom (approved):

lukasIO (commented):
you'll need to add the same exports to the `index.d.ts` file, too

PR #1583, opened 2026-05-26
Add Krisp Viva to FeatureUsageInfo enum

lukasIO:


protobufs/livekit_analytics.proto
```diff
@@ -232,6 +233,7 @@ message FeatureUsageInfo {
     KRISP_NOISE_CANCELLATION = 0;
     KRISP_BACKGROUND_VOICE_CANCELLATION = 1;
     AIC_AUDIO_ENHANCEMENT = 2;
+    KRISP_VIVA_VOICE_ISOLATION = 3;
```
davidzhao: should we just use KRISP_VIVA here and then set specifics (voice isolation, noise cancellation, other features) in `feature_info`?

this is how we are currently handling AIC

lukasIO: Yeah, that makes sense to me.

davidzhao (approved):

PR #1595, opened 2026-06-01
Data blobs

ladvoc:
To support data tracks schema metadata (see #1553), this PR introduces a general purpose mechanism for storing large (in the order of KBs), arbitrary data blobs in a room that can be retrieved by other participants on demand.


boks1971 (approved):
lgtm! can we hold merging till integration on server side and getting feedback on server side changes? Just want to make sure if there are other changes, it will all be in one PR.

protobufs/livekit_rtc.proto
```diff
@@ -575,6 +596,8 @@ message RequestResponse {
     UpdateLocalVideoTrack update_video_track = 9;
     PublishDataTrackRequest publish_data_track = 10;
     UnpublishDataTrackRequest unpublish_data_track = 11;
+    StoreDataBlobRequest store_data_blob = 12;
```
ladvoc: @boks1971, I just realized adding this case to _RequestResponse_'s _request_ oneof is not a good idea since that means the entire blob contents will be echoed back in case of error. Is the best solution to omit it here and add request ID fields to associate request and response?

boks1971: All request responses include the response. We can special case this and not include the request and use the request Id. I think single error with large request is okay, but if somebody spams, then it is problematic.

boks1971:
lgtm! ✅ once the conflicts are resolved.

PR #1597, opened 2026-06-02
fix nil != empty audio config

frostbyte73:


protobufs/livekit_egress.proto
```diff
@@ -114,9 +115,8 @@ message ParticipantVideo {
 // Unified audio selection and channel routing.
 // Each route specifies both which audio to capture and which channel to output to.
 message AudioConfig {
-  // If empty, all audio captured in both channels.
-  // If non-empty, only matching audio is captured and routed. Unmatched is excluded.
-  repeated AudioRoute routes = 1;
+  bool capture_all = 1;
```
boks1971: does `capture_all` take precedence? Would be good to have a line of comment on precedence here.

boks1971 (approved):

PR #1606, opened 2026-06-04
deps: migrate jwt/xsync/errors libraries + add jwtutil helper

paulwe:
## Summary

Source-level migration off three unmaintained / breaking-changed deps, plus an internal helper to insulate downstream consumers from JWT library churn.

- **go-jose/v3 -> golang-jwt/jwt/v5** (`auth/accesstoken.go`, `auth/verifier.go`, tests). Took a brief detour through go-jose/v4 (4ec17943) before switching to golang-jwt (878cf9bb) because go-jose/v4 enforces an HS256 >=32-byte secret minimum that breaks the OSS `livekit-server --dev` default. golang-jwt/v5 matches v3's permissive behavior and is ~18x more popular.
- **xsync/v3 -> v4** — `MapOf` is now an alias for `Map`; `LoadOrCompute` callback signature changed to `func() (V, bool)` (the new bool is "cancel").
- **pkg/errors -> stdlib** `errors` / `fmt.Errorf %w` (b6c88bd4). Also dropped `go.uber.org/atomic` in the same commit, then reverted (ab1b54f9) when it produced ABI friction with downstream cloud-protocol struct fields.
- **`utils/jwtutil`** (29568b8d) — new package wrapping `*jwt.Parser` + an `HMACKeySet` for kid-based key rotation, so cloud-protocol/s2sa and other consumers don't each re-implement the verifier.
- **webhook test fixture** (658576c8) — bumped the HS256 secret in tests to satisfy go-jose v4's length minimum during the intermediate commit; left in place after the golang-jwt switch since longer secrets are still better.
- `go get -u` sweep + counterfeiter v6.12.2 (a7a83da5).

## Downstream impact

- `auth.APIKeyTokenVerifier` internals changed (no longer holds a `*jose.JSONWebToken`); public API is unchanged.
- `utils/jwtutil` is the new recommended way to verify JWTs against a rotated key set; cloud-protocol's worker token provider already consumes it.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` (modulo pre-existing Docker/Postgres/AWS-creds-gated tests)


boks1971 (approved):

logger/pionlogger/zaplogadapter.go
```diff
@@ -17,9 +17,9 @@ package pionlogger
 import (
 	"fmt"
 
-	"go.uber.org/atomic"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"sync/atomic"
```
biglittlebigben: nit: these imports should go below "fmt" in the stdlib import block

biglittlebigben (approved):
Thanks!

PR #1609, opened 2026-06-07
agent_simulation: scenario yaml source-of-truth + SimulationFinalize

theomonnom:


protobufs/agent/livekit_agent_session.proto
```diff
@@ -290,6 +290,12 @@ message SessionRequest {
     optional Input input = 1;
     optional Output output = 2;
   }
+  // Sent by a simulation controller after it has computed a provisional
+  // verdict, giving the agent under test a chance to override it.
+  message SimulationFinalize {
```
davidzhao:
```suggestion
  message FinalizeSimulation {
```

following the pattern of Action-Subject

davidzhao (approved):

PR #1611, opened 2026-06-09
Relax SIP outbound validation to support carriers.

dennwc:


livekit/sip.go
```diff
@@ -764,8 +764,8 @@ func (p *UpdateSIPDispatchRuleRequest_Update) Apply(info *SIPDispatchRuleInfo) (
 }
 
 func (p *CreateSIPParticipantRequest) Validate() error {
-	if p.SipTrunkId == "" && p.Trunk == nil {
-		return errors.New("missing sip trunk id")
+	if p.SipTrunkId == "" && p.Trunk == nil && p.SipNumber == "" {
+		return errors.New("missing sip trunk id or sip number to use")
```
hechen-eng: `missing sip trunk id and sip number`?

hechen-eng (approved):

livekit/sip.go
nishadmusthafa: @dennwc we also need to check that this is belonging to the account right ?

dennwc: This is a protocol package, it cannot make this decision here. It will be done in IO.

PR #1613, opened 2026-06-10
agent_simulation: per-run concurrency + SimulationMode on dispatch

theomonnom:
## What

- `SimulationRun.Create.Request.concurrency` (optional int32): max simulate jobs running in parallel for a run. Clamped server-side to the project's concurrent-simulate quota; 0/unset keeps the scheduler default. Echoed back on `SimulationRun`.
- `SimulationMode` enum (`TEXT` / `AUDIO`) and `SimulationDispatch.mode`: tells the agent under test how the simulated user interacts, so e.g. text simulations can skip STT/TTS. Simulations are text-only today; `AUDIO` is plumbing for upcoming work.

## Notes

- Stacked on #1610 (agent_description drop), hence the base branch.
- Consumers: agents-private#68 (claim filter + dispatch), cloud-api-server#1892 (passthrough), livekit-cli#870 (`--concurrency`), livekit/agents#6036 (skip STT/TTS on text sims).
- Field 5 in `Create.Request` is skipped (unclear history); concurrency uses 8.


protobufs/livekit_agent_simulation.proto
```diff
@@ -191,12 +197,24 @@ message ScenarioGroup {
...
 
+// How the simulated user interacts with the agent under test.
+enum SimulationChannel {
+  SIMULATION_CHANNEL_UNSPECIFIED = 0;
+  // The simulator chats over text streams (no audio).
+  SIMULATION_CHANNEL_TEXT = 1;
+  // The simulator publishes/subscribes audio in the room.
+  SIMULATION_CHANNEL_AUDIO = 2;
+  // The simulator dials the agent via a SIP phone number.
+  SIMULATION_CHANNEL_SIP = 3;
```
davidzhao: how is this being used? if simulator needs to dial the agent, shouldn't it need to provide phone numbers/etc? where does the details come from?

theomonnom: This isn't implemented yet, for now only TEXT is available

davidzhao (approved):

PR #1620, opened 2026-06-12
Add codec config to inbound and outbound SIP trunks.

dennwc:


hechen-eng (commented):
The tests are failing due to `Upgrade` will create `p.Media` even if there is nothing to migrate 🤔

Maybe do
```
if p == nil || (p.Media == nil && p.MediaEncryption == SIPMediaEncryption_SIP_MEDIA_ENCRYPT_DISABLE) {
    return
}
```

hechen-eng (approved):

PR #1621, opened 2026-06-12
Protocol updates for data streams v2

1egoman:
Adds a few new fields required for data streams v2.

This project improves two main things:
1. Adds the ability to send a single packet data stream for data streams with a very small payload (ie, rpc v2, small agent messages, etc). This can shrink a data stream from 3 packets to 1 packet, which on devices with poor network connectivity like phones can significantly boost delivery rates.
2. Adds opt-in compression for clients which have hardware compression support. After doing a bunch of benchmarking / research, something DEFLATE based looked like it had the best set of tradeoffs (decent compression ratios / hardware acceleration, plus wide built in platform support so we don't need to include large compression libraries in each client), and in particular "raw" DEFLATE drops extra header and checksum metadata which shouldn't be needed on a reliable / ordered SCTP packet stream.

Both of these improvements (in the right circumstances) have experimentally resulted in a 2x throughput increase to data streams 🎉 , and will have a significant impact in reducing data channel congestion.


1egoman:
(the diff is kinda cluttered, all the changes are in `protobufs/livekit_models.proto`)

protobufs/livekit_models.proto
```diff
@@ -804,6 +805,15 @@ message DataStream {
     string name = 1;
   }
 
+  // The compression type of the whole data stream
+  //
+  // This will only get populated when send to participants with a
+  // client protocol >= 2 which advertise a client capability of CAP_COMPRESSION_DEFLATE_RAW
+  enum CompressionType {
+    NONE = 0;
+    DEFLATE_RAW = 1;
```
xianshijing-lk: can you add a comment to explain what this "DEFAULT_RAW" mean ? does it mean compressed deflate bitstream, no header and no checksum/trailer ? 

is RAW_DEFLATE or DEFLATE_RAW better term here ?

boks1971 (approved):

lukasIO (approved):

protobufs/livekit_models.proto
1egoman: > can you add a comment to explain what this "DEFAULT_RAW" mean ?

From the PR description:

> After doing a bunch of benchmarking / research, something DEFLATE based looked like it had the best set of tradeoffs (decent compression ratios / hardware acceleration, plus wide built in platform support so we don't need to include large compression libraries in each client), and in particular "raw" DEFLATE drops extra header and checksum metadata which shouldn't be needed on a reliable / ordered SCTP packet stream.

I will copy this over and include it in the protobuf file as well.

> is RAW_DEFLATE or DEFLATE_RAW better term here ?

I don't have a strong preference, but [the web uses `deflate-raw`](https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream/CompressionStream#deflate-raw) which is why I called it that.

1egoman: Done in dd9e96a

PR #1625, opened 2026-06-14
agent: add tool status events to AgentSession

longcw:
Adds a `ToolStatusUpdated` event to `AgentSessionEvent` so the agent SDK can stream per-call tool lifecycle to remote sessions (frontends, simulators).

The event carries one of three updates: `Started` (a tool call was dispatched), `CallUpdated` (a progress update or the terminal entry — running/done/error/cancelled), and `ReplyUpdated` (the deferred reply that voices buffered updates — scheduled/completed/interrupted/skipped).


protobufs/agent/livekit_agent_session.proto
```diff
@@ -249,6 +263,37 @@ message AgentSessionEvent {
     AgentSessionUsage usage = 1;
   }
 
+  // Per-call tool lifecycle. One flat status update per occurrence:
+  // a call started -> the call updated (progress and the terminal entry) ->
+  // the deferred reply voicing buffered updates advanced.
+  message ToolStatusUpdated {
+    message Started {
```
theomonnom, line 270: Just for completeness, should we also have Ended?

theomonnom (approved):

protobufs/agent/livekit_agent_session.proto
longcw: CallUpdated with `done/error/cancelled` status is the end of the tool call, do you think we should have a separate event for that?

theomonnom: Yeah even if it's duplicating the logic, it's better for consistency IMO

theomonnom: You can keep CallStatus but add Ended? or Completed

PR #1626, opened 2026-06-15
redis: add credentials provider option and built-in Entra ID auth

cosmin-staicu:
Adds support for runtime-rotated Redis credentials to `GetRedisClient`, with two entry points: a vendor-neutral `WithStreamingCredentialsProvider` option and a config-driven `azure_entra` flag for Microsoft Entra ID. Backwards compatible — existing callers and YAML are unaffected.

##

`GetRedisClient` currently only accepts a static `Username`/`Password`. That doesn't work with managed Redis offerings that authenticate via short-lived identity tokens — **Azure Managed Redis** and **Redis Enterprise Cloud** issue Microsoft Entra ID tokens that expire (~1h) and must be refreshed and re-`AUTH`'d on the live connection before they lapse.

With only static credentials, a long-running deployment gets disconnected when the token expires. It's worst for the coordination bus that LiveKit relies on: managed Redis evicts pub/sub connections when their token expires, and under RESP2 a pub/sub socket can't process an out-of-band `AUTH`, so it must reconnect with a fresh token. Every component that builds its client through `protocol` — server, sip, egress, ingress — is affected.

go-redis already supports this through `StreamingCredentialsProvider` (≥ v9.6; this module is on v9.20.0), and Redis publishes an official extension, [`go-redis-entraid`](https://github.com/redis/go-redis-entraid), that implements


cosmin-staicu:
Rebased onto current `main` (was conflicting) and split: this PR is now the redis commit alone, with a changeset added.

The log-config-watch work that had crept onto this branch has moved to #1791, which is stacked on this one.

PR #1627, opened 2026-06-15
agent_simulation: add audio mode

Bobronium:
## What

- `SimulationRun.Create.Request.mode` (`SimulationMode`, field 9): lets a caller
  choose the conversation channel for a run. Unspecified = `TEXT`, so existing
  callers are unaffected.
- `SimulationRun.mode` (field 18): echoes the chosen mode back on Get/List, so
  the dashboard / CLI can display it and detect a run that silently fell back
  to `TEXT` against an older server.

## Notes

- Extends #1613, which added the `SimulationMode` enum and
  `SimulationDispatch.mode` (the read side, consumed by the agent under test).
  1613 made `AUDIO` *defined* but unreachable: nothing could set it but the
  server, which hardcodes `TEXT`. This PR adds the request→readback path so a
  caller's choice reaches the dispatch.
- Consumers: https://github.com/livekit/agents-private/pull/102, https://github.com/livekit/livekit-cli/pull/876.


protobufs/livekit_agent_simulation.proto
```diff
@@ -95,6 +95,9 @@ message SimulationRun {
       // Maximum simulate jobs running in parallel for this run. Clamped to the
       // project's concurrency quota; 0/unset uses the server default.
       optional int32 concurrency = 8;
+
+      // Conversation channel for every job in this run; unspecified = TEXT.
```
davidzhao: nit:

```suggestion
      // Conversation mode for every job in this run; unspecified = TEXT.
```

protobufs/livekit_agent_simulation.proto
```diff
@@ -160,6 +163,8 @@ message SimulationRun {
   Usage usage = 16;
   // Maximum simulate jobs running in parallel for this run (0 = server default).
   int32 concurrency = 17;
+  // Conversation channel for every job in this run; unspecified = TEXT.
```
Bobronium:
```suggestion
  // Conversation mode for every job in this run; unspecified = TEXT.
```

theomonnom (approved):
lgtm!

PR #1644, opened 2026-06-24
agent-simulate: add JobMetrics/RunMetrics to SimulationRun

Bobronium:
Adds quality metrics for agent simulations: per-job `SimulationRun.JobMetrics`
and run-level `SimulationRun.RunMetrics` aggregates.


protobufs/livekit_agent_simulation.proto
```diff
@@ -72,13 +72,152 @@ message SimulationRun {
...
+    repeated Metric metrics = 3;
+
+    // Conversation timeline: one entry per transcript turn, both speakers.
+    repeated Turn turns = 4;
+
+    string metrics_version = 5;         // comparability key; manual ("1")
+    string suite_signature = 6;         // exact code fingerprint; debug only
+    string judge_model = 7;             // "" if no text judge ran
+    string audio_judge_model = 8;       // "" if no audio judge ran
+    bool has_remote_session = 9;        // false = waveform-only capture (SIP)
```
theomonnom:
```suggestion
  message JobMetrics {
    float accuracy_score = 1;    // did the agent achieve the scenario outcome
    float experience_score = 2;  // quality of the caller experience
    
    
    float stt_wer = 3;
    float stt_transcription_latency = 4;
    float tts_wer = 5;
```

Bobronium: Folded back to concrete fields.

I think we should still divide metrics per types instead of relying on prefixes. This also allows reusing the types per run instead of re-declaring every field:

https://github.com/livekit/protocol/blob/b17db04611b6d37971d21a4d29ce235325a71440/protobufs/livekit_agent_simulation.proto#L195-L208

protobufs/livekit_agent_simulation.proto
```diff
@@ -72,13 +72,135 @@ message SimulationRun {
...
+    // Conversation timeline: one entry per transcript turn, both speakers.
+    repeated Turn turns = 9;
+
+    string judge_model = 10;            // text judge for judged scores; "" if none ran
+    string audio_judge_model = 11;      // audio (LALM) judge; "" if none ran
+    bool has_remote_session = 12;       // false = waveform-only capture (e.g. SIP)
+    google.protobuf.Timestamp t0 = 13;  // call start (first speech edge); turn times relative to it
+
+    // The agent's perception of the caller.
+    message Stt {
```
theomonnom:
```suggestion
    message STT {
```
I would use uppercase IMO, nit tho

theomonnom (approved):
lgtm!

protobufs/livekit_agent_simulation.proto
Bobronium: I would too, but the internet convinced me otherwise: https://www.thingsaboutweb.dev/en/posts/naming-with-acronyms. 

However, I've checked the repo and we're already using 3 and more letters uppercase names (SIP, etc.), so using uppercase here makes it consistent.

Renamed to LLM/TTS/STT, thanks for highlighting it!

PR #1648, opened 2026-06-29
TEL-730: validate malformed transfer to return early

hechen-eng:


alexlivekit (changes requested):
code/protocol/livekit/sip_validation.go and code/protocol/livekit/sip_validation_test.go would probably be better places for this.

I actually have a PR coming up for this, so let me know how we're playing that.

hechen-eng:
> code/protocol/livekit/sip_validation.go and code/protocol/livekit/sip_validation_test.go would probably be better places for this.
> 
> I actually have a PR coming up for this, so let me know how we're playing that.

sounds good 👍

PR #1649, opened 2026-06-29
Use camel case log name in `DataBlobKey`

ladvoc:
Follow-up to #1553

### DIFF
--- a/protobufs/livekit_rtc.proto
+++ b/protobufs/livekit_rtc.proto
@@ -326,7 +326,7 @@ message StoreDataBlobResponse {
 message GetDataBlobRequest {
   uint32 request_id = 1 [(logger.name) = "requestID"];
   // Identity of the participant who owns the blob.
-  string participant_identity = 2;
+  string participant_identity = 2 [(logger.name) = "participantIdentity"];
   // Unique key of the data blob to retrieve.
   DataBlobKey key = 3;
 }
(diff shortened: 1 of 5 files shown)


protobufs/livekit_rtc.proto
```diff
@@ -326,7 +326,7 @@ message StoreDataBlobResponse {
 message GetDataBlobRequest {
   uint32 request_id = 1 [(logger.name) = "requestID"];
   // Identity of the participant who owns the blob.
-  string participant_identity = 2;
+  string participant_identity = 2 [(logger.name) = "participantIdentity"];
```
boks1971: don't think we need for this.

There was a `schema_id`, that needs to be `schemaID` logger name.

ladvoc: Makes sense, added in [16b5572](https://github.com/livekit/protocol/pull/1649/commits/16b5572f2fa7bccc5570b0622a41c1eb45a35ff7). Is the standard every field that gets logged in a request/response message should be camal case? Happy to revert for `participant_identity` if it's not necessary.

boks1971: This is not needed. We do not have this anywhere else. It is just that we want `ID` upper case that needs the logger name option.

boks1971 (approved):

PR #1650, opened 2026-06-29
TEL-730: Adding name-addr validation

alexlivekit:


livekit/sip_validation.go
```diff
@@ -315,7 +325,7 @@ func validateNameAddrHeader(value string) error {
 		}
 	} else {
 		// This is a bare URI, and should comply with addr-spec, no special characters
-		if strings.ContainsAny(value, ";,? ") {
```
alexlivekit, line 318: A semicolon here would mean that what's after that is header params, not the URI. However, it is still valid syntax

livekit/sip_validation.go
```diff
@@ -253,10 +260,13 @@ func ValidateHeaderValue(name, value string) error {
...
-	if _, exists := nameAddrHeaders[lowerName]; exists && false {
-		// TODO: Disabled since all supported headers are forbidden, re-enable when we allow some
+	if _, exists := nameAddrHeaders[lowerName]; exists {
 		if err := validateNameAddrHeader(value); err != nil {
-			return fmt.Errorf("header %s: value: %w", name, err)
+			err = fmt.Errorf("header %s: value: %w", name, err)
+			// For now do not actually error out on header values, just note failure.
+			if cb := softFailureReporter; cb != nil {
+				cb(err)
+			}
```
alexlivekit, line 269: Should we do `} else { return err }` here? Was considering it.

alexlivekit (commented):
We definitely don't want to start off with strict erroring. This will allow us to at least log errors and understand the frequency and impact of something like this before customer start seeing failures.

livekit/sip_validation.go
hechen-eng: why don't we return error when we know the validation failed?

alexlivekit: Because if some customers are using the system with invalid values today, we don't want to cause an outage when this lands. So we "soft-fail" first, allowing us to log the would-be errors for a while before we turn this on.

hechen-eng: not sure how would that be an outage since those calls are already failing.. 🤔

hechen-eng (approved):
have some comments, but nothing blocking

livekit/sip_validation.go
alexlivekit: We would be enabling this check not just for refer messages.

PR #1662, opened 2026-07-09
Dynacast support for data tracks

ladvoc:
Closes BOT-446

### DIFF
--- a/protobufs/livekit_models.proto
+++ b/protobufs/livekit_models.proto
@@ -292,9 +292,6 @@ message DataTrackInfo {
 
   // ID of the schema used by frames on this track if the track is typed.
   optional DataTrackSchemaId schema = 6;
-
-  // Whether dynacast is enabled for this track.
-  bool is_dynacasted = 7;
 }
 
 // Encoding for frame payloads.
--- a/protobufs/livekit_models.proto
+++ b/protobufs/livekit_models.proto
@@ -292,6 +292,9 @@ message DataTrackInfo {
 
   // ID of the schema used by frames on this track if the track is typed.
   optional DataTrackSchemaId schema = 6;
+
+  // Whether dynacast is enabled for this track.
+  bool is_dynacasted = 7;
 }
 
 // Encoding for frame payloads.
--- a/protobufs/livekit_rtc.proto
+++ b/protobufs/livekit_rtc.proto
@@ -236,9 +236,8 @@ message DataTrackDemandUpdate {
   // Publisher handle of the track.
   uint32 pub_handle = 1;
 
-  // Number of participants currently subscribed to the track. Used
-  // by the publisher when dynacast is enabled for the track.
-  uint32 subscriber_count = 2;
+  // Whether or not the track currently has any subscribers.
+  bool has_subscribers = 2;
 }
 
 message TrickleRequest {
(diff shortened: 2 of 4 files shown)


protobufs/livekit_models.proto
```diff
@@ -292,6 +292,9 @@ message DataTrackInfo {
 
   // ID of the schema used by frames on this track if the track is typed.
   optional DataTrackSchemaId schema = 6;
+
+  // Whether dynacast is enabled for this track.
+  bool is_dynacasted = 7;
```
boks1971: I would drop this if not used. Easier to add later if needed. Just thinking about the feature, this should be needed in the model?

protobufs/livekit_rtc.proto
```diff
@@ -202,6 +204,9 @@ message PublishDataTrackRequest {
   // ID of the schema used by frames on this track if the track is typed.
   // If set, the associated schema must be stored with `StoreDataBlobRequest`.
   optional DataTrackSchemaId schema = 5;
+
+  // Whether dynacast is enabled for this track.
+  bool is_dynacasted = 6;
```
boks1971: Same here.

protobufs/livekit_rtc.proto
```diff
@@ -230,6 +235,15 @@ message DataTrackSubscriberHandles {
   map<uint32, PublishedDataTrack> sub_handles = 1;
 }
 
+message DataTrackDemandUpdate {
+  // Publisher handle of the track.
+  uint32 pub_handle = 1;
+
+  // Number of participants currently subscribed to the track. Used
+  // by the publisher when dynacast is enabled for the track.
+  uint32 subscriber_count = 2;
```
boks1971: this can be a boolean, I think counter is also fine, but it may not be accurate always, but that should be fine.

protobufs/livekit_models.proto
ladvoc: Makes sense, addressed in [eff1757](https://github.com/livekit/protocol/pull/1662/commits/eff1757e4ebb7244fb1b74b8682da6eda592f6f6).

protobufs/livekit_rtc.proto
ladvoc: Also addressed in [eff1757](https://github.com/livekit/protocol/pull/1662/commits/eff1757e4ebb7244fb1b74b8682da6eda592f6f6).

protobufs/livekit_rtc.proto
ladvoc: Addressed in [3544490](https://github.com/livekit/protocol/pull/1662/commits/35444902a9db2f079bad79030f00ab73fc612752).

ladvoc: Just curious, when you say it may not always be accurate, do you mean it might take time to become consistent with the actual subscriber count?

boks1971: Yeah. I think "not always accurate" is a bit of poor framing on my part. What I meant is that if the room is spread across many regions, inter-region messages have to propagate to the publisher node indicating interest from remote nodes and then back to the publisher. I say "poor framing" because that notification has to traverse the network. So, there is no magic to instantly get the subscriber count. In a single node case, it is instantaneous, so it will look right all the time. That's what I meant by not always accurate if it is not from a single node room.

The other point about using `boolean` is that you just have to send it when count becomes non-zero or zero. But, if you are using count, you have to technically send it anytime it changes. Again, that does not make it inaccurate, but inefficient for the purposes of dynacast, i. e. too many notifications.

ladvoc: Thanks for the info!

PR #1666, opened 2026-07-14
TEL-730: Update various SIP validation methods to return error slices.

alexfish8:
Motivation: we'd like like to get a better sense as to which clients are sending invalid REFER messages (see [TEL-730](https://linear.app/livekit/issue/TEL-730/lost-refer-messages)). In order to do so, we're updating various SIP validation methods to return slices of errors. Some of these errors might be "soft errors" (i.e., encountering them shouldn't cause the surrounding request to fail). Previously, these soft errors were could be iterated over via a global `SoftFailureNotifier`. The global nature of this notifier made it cumbersome to log these soft errors with request-scoped loggers. This PR updates the validation methods to include these soft errors in their return values so that callers may log them as they please.

A few additional notes:
- In general, I've tried to avoid updating `err` return values to `[]error` unless I needed to do so.
- Additionally, so as to avoid introducing unintended behavior changes, I've tried to keep all return paths consistent in functions where I've updated the return type signature (i.e., if, previously, an err was returned eagerly, now, we'd eagerly return an error slice consisting of just that error). In the future, we might consider updating some of these methods to return a collection of all validation failures, but, I'd like to avoid introducing a lot of changes all at once.
- Add a few helper functions to extract "hard errors" from these error slices (thanks @alexlivekit for the suggestion on `LogSoftErrors`).

TODO: Maintain backwards compatibility so that we don't have to do a major version bump.

### DIFF
--- a/livekit/sip.go
+++ b/livekit/sip.go
@@ -308,16 +308,17 @@ func (p *SIPOutboundTrunkInfo) AsTrunkInfo() *SIPTrunkInfo {
 }
 
 // validateHeaders makes sure header names/keys and values are per SIP specifications
-func validateHeaders(headers map[string]string) error {
+func validateHeaders(headers map[string]string) []error {
+	var errs []error
 	for headerName, headerValue := range headers {
 		if err := ValidateHeaderName(headerName, true); err != nil {
-			return fmt.Errorf("invalid header name: %w", err)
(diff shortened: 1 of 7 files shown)


alexfish8:
Closed in favor of https://github.com/livekit/protocol/pull/1667.

PR #1667, opened 2026-07-14
TEL-730: Add validation method variants that include soft errors.

alexfish8:
This PR adds `ValidateResult` variants of some of the existing SIP `Validate` methods. Logically, these are the same as the current ones, except they return a `ValidationResult` instead of an `error`, where said `ValidationResult` is composed of an error and zero or more "soft validation errors".

Context: for certain kinds of validation failures, we'd like to log something, but swallow the error, so that the surrounding processing might still continue (we call these "soft errors"). There already exists a mechanism  for logging these "soft" errors (SoftFailureReporter), but the SoftFailure reporter is global, and so it is cumbersome to pass the SoftFailure reporter request-scoped logging context (note: I had originally experimented with extending the SoftFailureReporter function signature to include a `Context` and adding flavors of the `Validate` methods that had a `Context` parameter, but this felt a bit error-prone).

Now, if callers want to log soft validation errors, they may do so by invoking the`ValidateResult` variant of the `Validate` method and logging the soft errors of the returned value.

General notes:

- Initially, when I wrote this, I had added variants of `Validate` that returned an `[]error` rather than a `ValidationResult` (where the error slice contained both hard and soft errors). This was simpler in some ways (avoided introducing a new type), but it became a bit annoying to continually re-check for hard errors in various contexts. This motivated me to add a new type with a simple `OK()` method that could allow callers to conveniently check whether or not they should continue processing or not.
- In the future, we might consider extending `ValidationResult` to contain a slice of hard errors. For now, I've opted to keep behavioral changes to a minimum, so that the validation logic will continue to eagerly return if a hard error is encountered.
- Added various utility methods for logging and combining ValidationResults (thanks @alexlivekit for the `LogSoftErrors` suggestion!).

[Example of how these methods will be used](https://github.com/livekit/livekit/pull/4671)

### DIFF
--- a/livekit/sip_validation.go
+++ b/livekit/sip_validation.go
@@ -19,6 +19,8 @@ import (
 	fmt "fmt"
 	"strconv"
 	"strings"
+
+	"github.com/livekit/protocol/logger"
 )
 
 // RFC 3261 compliant validation functions for SIP headers and messages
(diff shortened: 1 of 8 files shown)


livekit/sip_validation.go
```diff
@@ -212,10 +214,68 @@ var nameAddrHeaders = map[string]bool{
...
-	softFailureReporter = reporter
+// Combine returns a new validation result that is comprised of this result and
+// the other result.
+func (vr ValidationResult) Combine(other ValidationResult) ValidationResult {
+	var err error
+	if vr.err != nil {
+		err = vr.err
+	} else {
+		err = other.err
+	}
```
alexlivekit: NIT: Just use `errors.Join(vr.err, other.err)`. I think it just does this.

livekit/sip_validation.go
```diff
@@ -212,10 +214,68 @@ var nameAddrHeaders = map[string]bool{
...
 
-var softFailureReporter func(err error)
+// Deprecated: has no effect. Use ValidationResult method variants instead.
+func SetSoftFailureReporter(reporter func(err error)) {}
+
+// ValidationResult holds information about something that's been validated.
+type ValidationResult struct {
+	err      error
+	softErrs []error
+}
```
alexlivekit: NIT: I don't think this belongs in `sip_validation.go`, given it's a part of the general interface used in `sip.go`.

alexlivekit (approved):
a few nits, but overall lgtm

livekit/sip_validation.go
alexfish8: Moved this struct and its methods to sip.go (and moved their tests to sip_test.go).

livekit/sip_validation.go
alexfish8: Done.

alexfish8:
> a few nits, but overall lgtm

Thanks for the review!

PR #1701, opened 2026-08-05
Resolve current stack of stuck lock holders

erikhortsch:
## Motivation

During a recent incident, the lock tracker's stuck-lock logs told us where goroutines were *waiting*, but not where the *holder* was blocked. The tracker records the acquisition stack (`FirstLockedAtStack`), which points at where the lock was grabbed — not at the call the holder is currently stuck in. And since holder goroutine ids aren't recorded, their stacks can't be picked out of a goroutine dump either: waiters are self-evident (parked in `sync.Mutex.Lock`), holders are anonymous.

## Changes

**Single-holder tracking for exclusive locks** (`lock_tracker.go`): `Mutex` has exactly one holder and `trackLock`/`trackUnlock` run while it's held, so the holder gid is a plain field — no atomics except the waiter count, and a cross-goroutine unlock simply clears it. Lock paths guard tracking behind one nil check; unlock paths use a bare atomic load (`loadTracker`) instead of the lazy-init.

**Read-holder tracking as an RW extension** (`lock_tracker_rw.go`): `RWMutex` write holds use the embedded single-holder path; concurrent readers use atomic gid slots (up to 8, overflow-counted). Reader accounting removes exactly one entry per RUnlock — own gid, else overflow credit, else arbitrary eviction (the same cross-goroutine-unlock policy go-deadlock uses; compare `HolderGoroutineIDs` with `NumGoroutineHeld` to detect overflow).

**Holder stacks**: `PopulateHolderStacks(stuck)` resolves each holder's **current** stack from a single `runtime.Stack(buf, true)` snapshot. The snapshot stops the world, so consumers run it once per detection episode rather than on every scan.

**StuckLock API**: `HolderGoroutineIDs()`, `HolderStacks()`, and `HolderStrength()` (exclusive vs shared readers).

## Cost (Apple M5 Pro, per Lock/Unlock pair)

| | before | after |
|---|---|---|
| wrapped Mutex | 6.2ns | 6.3ns |
| wrapped RWMutex (write) | ~6.5ns | 8.3ns |
| native sync.Mutex | 4.0ns | 4.0ns |

Holder tracking on exclusive locks is effectively free; the RW read path pays goid+CAS each way.

## Testing

Holder-stack resolution, cross-goroutine unlock, reader overflow, and holder-strength tests; full `./utils` suite passes with `-race`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


utils/lock_tracker.go
paulwe: we can get rid of this?

paulwe (approved):
does this meaningfully impact overhead?

utils/lock_tracker.go
```diff
@@ -195,8 +278,20 @@ func (t *lockTracker) trackLock() {
 
 func (t *lockTracker) trackUnlock() {
 	if t != nil {
+		gid := goid.Get()
```
paulwe: is this guaranteed to be the same g that acquired the lock? this + the max 16 holders for rw locks seems a bit sketchy

boks1971: Given the comment above in `trackLock`, guessing it is one of the goroutines that is holding the lock. So, it should ideally be in one of these slots or an overflow. Wonder what is the case where the overflow removal also fails and a random one is removed from the slots.

boks1971 (approved):
this will be very useful. Thank you @erikhortsch

utils/lock_tracker.go
erikhortsch: It is possible that a lock is acquired from 1 goroutine and then released from another - that feels like a hacky pattern that should probably be avoided. Maybe something that we specifically log on in staging

PR #1706, opened 2026-08-07
fix(auth): require exp claim when verifying access tokens

SashaMIT:
## Problem

`APIKeyTokenVerifier.Verify` parses with `WithValidMethods` / `WithIssuer` / `WithLeeway` but not `WithExpirationRequired`. Under golang-jwt v5, a token carrying **no `exp` claim at all** passes validation and never expires.

First-party SDKs always set `exp` (the Go SDK unconditionally defaults it in `ToJWT`), so this doesn't affect well-behaved minters — but any hand-rolled or third-party minter that forgets `exp` silently mints a permanent credential, and the server accepts it without complaint.

## Fix

Add `jwt.WithExpirationRequired()` to the parse options. One line; tokens with a valid `exp` are unaffected.

## Tests

Added `token without exp is rejected` to `TestVerifier`, using a hand-rolled `jwt.MapClaims` token (the Go SDK always sets `exp`, so the case has to be built directly). `go test ./auth/` passes.

Made with Cursor

Made with [Cursor](https://cursor.com)

### DIFF
--- a/auth/verifier.go
+++ b/auth/verifier.go
@@ -77,6 +77,11 @@ func (v *APIKeyTokenVerifier) Verify(key interface{}) (*jwt.RegisteredClaims, *C
 		jwt.WithValidMethods(allowedSigningMethods),
 		jwt.WithIssuer(v.apiKey),
 		jwt.WithLeeway(tokenLeeway),
+		// Without this, a token carrying no exp claim passes validation and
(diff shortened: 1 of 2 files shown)


cnderrauber (approved):
looks good!

SashaMIT:
> looks good!

Thanks for the quick review and merge @cnderrauber, great to get to contribute!

PR #1716, opened 2026-08-14
Add egress passthrough field, remove unused UpdateEgress RPC

frostbyte73:


milos-lk (approved):

protobufs/livekit_egress.proto
```diff
@@ -61,6 +60,9 @@ message StartEgressRequest {
   oneof encoding {
     EncodingOptionsPreset preset = 5;
     EncodingOptions advanced = 6;
+
+    // Remux a single explicitly-identified track without decoding or re-encoding
```
biglittlebigben, line 64: Should be more explicit about restrictions? (Media source, only track)

biglittlebigben: Not necessary a suggestion, since I haven't thought too deeply about how it would work in practice, but could passthrough be a preset?

frostbyte73: I do like that idea

PR #1717, opened 2026-08-14
TEL-892 Adding to_user_override field

genseric-ghiro:
## Summary
- Adding `to_user_override` field in CreateSIPParticipantRequest proto
- This will allow to change the `user` part of the `To` header in outbound `INVITE`s.
- The alternative, directly adding a new `To` header, led to two `To` headers in the request which created a malformed request


livekit/sip_validation.go
```diff
@@ -188,6 +188,8 @@ var ForbiddenSipHeaderNames = map[string]bool{
 	"record-route":     true,
 	"refer-to":         true, // rfc3515
 	"reply-to":         true,
+	"to":               true, // Set the user part with CreateSIPParticipantRequest.to_user_override
+	"via":              true, // Multiple Via are legal, but a caller's own breaks response routing
```
nishadmusthafa: Might help to do a quick check in DD to see if anyone is actually passing these in as part of CreateSipParticipant.

nishadmusthafa (approved):

livekit/sip_validation.go
genseric-ghiro: Headers are PII/redacted, so we can't search for them in DD. We do get one warning for malformed `To` headers. I checked in DD for the last 30 days, and I only saw them for the one case this PR is about. One caveat about this metric -> if a user adds a `To` header that's correctly formatted, we don't get a warning. For `via`, we don't get a warning at all whether it's malformed or not. An alternative would be to just add a warning for both for now, re-evaluate in a month, and then enforce it. However, given how these can make a request fail, I think it's safe to forbid them and return an error to the user if they try adding them, let me know what you think.

genseric-ghiro: Removed `To` and `Via` from the forbidden list for now. Instead logging a warning for both whether they're malformed or not.

nishadmusthafa (approved):

PR #1718, opened 2026-08-15
passthrough as a preset

frostbyte73:


protobufs/livekit_egress.proto
```diff
@@ -60,9 +60,6 @@ message StartEgressRequest {
   oneof encoding {
     EncodingOptionsPreset preset = 5;
     EncodingOptions advanced = 6;
-
-    // Remux a single explicitly-identified track without decoding or re-encoding
-    bool passthrough = 10;
```
boks1971, line 65: should this marked as `deprecated` and also set a `NEXT_ID` so that this is not used in the future?

protobufs/livekit_egress.proto
```diff
@@ -500,9 +498,6 @@ message ExportReplayRequest {
   oneof encoding {
     EncodingOptionsPreset preset = 7;
     EncodingOptions advanced = 8;
-
-    // Remux a single explicitly-identified track without decoding or re-encoding
-    bool passthrough = 12;
```
boks1971, line 505: same here

protobufs/livekit_egress.proto
frostbyte73: it was added in the previous commit and hasn't been released yet

boks1971 (approved):

PR #1728, opened 2026-08-20
Add the agent HTTP endpoints data plane protocol

theomonnom:
Protocol for the agent HTTP endpoints data plane. The worker connection is WebTransport (QUIC): one session per worker, a control stream carrying the existing `WorkerMessage`/`ServerMessage` exchange, and one QUIC stream per HTTP exchange (QUIC provides multiplexing, per-stream flow control, half-close, and reset natively, so there is no capsule/credit layer).

- `AgentHttp` namespace in `livekit_agent.proto`: endpoint manifest (`AgentEndpoint`: path template, methods, public), kind enum, `AgentEndpointSettings`, and `GoAway`.
- `RegisterWorkerRequest`: `endpoints` manifest, `instance_id` (epoch fencing), `endpoint_protocol`. Response carries negotiated `endpoint_settings`.
- `UpdateWorkerStatus`: `draining` + `seq`; `ServerMessage.go_away` for control-connection drain.
- `ValidateAgentName` + `ValidateDeployment` tightened to a URL-safe charset (both are URL path segments).

Based on the livekit-server go.mod pin; rebases forward with it.


agent/environment.go
```diff
@@ -4,6 +4,27 @@ import "fmt"
...
+// Unlike a deployment, an agent name commonly contains underscores, so they are
+// permitted here.
+func ValidateAgentName(agentName string) error {
+	if len(agentName) > MaxAgentNameLength {
+		return fmt.Errorf("agent name exceeds %d bytes", MaxAgentNameLength)
+	}
+	for i := 0; i < len(agentName); i++ {
+		c := agentName[i]
+		switch {
+		case c >= 'a' && c <= 'z', c >= 'A' && c <= 'Z', c >= '0' && c <= '9', c == '-', c == '.', c == '_':
```
erikhortsch: This seems like something that would be better expressed as a regex?
```
"^[a-zA-Z0-9_.-]+$"
```

PR #1729, opened 2026-08-20
rpc: gate the claim skip process-wide for every server

erikhortsch:
## Summary

Wires psrpc's new `WithServerSkipClaim` (livekit/psrpc#122) into
`WithServerObservability`.

`WithServerObservability` is the one seam every server constructor shares:
`WithDefaultServerOptions` calls it, and the constructors that take only a logger
reach it too. In cloud that is 33 server sites — 23 via `WithDefaultServerOptions`
and 10 via `WithServerObservability` directly — covered by a single edit.

The setting is process-wide rather than per-service because the claim is a
transport policy, and it is read per request, so callers may set it before or
after their servers are constructed and revoke it at runtime without a redeploy.

## Why not per-config

`WithDefaultServerOptions` already receives `PSRPCConfig`, so a field there was
the obvious home — but `PSRPCConfig` is passed by value at construction, which
would lose the runtime kill switch that motivated the client-side version
(livekit/cloud#4567). It also would not reach the 10 logger-only sites. A
process-wide setter keeps both properties.

## Verification

- `go build ./...` — clean
- `go test -race ./rpc/... ./utils/...` — passes
- New `TestServerSkipClaim` covers the three properties the call sites depend on:
  unset means claim, the func is consulted per call rather than captured, and it
  stays revocable.

## Depends on

livekit/psrpc#122 — merged and released as **v0.7.5**, which `go.mod` now pins.
Ready to land.

Note the merged psrpc PR evolved past the revision this description was first
written against: the client option was removed in favor of an always-on
advertisement (new field number, old one reserved), so the server election wired
here is the only switch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rpc/typed_api.go
```diff
@@ -97,12 +98,29 @@ func (p *ClientParams) Args() (psrpc.MessageBus, psrpc.ClientOption) {
 	return p.Bus, psrpc.WithClientOptions(p.Options()...)
 }
 
+var serverSkipClaim atomic.Pointer[func() bool]
```
boks1971: maybe rename this to `psrpcServerSkipClaim`.  I don't think it is super clear even then, but there are so many references to server and client everywhere, hoping it makes at least a bit clearer. And the function names change to `PsRPCSetServerSkipClaim` etc.

boks1971 (approved):

rpc/typed_api.go
erikhortsch: Done in 7674583 — went with `SetPSRPCServerSkipClaim` / `psrpcServerSkipClaim` rather than `PsRPCSetServerSkipClaim`: this package already spells the initialism `PSRPC` (`PSRPCConfig`, `PSRPCMetricsObserver`), and the verb-first setter matches Go convention. Cloud's call site updated in its PR.

PR #1743, opened 2026-08-27
Add realtime billing reporters

russellmartin-livekit:
## Summary
- expose generated reporters for billable realtime text-input messages
- expose generated reporters for discounted cached-image tokens already present in the merged observability schema
- keep both dimensions distinct from operational request counts and gross image input

## Test plan
- [x] `GOWORK=off go test ./observability/gatewayobs`
- [x] `git diff --check origin/main...HEAD`

### DIFF
--- a/observability/gatewayobs/gen_reporter.go
+++ b/observability/gatewayobs/gen_reporter.go
@@ -6,7 +6,7 @@ import (
 	"time"
 )
 
-const Version_F2D2A50 = true
+const Version_3V1GTG0 = true
 
 type KeyResolver interface {
 	Resolve(string)
@@ -118,6 +118,7 @@ type modelReporter interface {
 	ReportRealtimeOutputAudioTokens(v uint64)
 	ReportRealtimeCachedTextTokens(v uint64)
 	ReportRealtimeCachedAudioTokens(v uint64)
+	ReportRealtimeCachedImageTokens(v uint64)
 	ReportRealtimeTextInputMessages(v uint64)
 	ReportSttDuration(v uint32)
 	ReportTtsChars(v uint32)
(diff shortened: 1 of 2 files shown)


shawnfeldman (approved):
Reviewed the generated observability surface.

**Verified**
- Both new methods (`ReportRealtimeCachedImageTokens`, `ReportRealtimeTextInputMessages`) land on `noopModelReporter` *and* `noopModelTx` — the only two `modelReporter` implementers in this repo, so the noop stays complete.
- The `Version_J1BFFVO` → `Version_3V1GTG0` handshake does its job: `backend-common/observability/gatewayobsimpl/gen_reporter.go` pins the old const, so a consumer that forgets to regenerate fails to compile instead of silently dropping columns. backend-common#1053 already flips it and implements both methods on `modelReporter` + `modelTx`.
- `CachedImageTokens` pairs with the existing `InputImageTokens` the same way `CachedText`/`CachedAudio` pair with their gross counterparts.

**Suggestion — naming (fix belongs upstream)**
`ReportRealtimeTextInputMessages` inverts the word order every sibling uses: `RealtimeInputTextTokens`, `RealtimeInputAudioTokens`, `RealtimeInputImageTokens`, `RealtimeInputVideoTokens`. `ReportRealtimeInputTextMessages` would match. It comes from the cloud-observability schema, so not this PR's job — but worth doing before `realtime_total_text_input_messages` is load-bearing in ClickHouse, since renaming a billing dimension later is a migration.

**Merge order (not a defect here)**
`agent-gateway/pkg/testutil/obsfake/reporter.go` hand-implements `gatewayobs.ModelReporter` with no embedded noop (`struct{ parent *Reporter }`, 22 methods), so agent-gateway's test build breaks on the pin bump until obsfake gains these two. agent-gateway#1262 already touches that file.

Approving.

PR #1744, opened 2026-08-27
auth: add AgentGrant.DispatchAdmin

erikhortsch:
Adds `AgentGrant.DispatchAdmin`, granting access to manage a project's agent dispatch
queue.

The service it authorizes is defined in cloud-protocol; only the grant type is shared, so
only the grant lands here.

### DIFF
--- a/.changeset/agent-dispatch-admin-grant.md
+++ b/.changeset/agent-dispatch-admin-grant.md
@@ -1,0 +1,6 @@
+---
+"github.com/livekit/protocol": patch
+"@livekit/protocol": patch
+---
+
+Add `AgentGrant.DispatchAdmin`
--- a/auth/grants.go
+++ b/auth/grants.go
@@ -583,6 +583,8 @@ type AgentGrant struct {
 	SimulationAdmin bool `json:"simulationAdmin,omitempty"`
 	// DatabaseAdmin grants access to a project's agent databases (AgentDB).
 	DatabaseAdmin bool `json:"databaseAdmin,omitempty"`
+	// DispatchAdmin grants access to manage a project's agent dispatch queue.
+	DispatchAdmin bool `json:"dispatchAdmin,omitempty"`
 }
 
 func (s *AgentGrant) Clone() *AgentGrant {
@@ -603,6 +605,7 @@ func (s *AgentGrant) MarshalLogObject(e zapcore.ObjectEncoder) error {
 	e.AddBool("Admin", s.Admin)
 	e.AddBool("SimulationAdmin", s.SimulationAdmin)
 	e.AddBool("DatabaseAdmin", s.DatabaseAdmin)
+	e.AddBool("DispatchAdmin", s.DispatchAdmin)
 	return nil
 }
 
--- a/auth/grants_test.go
+++ b/auth/grants_test.go
@@ -120,7 +120,8 @@ func TestGrants(t *testing.T) {
 
 	t.Run("clone with Agent", func(t *testing.T) {
 		agent := &AgentGrant{
-			Admin: true,
+			Admin:         true,
+			DispatchAdmin: true,
 		}
 		grants := &ClaimGrants{
 			Identity: "identity",


erikhortsch:
## Verification

Bindings generated by CI (`Generate` ✅, `Test` ✅) — `protoc-gen-go v1.36.12` / `protoc v7.35.1`, matching the rest of the repo. Only the two new files were added; no version-stamp churn across other twirp files.

**Contract surface** — asserted with a throwaway test in `livekit/` (removed after):

- `AgentDispatchQueue` exposes exactly 12 methods; both `NewAgentDispatchQueueProtobufClient` and `NewAgentDispatchQueueJSONClient` satisfy it.
- `AddJobsRequest` round-trips through `proto.Marshal`/`Unmarshal` preserving `group_id`, `room_name`, and `WebhookConfig.signing_key`.
- `QUEUED_JOB_STATUS_UNSPECIFIED` is the zero value.

**No OSS breakage.** `AgentDispatchService` still has exactly 3 methods, and `git diff main..HEAD -- livekit/livekit_agent_dispatch.pb.go livekit/livekit_agent_dispatch.twirp.go protobufs/livekit_agent_dispatch.proto` is **0 lines** — the interface `livekit/pkg/service/agent_dispatch_service.go` implements is byte-identical, so no implementer can have broken.

Worth noting how that was checked: `livekit` does not currently build against protocol `main` at all — `pkg/rtc/participant.go:415` fails with `sessionTimer.Advance returns 3 values`. I confirmed that is **pre-existing drift** by reproducing it against protocol `dc17522` without this branch's commit. It is unrelated to this PR, but it does mean a whole-repo `go build` could not be the proof; the byte-identical-interface check above is used instead, and is strictly stronger.

**AP-979 exit test — `agents-private` builds against the contract.** With a temporary local `replace`, a scratch test in the `agent-service` module constructed `livekit.AddJobsRequest` / `QueuedJobInput`, instantiated the generated client, and read `auth.AgentGrant{DispatchAdmin: true}`. Passes. Scratch test and `replace` reverted; `agents-private` and `livekit` are both back to a clean tree.

**protocol** — `go build ./...` clean, `go test -race ./auth/... ./livekit/...` passes.

protobufs/livekit_agent_dispatch_queue.proto
```diff
@@ -0,0 +1,315 @@
...
+  // set means every distinct value gets its own budget of this size, so one
+  // rule can cover all agents without enumerating them.
+  string key = 2;
+
+  // Jobs from this scope running at once. Zero leaves the scope uncapped.
+  uint32 max_concurrent = 4;
+  // Dispatch starts allowed per second. The scheduler releases the budget in
+  // slices across admission rounds rather than at the top of each second. Zero
+  // leaves the scope unrated.
+  uint32 max_starts_per_second = 5;
```
erikhortsch: Part of me wonders if this should be a float, so that rates like "1 every 3 seconds" could be set.

shreyasj-livekit (approved):

protobufs/livekit_agent_dispatch_queue.proto
```diff
@@ -0,0 +1,338 @@
...
+message ResumeJobGroupRequest {
+  optional string group_id = 1 [(logger.name) = "groupID"];
+}
+
+message ResumeJobGroupResponse {}
+
+// Cancels the queued jobs in the group; jobs already in flight run to
+// completion.
+message CancelJobGroupRequest {
+  optional string group_id = 1 [(logger.name) = "groupID"];
```
shreyasj-livekit: For JobGroup related requests, what's the reason for group_id to be optional? If the value is not passed, what is the expected outcome?

protobufs/livekit_agent_dispatch_queue.proto
```diff
@@ -0,0 +1,338 @@
...
+  // re-queueing it immediately. Unset clears any start_after the job carried,
+  // since the original hold has already elapsed.
+  optional google.protobuf.Timestamp start_after = 3;
+}
+
+message RetryJobsResponse {
+  repeated QueuedJob jobs = 1;
+}
+
+message ListJobGroupsRequest {}
```
shreyasj-livekit: should we have a pagination param here as well?

protobufs/livekit_agent_dispatch_queue.proto
erikhortsch: This is a good question for cancel - on other requests the idea is a default "null" group, but I don't know that it makes sense for cancel to take that as a parameter.

protobufs/livekit_agent_dispatch_queue.proto
erikhortsch: Good point, yes, I think so

protobufs/livekit_agent_dispatch_queue.proto
```diff
@@ -0,0 +1,338 @@
...
+  message StatusCount {
+    QueuedJobStatus status = 1;
+    uint32 count = 2;
+  }
+}
+
+// Which slice of the project's queued traffic a DispatchLimit applies to.
+enum DispatchLimitScope {
+  // The project's whole queue.
+  DISPATCH_LIMIT_SCOPE_UNSPECIFIED = 0;
```
shreyasj-livekit: confused by this: the comment says project's whole queue but the variable is `UNSPECIFIED`. Also we may want to ensure in validation that if a 0 scope value is passed in, we reject it (since proto defaults to 0 if it isn't set). Does it also make sense then to have a separate scope value like `DISPATCH_LIMIT_SCOPE_PROJECT = 2;` for specifying a project-wide cap?

erikhortsch: That's a good point. The intent was not specifying a scope means it applies to everything (entire project), but it's probably the right call to make UNSPECIFIED invalid input to require specifically setting a scope.

protobufs/livekit_agent_dispatch_queue.proto
shreyasj-livekit: Yeah good point, I see the intent here but i wonder if floats would be difficult for customers to reason about e.g. .33 is 1 every 3 seconds (also truncating long floats in input validation could become annoying since 1 / 3 is 0.33333333...?). 

What if this was `max_starts_per_minute`? So then 1 every 3 seconds would just be 20. Although by introducing floats per second or max_starts_per_minute, does that introduce ambiguity of uniform distribution or burst?

Another idea is `min_start_interval_ms` so `200` would mean a 5 jobs/ second. (i put `ms` because then you can express more than 1 job start per second).

erikhortsch: My hesitation with `min_start_interval_ms` is that it doesn't exactly map to reality. If someone said 5ms, we wouldn't start a job every 5ms. We'd start 40 of them every 200ms when the scheduler runs.

Claude first had this as 2 fields, a value and a duration, but I didn't really like that either as it kind of implied something we wouldn't enforce in the opposite direction.

That said, starts_per_minute, while also not the thing we'd technically be enforcing, is the simplest.

protobufs/livekit_agent_dispatch_queue.proto
```diff
@@ -0,0 +1,355 @@
...
+  // from a deliberate zero, so a limit that omits one is rejected rather than
+  // silently given the widest possible reach.
+  DISPATCH_LIMIT_SCOPE_UNSPECIFIED = 0;
+  // QueuedJobInput.agent_name.
+  DISPATCH_LIMIT_SCOPE_AGENT = 1;
+  // The project's whole queue.
+  DISPATCH_LIMIT_SCOPE_PROJECT = 3;
+
+  // 2 was a per-destination label scope, dropped before release.
+  reserved 2;
```
biglittlebigben: We don't need to keep backward compatibility before release

protobufs/livekit_agent_dispatch_queue.proto
biglittlebigben: A lifetime of work in media made we very wary of floats as a way to measure duration. Is there a standard fractional duration type in proto? Or simply a fractional type?

shreyasj-livekit (approved):

real-danm (approved):

shreyasj-livekit (approved):

PR #1750, opened 2026-08-29
fix(sip): raise header value cap to 4 KB for carrier tokens like X-Twilio-CallToken

rkfshakti:
## Summary

`CreateSIPParticipant` rejects valid Twilio `X-Twilio-CallToken` values over 1,024 bytes:

```
TwirpError: invalid header value for X-Twilio-CallToken:
header X-Twilio-CallToken: value too long (max 1024 characters)

HTTP status: 400
Twirp code: invalid_argument
at SipClient.createSipParticipant
```

Twilio's Immutable Call Forwarding flow requires this token to be forwarded unchanged on the outbound leg so the original caller ID can be presented. The token is opaque and may contain SHAKEN/STIR and DIV PASSporTs, so truncating it is not valid either.

## Root cause

`ValidateHeaderValueResult` caps **every** header value at 1,024 characters. RFC 3261 does not bound individual header values — §7.1 bounds the whole message (8 KB datagram over UDP, larger over TCP). The 1 KB figure is this codebase's own policy and sits below the size real carrier tokens reach.

## Fix

- Introduce `maxHeaderValueLength = 4096` and reject values above it. 4 KB keeps a single header from dominating an 8 KB datagram while tolerating real-world carrier tokens. Character-set validation is unchanged.
- Update the valid/invalid test boundary cases (1024 & 4096 pass; 4097 fails) and add a changeset.

Fixes #789


rkfshakti:
Friendly ping for review — this raises the SIP header value cap so long X-Twilio-CallToken headers (large Twilio call SIDs) aren't rejected; pairs with livekit/sip #822 which depends on it. Test and CLA are green on the current head. Could someone take a look when convenient? Happy to make any adjustments needed.

PR #1753, opened 2026-08-31
utils/list: make mutations on unowned nodes no-ops

paulwe:
`Remove`, the moves, and the inserts corrupted the list when handed a node it did not own: `unlink` on a zeroed hook cleared `head` and `tail`, silently stranding every remaining element, and inserting an already-linked node forked the list. The failure surfaces arbitrarily far from the misuse — we hit it as a session actor whose remaining peers all kept their map entries and deadlines but became unreachable by every list walk, freezing its timer loop.

Every mutation is now guarded by a membership check and no-ops on an unowned node, matching `container/list`, where unowned-element mutations are no-ops. A zeroed hook alone is ambiguous between "not linked" and "sole element", so the check settles it against `head`. A node linked into a *different* list is indistinguishable without a per-node list pointer and remains the caller's contract (documented on `List`).

Tests cover the double-remove, unlinked moves, unlinked insert marks, re-insert of a linked node, and the sole-element edge; existing tests are unchanged.

### DIFF
--- a/utils/list/list.go
+++ b/utils/list/list.go
@@ -32,9 +32,19 @@ type Hooked[T any] interface {
 	hookAccessor[T]
 }
 
+// List is an intrusive doubly-linked list. Mutations involving a node the list
+// does not own are no-ops, as in container/list. A node linked into a DIFFERENT
+// list is indistinguishable from a member without a per-node list pointer and
+// remains the caller's contract.
 type List[T any, P Hooked[T]] struct {
 	head P
 	tail P
+}
+
+// linked reports whether it is a member of this list. A zeroed hook is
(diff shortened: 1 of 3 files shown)


utils/list/list.go
```diff
@@ -58,10 +68,16 @@ func (l *List[T, P]) PushBack(it P) {
 }
 
 func (l *List[T, P]) InsertBefore(it, mark P) {
+	if !l.linked(mark, mark.getListHook()) {
```
boks1971: Local variable for getListHook like a couple of methods have done below. This one and the next method calling it twice.

cnderrauber (approved):

boks1971 (approved):

Issue #1778, opened 2026-09-08
SimulationRun.Create.Request: add deployment so Cloud simulations can pin AgentDispatch


rdysell:
## Summary

`SimulationRun.Create.Request` in `protobufs/livekit_agent_simulation.proto` has no `deployment` field. Cloud simulations that target a live agent (`agent_name`) therefore cannot select a non-production deployment. Omitted AgentDispatch `deployment` defaults to **production**.

`CreateAgentDispatchRequest.deployment` already exists. Please add the same field to the simulation create request and have the Cloud simulation service pass it through when dispatching the agent job.

Companion CLI issue: https://github.com/livekit/livekit-cli/issues/965

## Required protocol change

Add `string deployment = …;` to `SimulationRun.Create.Request` (next available field after `ci = 13`). Empty/unset remains production, matching AgentDispatch.

## Safety motivation

Projects that share one Cloud agent name across production + staging + dev cannot run staging simulations safely until this field exists and is honored end-to-end (CLI → SimulationRun create → AgentDispatch).

Verified against `@livekit/protocol@1.51.0` (`94a0a8d`): create request fields are `project_id`, `agent_name`, `num_simulations`, `region`, `scenario_group`, `concurrency`, `mode`, audio impairments, `ci` — no `deployment`.

rdysell:
Opened https://github.com/livekit/protocol/pull/1780 — adds `deployment` on `SimulationRun.Create.Request` (field 14), `SimulationRun` (field 23), and `SimulationDispatch` (field 5). Empty/unset stays production.

Companion CLI: https://github.com/livekit/livekit-cli/pull/966

The pin is not complete until the Cloud simulation service copies create `deployment` onto `CreateAgentDispatchRequest.deployment` so the simulated user waits in the same room as the pinned worker.

PR #1780, opened 2026-09-08
Add deployment to SimulationRun create for AgentDispatch pinning

rdysell:
## Summary

`SimulationRun.Create.Request` had no `deployment` field, so Cloud simulations against a live agent (`agent_name`) could not pin a non-production worker. Omitted `CreateAgentDispatchRequest.deployment` defaults to **production**. Projects that share one Cloud agent name across production + staging + dev cannot run staging simulations safely until this field exists and the Cloud simulation service copies it onto AgentDispatch.

Companion CLI: livekit/livekit-cli#965
Tracking: livekit/protocol#1778

Docs: [Non-production deployments](https://docs.livekit.io/deploy/agents/deployments/) — omit `--deployment` / empty string = production.

## Protocol change

| Message | Field | Notes |
|---|---|---|
| `SimulationRun.Create.Request` | `string deployment = 14` | CLI → create API |
| `SimulationRun` | `string deployment = 23` | Persist what the run targeted |
| `SimulationDispatch` | `string deployment = 5` | Room metadata; **does not** select the worker |

Empty/unset remains production, matching `CreateAgentDispatchRequest.deployment`.

## Cloud backend (required for the pin)

A proto field the simulation service ignores is not a pin. Please copy `Create.Request.deployment` onto `CreateAgentDispatchRequest.deployment` (or equivalent job routing) so the simulated user waits in the **same** room the pinned worker joins. Rewriting `--agent-name` plus a separate `lk dispatch create --deployment` does not work: Cloud simulate waits for the original agent name in its own room.

## Codegen

Regenerated with `go tool mage proto` and protoc **35.1** (same pin as `.github/workflows/generate.yaml`). JS gen left to the Generate workflow / `pnpm --filter @livekit/protocol run generate:proto`.

## Backwards compatibility

Purely additive. Existing callers that omit `deployment` keep production dispatch.


rdysell:
CLI companion: https://github.com/livekit/livekit-cli/pull/966

PR #1730, opened 2026-08-20
TEL-912 Adding a transferResponse with a reason and a status

genseric-ghiro:
## Summary
- Adding a TransferSIPParticipantResponse that contains a status and a reason
- New `SIPTransferReason` enum to give more details about the status of the transfer

PR #1782, opened 2026-09-08
TEL-1026: Transfer error details

genseric-ghiro:
## Summary

- Instead of adding the reason/status on transfer failures to a response (which gets stripped by psrcp and twirp when there's an error), add it directly to the error.


livekit/sip.go
```diff
@@ -53,6 +89,48 @@ func (p *SIPStatus) Error() string {
...
+		if p.Code < 200 {
+			code = codes.Unknown // 1xx are not final responses, something is wrong.
+		} else if p.Code < 300 {
+			return status.New(codes.OK, "OK") // Preserving previous behavior
+		} else if p.Code < 500 {
+			code = codes.InvalidArgument
+		} else if p.Code < 600 {
+			code = codes.FailedPrecondition // 5xx from remote server, per guideline (c) in gRPC docs
+		} else if p.Code < 700 {
+			code = codes.InvalidArgument // Same as 4xx ,but authoritative
```
alexfish8: nit: `4xx, but`

alexfish8 (approved):

protobufs/livekit_sip.proto
```diff
@@ -874,7 +874,13 @@ message TransferSIPParticipantRequest {
   google.protobuf.Duration ringing_timeout = 6;
 }
 
+// Deprecated: every transfer outcome other than STR_COMPLETED is reported as an
+// error carrying SIPTransferError. This message only still reports
+// STR_CALL_ENDED, which is temporarily returned as a success, and will carry
+// nothing once that becomes an error too.
 message TransferSIPParticipantResponse {
+  option deprecated = true;
```
dennwc: We shouldn't deprecate the whole message that is used as a response type in a non-deprecated RPC.

I think you've meant that `sip_status` field specifically is deprecated, right? Or maybe the `status` and `reason` as well? In either of these cases, please mark those fields as deprecated instead.

livekit/sip.go
```diff
@@ -26,24 +26,60 @@ const MaxSIPMediaTimeout = 10 * time.Minute
...
+
+	_ xtwirp.ErrorMeta = (*SIPTransferError)(nil)
 )
 
-// SIPStatusFrom unwraps an error and returns associated SIP call status, if any.
-func SIPStatusFrom(err error) *SIPStatus {
+// SIPTransferErrorFrom unwraps an error and returns the associated SIP transfer
+// details, if any. A rejected transfer also carries a SIPStatus, see SIPStatusFrom.
+func SIPTransferErrorFrom(err error) *SIPTransferError {
 	st, ok := status.FromError(err)
```
dennwc, line 41: Unrelated to your change, but we should also test `errors.AsType[*SIPTransferError](err)` here at the very beginning as well, to quickly unwrap local errors.

The `status.FromError` is the path is still valid for remote errors. And it will work for local errors too, but with a long roundtrip through protobuf types.

Same for `SIPStatusFrom`.

livekit/sip.go
```diff
@@ -26,24 +26,60 @@ const MaxSIPMediaTimeout = 10 * time.Minute
...
+		if e, ok := d.(*SIPTransferError); ok {
 			return e
 		}
 	}
 	return nil
 }
 
-func (p SIPStatusCode) ShortName() string {
-	return strings.TrimPrefix(p.String(), "SIP_STATUS_")
+func (p *SIPTransferError) TwirpErrorMeta() map[string]string {
```
dennwc, line 86: The `SIPTransferError` should also implement two more optional interfaces:

```go
// Returns *SIPStatus, if set.
Unwrap() error

// Just calls the same method on SIPStatus?
GRPCStatus() *status.Status
```

protobufs/rpc/sip.proto
```diff
@@ -183,6 +183,13 @@ message InternalTransferSIPParticipantRequest {
   map<string, string> feature_flags = 6;
 }
 
+// Being deprecated: every transfer outcome other than STR_COMPLETED is reported
```
dennwc: We still have an internal API using this request. Only okay to attempt a deprecatio when we have a V2 replacing that API.

protobufs/livekit_sip.proto
genseric-ghiro: I recently added that response in https://github.com/livekit/protocol/pull/1730. Previously, it was just an `Empty` proto. I then realized its main purpose (providing more details to the caller if the transfer failed) was defeated by the fact that psrpc and twirp strip the response if there's an error. My goal with this deprecation was just to roll it back to an `Empty` proto, since that response is pretty useless atm. Let me know what you think

protobufs/rpc/sip.proto
genseric-ghiro: Same comment as above. This is a response that I added 2 weeks ago, but that doesn't serve its original purpose

livekit/sip.go
genseric-ghiro: Fixed

livekit/sip.go
genseric-ghiro: Fixed

protobufs/livekit_sip.proto
genseric-ghiro: Clarified the comment like discussed

protobufs/rpc/sip.proto
genseric-ghiro: ditto

dennwc (approved):

PR #1783, opened 2026-09-08
Add new phone number state for outbound-only.

dennwc:


hechen-eng (approved):

dennwc:
Abandoning this, have a better idea.

PR #1785, opened 2026-09-08
agent_simulation: add stable id to Scenario and ScenarioGroup

u9g:
Adds \`id\` to \`Scenario\` and \`ScenarioGroup\`.

Today nothing correlates runs of the same scenario over time: the label, instructions, and agent expectations can all be edited, and hashing the content would make every edit a new identity. A user-authored \`id\` in \`scenarios.yaml\` gives dashboards a stable key for charting a scenario's success rate across runs. The group-level \`id\` does the same for the scenarios file as a whole. Run-level identity already exists as \`SimulationRun.id\`.

Companion CLI change: livekit/livekit-cli (requires IDs in scenarios.yaml and offers to insert them).


theomonnom (approved):

u9g:
Done: ids are now generated with the shared guid utility, prefixed `SCN_` (scenario) and `SCNG_` (scenario group). Prefixes live in `utils/guid` and are re-exported from `utils`. The companion CLI PR livekit/livekit-cli#970 is updated to use them.

PR #1788, opened 2026-09-09
Add compoment min level logging

cnderrauber:
To Support per project component logging level

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -227,21 +227,42 @@ type ZapComponentLeveler interface {
 	ComponentLevel(component string) zapcore.LevelEnabler
 }
 
+// ZapComponentMinLeveler resolves a per-component minimum level for a logger,
+// on top of the levels configured globally. It returns nil for components it
+// has no opinion on.
+type ZapComponentMinLeveler interface {
+	ComponentMinLevel(component string) zapcore.LevelEnabler
+}
+
 type ZapLogger interface {
 	Logger
 	ToZap() *zap.SugaredLogger
 	ComponentLeveler() ZapComponentLeveler
 	WithMinLevel(lvl zapcore.LevelEnabler) Logger
+	// WithComponentMinLeveler lowers the level per component, for this logger
+	// and its descendants. Unlike WithMinLevel it is reported by
+	// ComponentLeveler, so it also opens up subsystems that do their own level
+	// gating.
+	WithComponentMinLeveler(ml ZapComponentMinLeveler) Logger
 }
 
 type zapLogger[T zaputil.Encoder[T]] struct {
 	zap *zap.SugaredLogger
 	*zapConfig
-	enc       T
-	component string
-	deferred  []*zaputil.Deferrer
-	sampler   *zaputil.Sampler
-	minLevel  zapcore.LevelEnabler
+	enc                 T
+	component           string
+	deferred            []*zaputil.Deferrer
+	sampler             *zaputil.Sampler
+	minLevel            zapcore.LevelEnabler
+	componentMinLeveler ZapComponentMinLeveler
+}
+
+// componentMinLevel returns the min leveler's opinion on component, or nil.
+func (l *zapLogger[T]) componentMinLevel(component string) zapcore.LevelEnabler {
+	if l.componentMinLeveler == nil {
+		return nil
+	}
+	return l.componentMinLeveler.ComponentMinLevel(component)
 }
 
 func FromZapLogger(log *zap.Logger, conf *Config, opts ...ZapLoggerOption) (ZapLogger, error) {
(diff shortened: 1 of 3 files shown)


boks1971 (approved):
`component` log levels already work

cnderrauber:
> `component` log levels already work

@boks1971  Does it already have project-level component logging? I have not found the relevant config.

logger/logger.go
```diff
@@ -331,6 +354,14 @@ func (l zapLoggerComponentLeveler[T]) ComponentLevel(component string) zapcore.L
 		component = l.zl.component + "." + component
 	}
 
+	// Note this reports the logger's component min leveler but not its min
+	// level: see ZapLogger.WithMinLevel. levelEnablers is shared by every
+	// logger built from this config, so a min leveler's opinion, which is
+	// particular to one logger, must not be cached in it.
+	if override := l.zl.componentMinLevel(component); override != nil {
+		return zaputil.OrLevelEnabler{l.zl.sc.ComponentLevel(component), l.zl.tap, override}
+	}
```
paulwe: this should be cached

logger/logger.go
```diff
@@ -298,12 +319,14 @@ func newZapLogger[T zaputil.Encoder[T]](zap *zap.SugaredLogger, zc *zapConfig, e
...
-	if l.minLevel == nil {
+	// writeEnablers is shared by every logger built from this config, so it can
+	// only cache enablers that depend on nothing but the component.
+	if componentMinLevel := l.componentMinLevel(l.component); componentMinLevel == nil && l.minLevel == nil {
 		console, _ = l.writeEnablers.LoadOrCompute(l.component, func() (*zaputil.WriteEnabler, bool) {
 			return zaputil.NewWriteEnabler(os.Stderr, l.sc.ComponentLevel(l.component)), false
 		})
 	} else {
-		enab := zaputil.OrLevelEnabler{l.minLevel, l.sc.ComponentLevel(l.component)}
+		enab := zaputil.NewOrLevelEnabler(l.minLevel, l.sc.ComponentLevel(l.component), componentMinLevel)
```
paulwe: this allocates for every derived logger

logger/logger.go
```diff
@@ -227,21 +227,42 @@ type ZapComponentLeveler interface {
 	ComponentLevel(component string) zapcore.LevelEnabler
 }
 
+// ZapComponentMinLeveler resolves a per-component minimum level for a logger,
+// on top of the levels configured globally. It returns nil for components it
+// has no opinion on.
```
paulwe: can we trim the slop comments?

logger/logger.go
cnderrauber: It is an equivalent of `zaputil.OrLevelEnabler{}` that doesn't allocate more.

cnderrauber: It is an equivalent of zaputil.OrLevelEnabler{} that doesn't allocate more.

boks1971:
> > `component` log levels already work
> 
> @boks1971 Does it already have project-level component logging? I have not found the relevant config.

Yeah, it is there. @paulwe did that. 

But, it is not working for pion logger. Does this address that?

boks1971:
> > > `component` log levels already work
> > 
> > 
> > @boks1971 Does it already have project-level component logging? I have not found the relevant config.
> 
> Yeah, it is there. @paulwe did that.
> 
> But, it is not working for pion logger. Does this address that?

Actually, let me retrace. I might be conflating things. Sorry.

cnderrauber:
> > > `component` log levels already work
> > 
> > 
> > @boks1971 Does it already have project-level component logging? I have not found the relevant config.
> 
> Yeah, it is there. @paulwe did that.
> 
> But, it is not working for pion logger. Does this address that?

yeah, for tracing project pion logs.

logger/logger.go
cnderrauber: cache both min level and component min level enablers in https://github.com/livekit/protocol/pull/1788/commits/fdfcb4092fe02ef0c7f5e4e30501e5d0a0499343

paulwe: we cannot serialize every `WithValues` call on a global mutex..

cnderrauber: There is a short circuit to return the OrLevelEnabler directly when ComponentMinLeveler is empty, so the mutex is hit only when the project component level is set in the config. I think in most cases the project component level is empty and behavior keeps unchange, except we turn it on for pion shortly.
Also asked claude about using xsync.Map and it said a mutex is better here..

> Technically yes — it's a drop-in, same LoadOrCompute shape as zapConfig. But I'd keep the mutex+map, because the allocation profile is what separates the two cases:
> 
> two xsync.Map:            6048 B/op, 10 allocs/op
> enablerCache{} (lazy):      24 B/op,  1 alloc/op
> enablerCache + 1 entry:    360 B/op,  3 allocs/op
> 
> zapConfig's two maps are allocated once per process and read by every logger for the lifetime of the binary — 6 KB amortizes to nothing, and lock-free reads are worth it under fleet-wide contention. This cache is allocated per project logger, and in cloud that means per participant, per relay downtrack, and per HTTP request: pkg/node/middleware.go:236 builds one inside the per-request auth middleware, as does settings.go:96. For a debug-enabled project that's a cache per request, and almost none of those requests ever resolve a pion component — with the lazy maps they pay 24 bytes and never allocate a map at all.

PR #1790, opened 2026-09-09
logger: replace WithTap with WithTee

paulwe:
`WithTap` took a `WriteSyncer`, so a tap could only ever receive bytes encoded by a zaputil-chosen encoder — `Encoder.Core` took two destinations, and `DevelopmentEncoder` carried a second JSON encoder that existed only to format tap output. `WithTee` hands the duplicate stream a `zapcore.Core` instead, built per derived logger from that logger's resolved enabler, so it brings its own encoder and sees the structured entry.

Four commits, each green on its own:

1. **`replay deferred writes through Check`** — `Deferrer.write`/`flush` called the wrapped core's `Write` directly. `zapcore.multiCore.Write` fans out to every child unconditionally, so a destination that filters in `Check` rather than in `Write` received entries it never enabled. `encoderCore` was immune because it re-checks each out's level inside `Write`; nothing else is, including `zaptest/observer`.
2. **`replace WithTap with WithTee`** — the API change, plus allocation benchmarks.
3. **`replace encoderCore with zapcore.NewCore`** — with one destination per core, `encoderCore` is `ioCore` plus the write-time level check that commit 1 made redundant.
4. **`drop the Encoder type parameter`** — both encoders are now one encoder and one destination, so the interface and its two implementations collapse.

### Level semantics

A tee shares the console's resolved level and cannot widen what the logger emits. That is what lets the component leveler drop its `OrLevelEnabler` and the `levelEnablers` cache go away entirely.

### Allocations

`WithValues`/`WithDeferredValues` run per participant and track, so `logger/logger_bench_test.go` covers objects allocated on both, with and without a tee. Every no-tee path is flat or better than before:

| | before | after |
|---|---|---|
| `WithValues` console | 18 | 10 |
| `WithValues` json | 11 | 10 |
| `WithDeferredValues` console | 14 | 9 |
| `WithDeferredValues` json | 10 | 9 |
| deferred write, console | 10 | 9 |
| deferred flush, console | 30 | 24 |
| plain writes | 8 / 4 / 1 | unchanged |

Console mode gains most: dropping `DevelopmentEncoder`'s second encoder removes a `Clone()` per derivation and a core plus a `NewTee` slice per `makeZap`. Commit 1 also stopped the deferred path encoding every entry to JSON for the discard tap before throwing it away (−1 alloc, ~12% off dev-mode deferred writes).

### Breaking

Removes `logger.WithTap`, `zaputil.NewDiscardWriteEnabler`, `zaputil.NewEncoderCore`, and `zaputil.Encoder[T]`/`DevelopmentEncoder`/`ProductionEncoder` as distinct types. Minor changeset included. Known downstream caller: `cloud-protocol/audit_test.go`, a one-line swap — it asserts `WithValues` fields on the tap, so it is also the live check that a tee accumulates them.

### Note

Overlaps #1788, which also touches `makeZap`/`WithMinLevel`.


boks1971 (approved):

cnderrauber (approved):
Looks good! Will wait for this PR to land

PR #1799, opened 2026-09-14
agent_simulation: sample scenarios within a run for pass@k / pass^k

u9g:
A run today holds exactly one job per scenario, so a single flaky attempt is indistinguishable from a broken scenario. Repeating a scenario only helps if repeating can also change its verdict, which is what the pass rate is for.

- `SimulationRun.Sampling`: `samples` (attempts per scenario, > 1) and `pass_rate` (the share of a scenario's samples that must pass for that scenario to pass, 0-1). A scenario needs `round(rate × samples)` passing samples, so `0.34` of 3 needs 1 and `0.67` of 3 needs 2; `1` requires every sample. Both are required when the message is present.
- `SimulationRun.Create.Request.sampling`, mirrored on `SimulationRun.sampling`. Absent runs each scenario once and requires it to pass, which is today's behaviour. Requires `scenario_group`. The mirror lets the dashboard render a scenario's verdict without being told the bar out of band.
- `SimulationRun.Job.scenario_id` / `sample`: every attempt of one scenario shares the id, so a run's jobs group back into scenarios. New field numbers; 9 stays reserved.

A scenario stops being all-or-nothing: `2/3` can now be a pass. The run still fails if any scenario fails, so there is no run-level percentage. pass@k and pass^k remain reporting metrics over scenarios and neither gates.

`pass_rate` is meaningless without `samples > 1` — at one sample `round(rate × 1)` is either a no-op or always-passes — so the two travel together rather than as two independently-optional fields that can be combined into a request that means nothing. Inside `Sampling` both are plain scalars, because `0 samples` and a `0` pass rate are naturally invalid and so already say "unset" without a presence bit.

Note for consumers: `Sampling.pass_rate` is the threshold a run was given, and `RunMetrics.scenario_pass_rate` is the observed share of passing jobs. They are a bar and a measurement, not two readings of the same thing.

**Consumer migration.** `samples` and `pass_rate` were briefly flat fields on this branch, never on main and never released, so `sampling` reuses their field numbers rather than reserving them. livekit/agents-private#457 is merged and pins a commit of this branch in `go.mod`, so it needs `GetSampling().GetSamples()` and a repin. livekit/livekit-cli#975 (`lk agent simulate --samples` / `--pass-rate`) and livekit/web#6190 follow.


protobufs/livekit_agent_simulation.proto
```diff
@@ -75,6 +75,12 @@ message SimulationRun {
     Usage usage = 13;
     // Quality metrics for this job's call. Unset when the job produced none.
     JobMetrics metrics = 14;
+    // The Scenario.id this job ran; empty for generated runs. Every attempt
+    // of one scenario shares it, so a run's jobs group back into scenarios.
+    string scenario_id = 15;
+    // 1-based attempt of that scenario within the run; always 1 when the run's
+    // repeats is 1.
+    int32 attempt = 16;
```
theomonnom:
```suggestion
    int32 sample = 16; 
```

theomonnom:
```suggestion
    int32 i = 16;
```

theomonnom (approved):

protobufs/livekit_agent_simulation.proto
u9g: Renamed: the count is now `samples` (f9e48c3) and the per-job index is `sample` (a58538a). agents-private#457 and livekit-cli#975 follow.

u9g: Renamed: the count is now `samples` (f9e48c3) and the per-job index is `sample` (a58538a). agents-private#457 and livekit-cli#975 follow.

u9g:
Added `min_sample_pass_rate` per the decision that the run is gated on a pass rate and that pass@k / pass^k only report. Re-review needed — this PR was approved before the field existed.

The semantics are narrower than "pass^k" suggested, and worth stating plainly because the field comment fixes them for every consumer:

**A scenario stops being all-or-nothing.** The rate is the share of *one scenario's samples* that must pass for that scenario to pass — `2/3` becomes a passing scenario at `--pass-rate 0.67`. It is not an exponent and not a percentage of scenarios. The run then fails if any scenario fails, exactly as it does today, so there is no run-level percentage anywhere.

The rounding is `round(rate × samples)`, so `0.34` of 3 samples needs 1 and `0.67` of 3 needs 2. A strict `c/k >= rate` comparison would land a step stricter than the number reads, because `1/3` is `0.333` and `2/3` is `0.667` — a typed `0.34` would demand 2 of 3 and a typed `0.67` would demand 3 of 3.

`0`/unset requires every sample, which is the current behaviour, so existing runs are unaffected.

protobufs/livekit_agent_simulation.proto
```diff
@@ -254,6 +260,14 @@ message SimulationRun {
...
       bool packet_loss = 12;
       optional CI ci = 13;
+      // Samples per scenario; 0/unset = 1. Requires scenario_group. The run
+      // holds scenarios × samples jobs, subject to the usual job cap.
+      int32 samples = 14;
+      // The share of a scenario's samples that must pass for that scenario to
+      // pass, 0-1; 0/unset requires every sample. A scenario needs
+      // round(rate × samples) of them. The run fails if any scenario fails;
+      // pass@k and pass^k report over scenarios and neither gates.
+      double min_sample_pass_rate = 15;
```
theomonnom: maybe just pass_rate?
```suggestion
      double pass_rate = 15;
```

u9g: Done, on both the request and the mirrored field on `SimulationRun`.

